PR #68, opened 2017-01-18
Add localTimestamp and bypass ngrok with localhost urls 

tomlm:
localTimestamp is new property 
ngrok rate limits and so no need to use for local to local activity

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "@reactivex/rxjs": "5.0.3",
     "async": "2.1.4",
     "base64url": "2.0.0",
-    "botframework-webchat": "0.4.0",
+    "botframework-webchat": "0.4.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -64,14 +64,15 @@
...
-    "botframework-webchat": "0.4.0",
+    "botframework-webchat": "0.4.3",
     "electron-debug": "1.1.0",
     "es6-shim": "0.35.2",
     "formidable": "1.0.17",
     "http-status-codes": "1.0.6",
     "jsonwebtoken": "7.2.1",
     "lock": "0.1.3",
     "mkdirp": "0.5.1",
+    "moment": "^2.17.1",
```
eanders-MS: Tom, This should be pinned to a specific version (carat needs to be removed).

PR #102, opened 2017-02-10
Added select all to menubar/titlebar

sozercan:
makes it easy to select the entire message to copy or delete


eanders-MS:
This is fantastic, thank you.

Issue #65, opened 2017-01-15
emulator using ngrok for localhost


amitbend:
It seems that the emulator is using for no reason ngrok when testing localhost bot.
the issue is that ngrok has limitation on traffic frequency, and i'm being denied using it.

eanders-MS:
Agreed, it shouldn't do this for localhost.

eanders-MS:
Referencing commit https://github.com/Microsoft/BotFramework-Emulator/commit/a4457d2282d95f1daeea3852fe2408c209e9e53a, which contains a fix for this issue. Will be included in the next emulator release.

amitbend:
Great, i believe we can close the issue.

kbremner:
If an ngrok path is set, the emulator reports in the logs that it is accessible at the ngrok URL. It is expected that if an ngrok path is set, the ngrok URL will be reported as the service URL.

This new behaviour is unexpected and also causes dockerised bots to report a 500 error when attempting to start a new conversation, as they are unable to, by default, connect to the localhost service URL.

In my opinion, a change should be made in one or more of the following ways:
1 - At least log that this is the behaviour (localhost service URL being used for localhost bot). At the moment, it is hidden, unexpected and confusing
2 - Remove the check [here](https://github.com/Microsoft/BotFramework-Emulator/blob/master/src/server/conversationManager.ts#L119). If the user doesn't want to use ngrok for a localhost bot, they can remove the ngrok path from app settings
3 - Make the behaviour configurable via app settings

kbremner:
Any response to my previous comment?

eanders-MS:
@kbremner Thanks for the feedback. It's definitely a rough edge in the design. Addressing your points:
1 - Good suggestion, thanks.
2 - Deleting the ngrok path feels inappropriate since it would break remote connections as well. I think a checkbox in App Settings might be better: "Bypass ngrok for localhost connections".
3 - Is this covered by the checkbox suggestion?

kbremner:
Thanks for taking it in to consideration :)

I agree in terms of the best course of action being to add a checkbox to app settings, with the logging mentioned in point 1.

I can open a pull request if you agree with those changes?

eanders-MS:
Certainly, that would be fantastic.

PR #123, opened 2017-03-15
Makes behaviour of ngrok for localhost bots configurable

kbremner:
Adds a checkbox that is used to determine whether ngrok should be used for bots served from localhost or not (see discussion in #65). Also logs the behaviour when ngrok is started or the value of the checkbox is changed.

### DIFF
--- a/src/client/dialogs/appSettingsDialog.tsx
+++ b/src/client/dialogs/appSettingsDialog.tsx
@@ -44,13 +44,13 @@ import * as Constants from '../constants';
 
 interface IAppSettings {
     ngrokPath?: string,
-    ngrokForLocalhostBot?: boolean
+    bypassNgrokLocalhost?: boolean
 }
 
 export class AppSettingsDialog extends React.Component<{}, { ngrokPath: string }> {
     settingsUnsubscribe: any;
     ngrokPathInputRef: any;
-    ngrokForLocalhostBotInputRef: any;
+    bypassNgrokLocalhostInputRef: any;
     showing: boolean;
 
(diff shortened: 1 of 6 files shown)


src/client/dialogs/appSettingsDialog.tsx
```diff
@@ -144,9 +147,21 @@ export class AppSettingsDialog extends React.Component<{}, {}> {
...
                             </div>
+                            <div className="input-group appsettings-checkbox-group">
+                                <label className="form-label">
+                                    <input
+                                        type="checkbox"
+                                        ref={ref => this.ngrokForLocalhostBotInputRef = ref}
+                                        className="form-input"
+                                        defaultChecked={ serverSettings.framework.ngrokForLocalhostBot }
+                                        disabled={ this.state ? !this.state.ngrokPath : !serverSettings.framework.ngrokPath } />
+                                    Use ngrok for bots served from localhost
```
eanders-MS: Let's reverse the logic here. Have the label state "Bypass ngrok for local addresses". This better matches Windows' proxy server settings screen:
image

eanders-MS: label and checkbox elements need a hover state of "cursor: pointer". I suggest adding a css class called `.clickable:hover` to emulator.css, and then apply `clickable` to these elements.

eanders-MS:
Thank you for this contribution! I just had a few minor change requests after which I will be happy to merge it in.

kbremner:
Thanks for the comments, both make sense and I've made the required changes

Issue #118, opened 2017-03-12
Add a confirmation dialog before closing


CXuesong:
This might be a rather minor aspect, but it kept bugging me these days. I start up the API service application in VS, and a browser window is shown. Then I try my bot in emulator. When I want to close the service app and do some changes, I often just instinctively close the emulator window before I realize what I should have closed is the browser. Then I have to restart the emulator, wait a moment, and reconfigure the API endpoint address.

I don't know if any other people have met similar situations, but IMO, it would be handy if a confirmation is shown before the emulator exits.

eanders-MS:
Thank you for the feedback. In my own workflow I think this would be more annoying than helpful, but I'd be interested to know how other users feel about the idea.

That said, I think it would be straight forward to implement this feature, if you wanted to make these changes for yourself and run from a local build.

danmarshall:
Perhaps the emulator can persist the UI settings

eanders-MS:
Also, reconnect to the last bot on startup might be generally useful.

kbremner:
Persisting the settings sounds like it would be useful and relatively simple to implement

cwhitten:
Closing for lack of activity

PR #128, opened 2017-03-18
Save/reload last connected bot

CXuesong:
This will solve #118 , I suppose.

An option has been added to the settings dialog to enable this feature.

### DIFF
--- a/src/client/dialogs/appSettingsDialog.tsx
+++ b/src/client/dialogs/appSettingsDialog.tsx
@@ -51,7 +51,6 @@ export class AppSettingsDialog extends React.Component<{}, { ngrokPath: string }
     settingsUnsubscribe: any;
     ngrokPathInputRef: any;
     bypassNgrokLocalhostInputRef: any;
-
     showing: boolean;
 
     pageClicked = (ev: Event) => {
--- a/src/client/dialogs/appSettingsDialog.tsx
+++ b/src/client/dialogs/appSettingsDialog.tsx
@@ -74,8 +74,8 @@ export class AppSettingsDialog extends React.Component<{}, { ngrokPath: string }
     onAccept = () => {
         ServerSettingsActions.remote_setFrameworkServerSettings({
             ngrokPath: this.ngrokPathInputRef.value,
-            bypassNgrokLocalhost: this.bypassNgrokLocalhostInputRef.checked
-            promptBeforeExit: this.promptBeforeCloseCheckRef.checked
+            bypassNgrokLocalhost: this.bypassNgrokLocalhostInputRef.checked,
+            promptBeforeExit: this.promptBeforeCloseCheckRef.checked,
         });
         AddressBarActions.hideAppSettings();
     }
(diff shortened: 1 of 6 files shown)


kbremner:
Is this necessary?
Based on the discussion in #118, it seems like the solution to the actual problem is to persist settings between runs of the emulator?

eanders-MS:
@CXuesong I appreciate your effort here but as @kbremner referenced, we're going to restore the bot connection on startup rather than prompt on exit.

CXuesong:
Sorry, but perhaps I opened this PR in a hurry… So the last chosen bot will be selected on the startup now.

Still, uhm… may I keep the "prompt before exit" in the settings panel? I will remove it if it's not appropriate.

eanders-MS:
Sorry for the wait. Unfortunately not going to be able to keep the "prompt before exit". Thanks for your understanding.

CXuesong:
Okay, now we only have "restore last connected bot" feature.

Btw, I just found out that, if you delete a bot from the auto-completion list no sooner than you connected the bot (successfully or not), some `"DirectLine: Cannot get activities. No active bot"` will be shown repetitively in the console. Should [conversationsControllerV3.ts#L260](https://github.com/Microsoft/BotFramework-Emulator/blob/master/src/server/controllers/directLine/conversationsControllerV3.ts#L260) be commented out or changed to `log.trace`?

src/client/dialogs/appSettingsDialog.tsx
```diff
@@ -71,7 +72,7 @@ export class AppSettingsDialog extends React.Component<{}, { ngrokPath: string }
     onAccept = () => {
         ServerSettingsActions.remote_setFrameworkServerSettings({
             ngrokPath: this.ngrokPathInputRef.value,
-            bypassNgrokLocalhost: this.bypassNgrokLocalhostInputRef.checked
+            bypassNgrokLocalhost: this.bypassNgrokLocalhostInputRef.checked,
```
eanders-MS: No need to add a comma here.

src/client/dialogs/appSettingsDialog.tsx
```diff
@@ -84,16 +85,17 @@ export class AppSettingsDialog extends React.Component<{}, { ngrokPath: string }
...
-            defaultPath: dir,
-            properties: ['openFile']
-        }, (filenames: string[]) => {
-            if (filenames && filenames.length) {
-                // TODO: validate selection
-                this.ngrokPathInputRef.value = filenames[0];
-                this.setState({ ngrokPath: filenames[0] });
-            }
-        })
+                title: 'Browse for ngrok',
```
eanders-MS: Please undo reformatting of whitespace

src/client/dialogs/appSettingsDialog.tsx
```diff
@@ -127,15 +129,15 @@ export class AppSettingsDialog extends React.Component<{}, { ngrokPath: string }
                             <li>
                                 <a href="javascript:void(0)"
                                     className={"emu-navitem emu-navitem-selected"}
-                                    >
+                                >
```
eanders-MS: There is a lot of whitespace-only modifications in this change. Please undo them.

eanders-MS:
I'm sorry but I cannot accept this PR as-is. Something in it is buggy. Few incorrect behaviors I noticed:
- When I select a bot endpoint and click connect, my bot credentials disappear (expected), but then reappear (unexpected).
- Pressing the refresh button causes my bot's credentials to appear.

CXuesong:
Oops… So now I have moved the bot reload logic to `componentDidMount`, as a callback for `serverSettings` event. Hope that solves the problem.

eanders-MS:
Hi @CXuesong, Apologies for letting this languish, and let me explain the reason. We'd like to make a new emulator release before pulling in this feature so that it has adequate test time before going live. Additionally, we have a potentially conflicting change incoming, and would like to get that change in first, then merge this one.

CXuesong:
Well, this is now a somewhat minor change, so I will try to follow up the branch and resolve the conflicts if possible ;-)

eanders-MS:
Going ahead and closing this PR. Thank you for your effort, sorry it wasn't accepted.

PR #212, opened 2017-06-02
Window positioning and zoom functionality

mgbennet:
Added some logic to check if the main window would appear off screen when the window is restored or created. If it would appear off screen, it tries to restore to the display it was last on, otherwise it goes to the nearest monitor to its last position.

Also, added zoom functionality. Ctrl+= zooms in, Ctrl+- zooms out, and Ctrl+0 restores to the default amount. Zoom levels are shared across all windows and is remembered between sessions.

### DIFF
--- a/src/server/main.ts
+++ b/src/server/main.ts
@@ -76,20 +76,13 @@ Electron.app.on('will-finish-launching', (event, args) => {
 
 
 var windowIsOffScreen = function(windowBounds: Electron.Rectangle): boolean {
-    const corners = [
-        {x: windowBounds.x, y: windowBounds.y},
-        {x: windowBounds.x + windowBounds.width, y: windowBounds.y},
-        {x: windowBounds.x, y: windowBounds.y + windowBounds.height},
-        {x: windowBounds.x + windowBounds.width, y: windowBounds.y + windowBounds.height}
-    ];
-    // only out of bounds if all corners are off screen
-    return corners.reduce((acc, c) => {
-            let nearestDisplay = Electron.screen.getDisplayNearestPoint(c).workArea;
-            if ((c.x < nearestDisplay.x || c.x > nearestDisplay.x + nearestDisplay.width) ||
-                (c.y < nearestDisplay.y || c.y > nearestDisplay.y + nearestDisplay.height))
-                return true && acc;
-            return false;
-        }, true)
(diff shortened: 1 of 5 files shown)


src/server/main.ts
```diff
@@ -74,26 +74,49 @@ Electron.app.on('will-finish-launching', (event, args) => {
...
 
+
+var windowIsOffScreen = function(windowBounds: Electron.Rectangle): boolean {
+    const corners = [
+        {x: windowBounds.x, y: windowBounds.y},
+        {x: windowBounds.x + windowBounds.width, y: windowBounds.y},
+        {x: windowBounds.x, y: windowBounds.y + windowBounds.height},
+        {x: windowBounds.x + windowBounds.width, y: windowBounds.y + windowBounds.height}
+    ];
+    // only out of bounds if all corners are off screen
```
eanders-MS: Imagine the emulator stretched across a 3x3 grid of monitors, with the primary monitor in the center. What would this logic do in that scenario?

src/server/main.ts
```diff
@@ -74,26 +74,49 @@ Electron.app.on('will-finish-launching', (event, args) => {
...
     const settings = getSettings();
+    let initBounds: Electron.Rectangle = {
+        width: settings.windowState.width || 0,
+        height: settings.windowState.height || 0,
+        x: settings.windowState.left || 0,
+        y: settings.windowState.top || 0,
+    }
+    if (windowIsOffScreen(initBounds)) {
+        let display = Electron.screen.getAllDisplays().find(display => display.id === settings.windowState.displayId);
+        display = display ? display : Electron.screen.getDisplayMatching(initBounds);
```
eanders-MS: Could be simplified to `display = display || Electron.screen.getDisplayMatching(initBounds)`

src/server/main.ts
eanders-MS: I think a better test might be: if the rectangle of the display and the rectangle of the window bounds don't overlap, then use defaults.

mgbennet: I've changed it to check for rectangle overlap with the nearest display, which should be a little more efficient. Either test should work fine with a large grid of monitors, it just looks for the nearest display and checks if the windowBounds rectangle overlaps with that.

eanders-MS:
Looks good, thanks!

Issue #205, opened 2017-05-26
State payload exceeds 64k


danmarshall:
Since the default state service provided by BotBuilder has a 64k limit, it would be nice if the emulator could match this constraint so that bugs can be caught and reproduced early in development. Perhaps this can be configurable to emulate a custom state service.

eanders-MS:
Thanks @mgbennet!

PR #217, opened 2017-06-09
State size limit (#205)

mgbennet:
Per [this](https://github.com/Microsoft/BotFramework-Emulator/issues/205) issue, a configurable limit has been placed on state storage.  It is accessible from a new tab in the App Settings dialog.

### DIFF
--- a/src/client/dialogs/appSettingsDialog.tsx
+++ b/src/client/dialogs/appSettingsDialog.tsx
@@ -198,12 +198,10 @@ export class AppSettingsDialog extends React.Component<{}, AppSettingsDialogStat
                         {this.renderNavTab("state", (<div>
                             <div className='emu-dialog-text'>
                                 Bots use the <a href="https://aka.ms/sw9dcl">Bot State service</a> to store and retrieve application data. The Bot Framework's bot state service has a size limit of 64KB. Custom state services may differ.
-                                <br />
-                                A value of zero means no limit.
                             </div>
                             <div className="input-group">
                                 <label className="form-label">
-                                    Size limit (KB):
+                                    Size limit (KB - zero for no limit):
                                 </label>
                                 <input
                                     type="number"
@@ -211,6 +209,7 @@ export class AppSettingsDialog extends React.Component<{}, AppSettingsDialogStat
                                     ref={ref => this.stateSizeLimitInputRef = ref}
                                     className="form-input appsettings-number-input appsettings-space-input"
                                     min={0}
+                                    max={4000000}
                                     defaultValue={String(serverSettings.framework.stateSizeLimit) || '64'} />
                             </div>
                         </div>) )}
--- a/src/client/dialogs/appSettingsDialog.tsx
+++ b/src/client/dialogs/appSettingsDialog.tsx
@@ -42,16 +42,29 @@ import * as path from 'path';
 import * as Constants from '../constants';
 
 
-interface IAppSettings {
+interface AppSettingsDialogState extends React.Props<AppSettingsDialog> {
+    curTab?: string,
     ngrokPath?: string,
-    bypassNgrokLocalhost?: boolean
 }
 
-export class AppSettingsDialog extends React.Component<{}, { ngrokPath: string }> {
+export class AppSettingsDialog extends React.Component<{}, AppSettingsDialogState> {
     settingsUnsubscribe: any;
     ngrokPathInputRef: any;
+    stateSizeLimitInputRef: any;
     bypassNgrokLocalhostInputRef: any;
     showing: boolean;
+
+    constructor(props) {
+        super(props);
+        let serverSettings = getSettings().serverSettings;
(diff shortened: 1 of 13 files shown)


src/client/dialogs/appSettingsDialog.tsx
```diff
@@ -91,7 +105,7 @@ export class AppSettingsDialog extends React.Component<{}, { ngrokPath: string }
             if (filenames && filenames.length) {
                 // TODO: validate selection
                 this.ngrokPathInputRef.value = filenames[0];
-                this.setState({ ngrokPath: filenames[0] });
+                this.setState({ ngrokPath: filenames[0] })
```
eanders-MS: Did you have something against that semi-colon? :)

src/server/controllers/emulator/emulatorController.ts
```diff
@@ -248,7 +248,7 @@ export class EmulatorController {
         windowManager.zoomIn();
     }
     static zoomOut = () => {
-        windowManager.zoomIn();
+        windowManager.zoomOut();
```
eanders-MS: Weird, I thought this was already in master.

src/utils.ts
```diff
@@ -126,4 +126,28 @@ export const safeStringify = (o: any, space: string | number = undefined): strin
         }
         return value;
     }, space);
+}
+
+export const approximateObjectSize = (object: any) => {
```
eanders-MS: What does this return if object is a string or a number?

src/server/controllers/emulator/emulatorController.ts
eanders-MS: It is in master. What is this comparing against?

mgbennet: Yeah, both changes in the last push I made are showing up here, even though they're in master.

src/utils.ts
mgbennet: Strings return 2*(the string length), which is correct, but it does it in a really ineffecient manner. With numbers and booleans it returns zero, which is incorrect. I could fix this pretty easily with a little refactor of the function structure.

src/server/controllers/emulator/emulatorController.ts
eanders-MS: Maybe try merging master into your branch?

src/utils.ts
eanders-MS: I think you will need to protect against infinite recursion:
```javascript
var obj = { s: "hello" }
obj.t = obj;
approximateObjectSize(obj);
```

eanders-MS: Take a look at how we do that in safeStringify

eanders-MS:
Changes look good, I will go ahead and merge them, but I have one change request: to move "KB" out of the label and put it next to the number input box:
image
Thanks!

src/utils.ts
```diff
@@ -126,4 +126,31 @@ export const safeStringify = (o: any, space: string | number = undefined): strin
...
+            return 8;
+        case 'string':
+            return object.length * 2;
+        case 'object':
+            let bytes = 0;
+            for (let i in object) {
+                let value = object[i];
+                if (typeof value === 'object' && value !== null) {
+                    if (cache.indexOf(value) !== -1) {
+                        return;
```
eanders-MS: Oh there's a bug right here. No return value. Should return 0?

eanders-MS: Actually, I think this should be a break, not a return.

eanders-MS: Er, make that a continue. Not a break. Not a return. A `continue`.

AndreMantas:
Will this size be configurable in dev.botframework.com for each bot?
I'm having lots of problems with this state size (Error 400 BadRequest in my bot code) and I'm changing my code to fix the problem but it would be really nice to have a bigger limit.

eanders-MS:
State size limit in the Bot Framework is fixed for all bots at 64kb. If this doesn't work for your bot, you can implement your own state service using the BotBuilder-Azure extensions: https://github.com/Microsoft/BotBuilder-Azure. I suggest taking a look at the Samples folder below the CSharp or Node folder (according to your language preference).

More about the state service: https://docs.microsoft.com/bot-framework/dotnet/bot-builder-dotnet-state

AndreMantas:
So in my dialog should I delete all my private fields that I suspect are big and use the Get/Set UserData methods to save and load those big objects?

eanders-MS:
I don't think it will help. All fields are stored in a single data blob, and the max size of that blob is fixed at 64kb. If you cannot pare down your overall state size, I recommend implementing your own state service.

AndreMantas:
So my best bet is to use Azure DocumentDB to store the bot state?

eanders-MS:
Yes, I think that's the right direction, though I haven't done it myself so I don't have much more information than that. If you need more guidance or discussion, you should be able to find it by asking over at https://github.com/Microsoft/BotBuilder/issues or on Stack Overflow (tag with bot framework).

PR #272, opened 2017-08-16
Split panes takes numbers, not strings, as props

mgbennet:
Also added a maxSize for both panes, so you can't inadvertently hide one pane entirely. This can still happen if the window is resized, though.

### DIFF
--- a/src/client/mainView.tsx
+++ b/src/client/mainView.tsx
@@ -232,12 +232,12 @@ export class MainView extends React.Component<{}, {}> {
         }
     }
 
-    initBotChatContainerRef(ref, initialWidth) {
+    initBotChatContainerRef(ref, initialWidth:number) {
         this.botChatContainer = ref;
         this.updateBotChatContainerCSS(initialWidth);
     }
 
-    botChatComponent(initialWidth) {
+    botChatComponent(initialWidth:number) {
         if (this.directline) {
             const settings = getSettings();
             const srvSettings = new ServerSettings(settings.serverSettings);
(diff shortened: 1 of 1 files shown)


eanders-MS:
react-split-pane can take a string or a number for minSize. From the [source](https://github.com/tomkp/react-split-pane/blob/48531c514232104d03329cfebaa4e059489a2d1e/src/SplitPane.js#L269):
```js
minSize: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```

Maybe the issue is the unit specification on the string value ("px")?

src/client/mainView.tsx
```diff
@@ -304,24 +304,18 @@ export class MainView extends React.Component<{}, {}> {
...
-        if (typeof settings.layout.horizSplit === typeof Number) {
-            horizSplit = `${settings.layout.horizSplit}px`;
-        }
+        let vertSplit = Number(settings.layout.vertSplit);
+        let horizSplit = Number(settings.layout.horizSplit);
         return (
             <div className='mainview'>
                 <div className='botchat-container'>
-                    <Splitter split="vertical" minSize="200px" defaultSize={vertSplit} primary="second" onChange={(size) => this.verticalSplitChange(size)}>
+                    <Splitter split="vertical" minSize={200} maxSize={-200} defaultSize={vertSplit} primary="second" onChange={(size) => this.verticalSplitChange(size)}>
```
eanders-MS: It isn't at all clear to me why you'd specify -200 for `maxSize`. What is the reason?

mgbennet: A negative number means the splitter can move to within that number of pixels from the border, basically setting a minimum width for the secondary panel. 200 was to match the other panel's minsize, and seemed a decent size given the content of the first panel.

mgbennet:
You're right, they can be strings but the addition of the 'px' breaks it. Additionally, maxSize doesn't work if it's both a string and a negative number, so to use that functionality we need numbers.

eanders-MS:
Thanks!

PR #276, opened 2017-08-24
Adding support for bot framework authentication endpoints v3.2

Jeffders:
This change does a few things:
1. Adds support for the Bot Framework v3.2 authentication endpoints
2. Performs a fallback to auth endpoints v3.1 if v3.2 does not work for the person
3. Does a check for the latest SDK versions and requests that the user update is the most recent version is not being used.


src/server/versionManager.ts
```diff
@@ -32,8 +32,12 @@
 //
 
 import * as log from './log';
+import * as request from 'request';
+import * as http from 'http';
+const DOMParser = require('xmldom').DOMParser;
```
eanders-MS: xmldom should be added to the dependencies section of package.json.

src/server/versionManager.ts
```diff
@@ -80,15 +152,11 @@ export class VersionManager {
...
-                        if (parts.length >= 3) {
-                            let version = {
-                                major: parseInt(parts[0]),
-                                minor: parseInt(parts[1]),
-                                subminor: parseInt(parts[2])
-                            }
-                            return version;
+                        let version = VersionManager.parseVersion(versionString);
+                        if (version) {
+                            version.type = i === 0 ? 'dotnet' : 'node';
```
eanders-MS: What about Python? May not be safe to fallback to 'node'.

src/server/versionManager.ts
eanders-MS: > const DOMParser = require('xmldom').DOMParser;

Possible to use TypeScript here? e.g.: `import {DOMParser} from 'xmldom';`

src/server/versionManager.ts
Jeffders: This is true, but we don't have that 'officially' yet...but I think we should add this as a separate change soon.

src/server/versionManager.ts
Jeffders: Done!

src/server/versionManager.ts
```diff
@@ -42,34 +46,102 @@ interface IVersion {
...
                 if (version) {
-                    log.warn('Warning: The Bot is using SDK version ' + VersionManager.toString(version) + '.');
+                    log.warn('Warning: The bot is using SDK version ' + VersionManager.toString(version) + '.');
                 } else {
-                    log.warn('Warning: The Bot is using an SDK version earlier than 3.5.0.');
+                    log.warn('Warning: The bot is using an SDK version earlier than ' + VersionManager.toString(VersionManager.minimumVersion) + '.');
                 }
-                log.warn('Warning: SDK versions earlier than 3.5.0 use an authentication configuration that will no longer work after July 31, 2017.');
-                log.warn('Warning: For your bot to continue working beyond this date, please update your bot to use an SDK version greater than or equal to 3.5.0.');
+                log.warn('Warning: SDK versions earlier than ' + VersionManager.toString(VersionManager.minimumVersion) + ' use an authentication configuration that will no longer work with the emulator after September 30, 2017.');
```
eanders-MS: Is the September 30 date still correct?

PR #278, opened 2017-09-02
Fix broken unit test in isSecureUrl, and add types to util.ts functions.

mgbennet:
Adding unit tests revealed a flaw in utils.isSecureUrl, which this fixes. Also added some types to the other util functions, just for good measure.

### DIFF
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -110,7 +110,7 @@ export const isLocalhostUrl = (urlStr: string): boolean => {
 
 export const isSecuretUrl = (urlStr: string): boolean => {
     const parsedUrl = url.parse(urlStr);
-    return (parsedUrl.protocol !== null && parsedUrl.protocol.startsWith('https'));
+    return (!!parsedUrl.protocol && parsedUrl.protocol.startsWith('https'));
 }
 
 export const safeStringify = (o: any, space: string | number = undefined): string => {
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -72,7 +72,7 @@ export const loadSettings = <T>(filename: string, defaultSettings: T): T => {
 /**
  * Save JSON object to file.
  */
-export const saveSettings = <T>(filename: string, settings: T) => {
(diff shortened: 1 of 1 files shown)


src/shared/utils.ts
```diff
@@ -81,37 +81,36 @@ export const saveSettings = <T>(filename: string, settings: T) => {
...
+export const isLocalhostUrl = (urlStr: string): boolean => {
     const parsedUrl = url.parse(urlStr);
     return (parsedUrl.hostname === 'localhost' || parsedUrl.hostname === '127.0.0.1');
 }
 
-export const isSecuretUrl = (urlStr: string) => {
+export const isSecuretUrl = (urlStr: string): boolean => {
     const parsedUrl = url.parse(urlStr);
-    return (parsedUrl.protocol.startsWith('https'));
+    return (parsedUrl.protocol !== null && parsedUrl.protocol.startsWith('https'));
```
eanders-MS: Better to avoid explicit comparisons when checking for falsehood. Prefer this:
```js
    return (parsedUrl.protocol && parsedUrl.protocol.startsWith('https'));
```

mgbennet: That would be fine if this was an if statement, but we're actually returning the value. [In an AND operator, the first half is returned if it can be converted to false](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators). Thus, if we pass a non-url string, protocol gets set to null, which we end up returning.

eanders-MS: Oh yikes. JavaScript gotcha right there.

eanders-MS: Still, I don't love it. How about reversing the logic:
```js
if (parsedUrl.protocol) {
    return parsedUrl.protocol.startsWith('https');
} else {
    return false;
}
```

eanders-MS: Oh, or maybe you could use `!!`:
```js
    return (!!parsed.protocol && parsedUrl.protocol.startsWith('https');
```

mgbennet: That works perfectly! Good old type coercion.

PR #288, opened 2017-09-14
fix command to run test bot to use ts-node

mgbennet:
Since the testbot is written in typescript, we need to use ts-node to run it instead of plain old node


eanders-MS:
Oh interesting. No compile step for testbot?

mgbennet:
ts-node handles the compiling for you, letting you just run it.

Issue #101, opened 2017-02-10
Cannot emulate a specific user and/or conversation


Raseru:
At the Microsoft emulator download page, it says you can "Emulate a specific user and/or conversation"

This newer emulator seems to have lost this feature.

eanders-MS:
Thank you for bringing this up. You are correct, the new emulator does not yet support these features. The ability to switch between users in a group conversation is coming soon. The ability to switch between conversations is also planned.

navzam:
Any update on this for the new emulator? In the previous version, I got around this by modifying `server.json` manually, but even that doesn't seem to work in the new version.

tonyanziano:
Hi @navzam ,

We do have a work item tracking this #522 , but it is currently in our backlog and has not yet been assigned to a milestone, although that may change.

vishwacsena:
dupe of #197

PR #303, opened 2017-10-11
Add user id change to the server settings

droidpl:
Adds the user Id setting to the server settings so we can modify this id. I think this can be quite helpful while the new implementation for it is getting ready with the stores

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,7 +1,7 @@
 {
     // See https://go.microsoft.com/fwlink/?LinkId=733558
     // for the documentation about the tasks.json format
-    "version": "0.1.0",
+    "version": "2.0.0",
     "command": "gulp",
     "isShellCommand": true,
     "args": [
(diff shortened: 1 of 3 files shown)


eanders-MS:
I sincerely appreciate you taking the time to implement what is one of our most asked for features. Thank you for that. We have a design for this feature that I'm afraid goes in a different direction than the way you've implemented userId here. Because of this, it pains me to say that I'm unable to accept your changes. To give an idea of how we're planning to support userId, here's an early comp of it:

emulator_sidebar

You will be able to add and remove users in the conversation on a per-bot basis, with the ability to edit names and userIds. You'll also be able to switch which user you are in the conversation.

Our team is occupied with other development work currently, but when we are able to turn attention back to the emulator, this will be among the first features we tackle.

Raseru:
Its been over 8 months since I brought this up, saying it was coming soon: https://github.com/Microsoft/BotFramework-Emulator/issues/101

Thanks droidpl for actually fixing this.

chakrihacker:
Please merge this

Issue #304, opened 2017-10-11
Suggestion: Use Emulator with Hotkeys


rinormaloku:
In the browser, we already have a lot of hotkeys that we are used to, e.g.:

* F6 for the address bar.
* Alt+F for the Menu (three vertical dots icon)
* F5 to refresh the conversation.

Working a lot with the bot lately and it gets tedious to use the mouse all the time.
If you agree and or have your suggestions to add, I could start to work on it immediately.

eanders-MS:
That would be fantastic! My suggestions:

On Windows / Linux:
* F5 / Ctrl+R - Refresh the conversation
* F10 / Alt+F - Open the menu
* F6 / Ctrl+L - Focus the address bar

On Mac, I think it would be sufficient to replace Ctrl with ⌘.

We already have Ctrl++/-/0 for zoom level. You should be able to pattern your changes after the implementation of those shortcuts.

Thanks. Looking forward to the PR.

mgbennet:
Related to #292, another request for keyboard short cuts.

PR #308, opened 2017-10-15
Adding support for hotkeys in Emulator (#304)

rinormaloku:
Support added for following hotkeys:

* F5 / Ctrl+R - Refresh the conversation.
* F10 / Alt+F - Open the menu.
* F6 / Ctrl+L - Focus the address bar.

The same applies for Mac using ⌘ instead of Ctrl.

### DIFF
--- a/src/client/addressBar/addressBarMenu.tsx
+++ b/src/client/addressBar/addressBarMenu.tsx
@@ -32,8 +32,8 @@
 //
 
 import * as React from 'react';
-import { getSettings } from '../settings';
-import { AddressBarActions, ConversationActions } from '../reducers';
+import { getSettings, Settings, addSettingsListener  } from '../settings';
+import { AddressBarActions, ConversationActions, HotkeyActions } from '../reducers';
 import { Emulator } from '../emulator';
 import * as Constants from '../constants';
 import { remote } from 'electron';
@@ -41,8 +41,23 @@ import { remote } from 'electron';
 const { Menu } = remote;
 
 export class AddressBarMenu extends React.Component<{}, {}> {
+    addressBarMenu: any;
+    settingsUnsubscribe: any;
 
-    showMenu() {
+    componentWillMount() {
+        this.settingsUnsubscribe = addSettingsListener((settings: Settings) => {
+            if (settings.hotkey.openMenu) {
+                HotkeyActions.clearOpenMenu();
+                this.showMenuAtCoordinates();
+            }
+        });
+    }
+
+    componentWillUnmount() {
+        this.settingsUnsubscribe();
+    }
+
+    showMenu(options?: any) {
         const settings = getSettings();
         const inConversation = ((settings.serverSettings.activeBot || '').length > 0 && (settings.conversation.conversationId || '').length > 0);
         const haveActiveBot = (settings.serverSettings.activeBot || '').length > 0;
@@ -211,13 +226,37 @@ export class AddressBarMenu extends React.Component<{}, {}> {
         ];
 
         const menu = Menu.buildFromTemplate(template);
-        menu.popup();
+        menu.popup(undefined, options);
+    }
+
+    showMenuAtCoordinates() {
+        this.showMenu(this.getOptionsWithCoordinates());
+    }
+
+    getOptionsWithCoordinates() {
+        let zoomLevel = getSettings().serverSettings.windowState.zoomLevel;
+        let zoomRatio = this.getZoomRatio(zoomLevel);
+
+        return this.calculateCoordinatesAccordingToZoomLevel(zoomRatio);
+    }
+
+    calculateCoordinatesAccordingToZoomLevel(zoomLevelRatio: any) {
+        let rect = this.addressBarMenu.getBoundingClientRect();
+        let left = Math.ceil((rect.left + rect.width / 2) * zoomLevelRatio);
+        let top = Math.ceil((rect.top + rect.height / 2) * zoomLevelRatio);
+        return {x: left, y: top};
+    }
+
+    getZoomRatio(zoomLevel: number): number {
+        return 0.4819 * Math.exp(0.1824 * (zoomLevel + 4));
     }
 
     render() {
         return (
             <a className='undecorated-text' href='javascript:void(0)' title='Settings'>
-                <div className="addressbar-menu" dangerouslySetInnerHTML={{ __html: Constants.hamburgerIcon('toolbar-button', 24) }} onClick={() => this.showMenu()} />
+                <div className="addressbar-menu"
+                     ref={ref => this.addressBarMenu = ref}
+                     dangerouslySetInnerHTML={{ __html: Constants.hamburgerIcon('toolbar-button', 24) }} onClick={() => this.showMenu()} />
             </a>
         );
     }
--- a/src/client/addressBar/addressBarTextBox.tsx
+++ b/src/client/addressBar/addressBarTextBox.tsx
@@ -41,6 +41,7 @@ import { AddressBarOperators } from './addressBarOperators';
 export class AddressBarTextBox extends React.Component<{}, {}> {
     settingsUnsubscribe: any;
     textBoxRef: any;
+    hasFocus: boolean;
 
     onChange(text: string) {
         text = text || '';
@@ -66,7 +67,18 @@ export class AddressBarTextBox extends React.Component<{}, {}> {
         }
     }
 
-     onFocus() {
+    onFocus() {
+        this.hasFocus = true;
+        this.focusAddressBarAndSelectText();
+        AddressBarActions.gainFocus();
+    }
+
+    onBlur() {
+        this.hasFocus = false;
+        AddressBarActions.loseFocus();
+    }
+
+    focusAddressBarAndSelectText() {
         this.textBoxRef.select();
         const settings = getSettings();
         const bots = AddressBarOperators.getMatchingBots(settings.addressBar.text, null);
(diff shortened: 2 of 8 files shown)


eanders-MS:
Thank you for this, but I am sorry to say that it needs to be done another way. These hotkeys should be implemented as accelerators on menu items. It can be done using Electron's accelerator support. No need for mousetrap. Search the code for 'show-about'. It is an accelerator to show the About Dialog. Follow its implementation. Let me know if you need further clarification. Also, let me know if you are out of time to work on this :)

rinormaloku:
@eanders-MS I updated the commit by using the pattern that 'show-about' uses.
Suggestions are very welcome (New to React and Redux)

src/client/addressBar/addressBarTextBox.tsx
```diff
@@ -95,6 +95,10 @@ export class AddressBarTextBox extends React.Component<{}, {}> {
 
     componentWillMount() {
         this.settingsUnsubscribe = addSettingsListener((settings) => {
+            if (settings.hotkey.focusAddressBar) {
+                settings.hotkey.focusAddressBar = false;
```
eanders-MS: `settings` is immutable. It can only be set in a reducer. A total pain, I know. Let me think about how to do this another way.

rinormaloku: We can set the other values to false e.g. `Object.assign({}, state, { openMenu: true, focusAddressBar: false, refreshConversation: false });`

src/server/main.ts
```diff
@@ -201,6 +201,19 @@ const createMainWindow = () => {
         windowManager.zoomTo(0);
     });
 
+    let registerHotkeys = (hotkeys, callback) => hotkeys.forEach(hotkey =>
+        Electron.globalShortcut.register(hotkey, callback));
```
eanders-MS: I recently became aware of [this Electron issue](https://github.com/electron/electron/issues/1334), where globally registered shortcuts trigger regardless of whether the app is in focus. I think we need to avoid that, otherwise, when you refreshed a browser window, it would also refresh your bot conversation in the background. A viable workaround would be to take a dependency on 'electron-localshortcut', which circumvents this issue. We'll have to do this for the Zoom shortcuts as well (we'll take care of that in a separate PR. @mgbennet for visibility)

src/client/addressBar/addressBarTextBox.tsx
eanders-MS: It's a guiding principle of Redux that the state object is immutable. Redux works by returning a new state object each time a change is affected in a reducer. To change the state locally creates an inconsistency between the locally changed state and the actual application state. The flag you're resetting isn't getting reset in the authoritative version of the state.

In fact, you can see it happen:

1. in src/server/main.ts, comment out the last line: `require('electron-debug')();`, so that you can test the F5 shortcut without reloading the app, then recompile and run the emulator.
2. connect to a bot
3. press F5 to refresh the conversation
4. say something in the conversation
5. notice the conversation refreshes again!

I think it can be fixed by clearing the flag via another reducer action, like `HotKey_FocusAddressBar_Clear`.

rinormaloku: Thank you very much for taking the time to share the knowledge. :)

src/server/main.ts
rinormaloku: Implemented electron-localshortcut. And tested it. Works like a charm.

src/client/addressBar/addressBarTextBox.tsx
eanders-MS: It's getting closer! I noticed two bugs:
* **Menu Position vs. Zoom Level**: When the zoom level is not the default level, the menu doesn't appear in the correct location when opened using the hotkey. Could you position it so that the top-left corner is in the center of the hamburger icon?
* **Double Conversation Refresh**: After I send a bot a message, pressing F5 once doesn't refresh the conversation. If pressed a second time, it will refresh. After the refresh, if I say something to the bot, my bot will reply but then the conversation will unexpectedly refresh again.

Getting into more fine-grained feedback on the behavior:
* Some hotkeys should toggle the state, e.g.: When the address bar is focused, pressing F6 should unfocus it. This means some hotkeys are "tri-state", having Unset, Focus, Blur states. I notice the menu hotkey already toggles, but am unsure why. It might be the behavior of `Menu.popup` to hide the menu if it is already showing. The Menu object does have a `closePopup` method.

###### Some tips for debugging client code

`Shift+Ctrl+I` will open up the Chrome debugger, which can be very useful to set breakpoints and stepping through client code, to see when the render calls happen, that the state looks like in reducers, etc.

If you have to comment out the inclusion of `electron-debug`, then you can uncomment the call to `mainWindow.webContents.openDevTools();` that appears earlier in main.ts. This will just always open the Chrome debugger.

Thank you again for taking this on!

eanders-MS: Er, I guess you wouldn't need a tri-state field actually, just toggle existing focus state. Sorry for the bad advice :)

rinormaloku: @eanders-MS I was checking to create a solution for *Menu Position vs Zoom Level* but when zoomed in/out the electron API reports the wrong ratio. [It is a known issue](https://github.com/electron/electron/issues/7458). 
Possible Solution: different ratio per zoom level? 
(Simulating click on the menu won't work as by default the menu appears wherever the mouse cursor is.)

rinormaloku: @eanders-MS 

Latest commit solves all issues:

* To implement toggling I moved changes up to the parent component AddressBar as it was affecting other components as well.
* Double Conversation Refresh: Clicking the button calls the newConversation method so instead the hotkey triggering a button click I called the newConversation directly.
* Menu Position: Solution doesn't make me happy. But couldn't come up or find any better solution [it is a bug, see the previous comment].

src/client/addressBar/addressBarTextBox.tsx
```diff
@@ -66,7 +67,23 @@ export class AddressBarTextBox extends React.Component<{}, {}> {
...
+    onFocus() {
+        this.hasFocus = true;
+        this.focusAddressBarAndSelectText();
+        AddressBarActions.gainFocus();
+    }
+
+    onBlur() {
+        this.hasFocus = false;
+        AddressBarActions.loseFocus();
+        //const settings = getSettings();
```
eanders-MS: If this is dead code, please remove.

src/client/addressBar/addressBarMenu.tsx
```diff
@@ -211,13 +226,43 @@ export class AddressBarMenu extends React.Component<{}, {}> {
...
+
+    calculateCoordinatesAccordingToZoomLevel(zoomLevelRatio: any) {
+        let rect = this.addressBarMenu.getBoundingClientRect();
+        let left = Math.ceil((rect.left + rect.width / 2) * zoomLevelRatio);
+        let top = Math.ceil((rect.top + rect.height / 2) * zoomLevelRatio);
+        return {x: left, y: top};
+    }
+
+    getZoomRatio(zoomLevel: number) {
+        let zoomInRatios = [1, 1.2, 1.437, 1.73, 2.07, 2.49, 2.99, 3.58, 4.3];
```
eanders-MS: Please describe how you calculated these values, or where they came from.

src/client/addressBar/addressBarTextBox.tsx
eanders-MS: This is looking good! Agreed the menu positioning is pretty rough with the fixed zoom ratios. I might rework the zoom system later to pull all the logic into some kind of zoom handler.

A few minor comments in the code, then we can merge it.

src/client/addressBar/addressBarMenu.tsx
rinormaloku: Added comment to it. The zoom in ratios are found very scientifically.
Binary Search and trial&error  :P . Tested it in two resolutions 1080p and 1600x900.

src/client/addressBar/addressBarTextBox.tsx
rinormaloku: done.

src/client/addressBar/addressBarTextBox.tsx
rinormaloku: Thanks for all the help along the way :)

src/client/addressBar/addressBarMenu.tsx
```diff
@@ -211,13 +226,45 @@ export class AddressBarMenu extends React.Component<{}, {}> {
...
+        let rect = this.addressBarMenu.getBoundingClientRect();
+        let left = Math.ceil((rect.left + rect.width / 2) * zoomLevelRatio);
+        let top = Math.ceil((rect.top + rect.height / 2) * zoomLevelRatio);
+        return {x: left, y: top};
+    }
+
+    getZoomRatio(zoomLevel: number) {
+        // Below zoom in and out ratios are found by trial and error.
+        // They fix the wrong coordinates reported on different zoom levels by Element.getBoundingClientRect().
+        let zoomInRatios = [1, 1.2, 1.437, 1.73, 2.07, 2.49, 2.99, 3.58, 4.3];
```
eanders-MS: OK, this was really bugging me so I came up with a mathematical approximation based on your data set:
```TypeScript
    getZoomRatio(zoomLevel: number): number {
        return 0.4819 * Math.exp(0.1824 * (zoomLevel + 4));
    }
```
Could you please make this change? One thing left to do is to make a constant for the `4` literal, but I can do that separately after your changes are in (since I'll want to thread it into the zoom in/out code as well).

src/client/addressBar/addressBarMenu.tsx
```diff
@@ -211,13 +226,37 @@ export class AddressBarMenu extends React.Component<{}, {}> {
...
+
+    calculateCoordinatesAccordingToZoomLevel(zoomLevelRatio: any) {
+        let rect = this.addressBarMenu.getBoundingClientRect();
+        let left = Math.ceil((rect.left + rect.width / 2) * zoomLevelRatio);
+        let top = Math.ceil((rect.top + rect.height / 2) * zoomLevelRatio);
+        return {x: left, y: top};
+    }
+
+    getZoomRatio(zoomLevel: number): number {
+        return 0.4819 * Math.exp(0.1824 * (zoomLevel + 4));
```
eanders-MS: When this is merged in I will comment it :)

eanders-MS:
@rinormaloku thank you for your contribution! It's been a pleasure working with you.

src/client/addressBar/addressBarMenu.tsx
rinormaloku: That was a piece of math 👍 
Btw. I wanted to make the 4 a constant but left it better to you as I was coming up with this name  REBALANCE_NEGATIVE_ZOOM_LEVEL :p

Issue #115, opened 2017-03-08
Emulator slows down with use


LarsLiden:
The emulator consistently starts running slower and slower the longer you use it.  Eventually it hangs and needs to be restarted.   It's evident that whenever I restart things run fast again.

McGern:
Hi,

I am getting the same issue. Am worried that because this is using the webchat interface code (assumption), that the webchat interface will also be suffering from the same memory problems.

LarsLiden:
Still verifying, but I think I have a cause.  I tried clearing the log and it sped up again.  The image button responded again too

LarsLiden:
Confirmed that this is definitely the Log that is causing the issue.  Speeds thing back up every time I clear it.

eanders-MS:
Ok, I've made a rendering optimization to the log view. Will be in the next release (due out very soon). The emulator has a more systemic issue of an ever-growing memory footprint. Will need to establish a method of paging out old data.

LarsLiden:
Hmm... Still seems to be an issue.  Clearing the log still improves it.  The performance issue is pretty severe when you compare running in the emulator vs running in Skype (or another channel) side by side.

eanders-MS:
Thanks for the update.

tonyanziano:
I believe this was fixed / addressed by PR #367. Can we close this?

cwhitten:
Closing for lack of activity

PR #367, opened 2017-12-15
Added performance test, improved performance of log view with many messages

mgbennet:
Related to #115 

Added simple spam performance test, which probably can be further improved, and improved the performance of the emulator when there are many entries in the log view by adding update logic to the entries.

In the current production build, after 255 messages it takes approximately 2.5x as long to send and receive a message from a bot than it does with a new session.  With the new code in this PR, the time stays approximately the same regardless of how many messages have been sent.

### DIFF
--- a/src/client/logView.tsx
+++ b/src/client/logView.tsx
@@ -59,54 +59,59 @@ interface ILogEntry {
     args: any
 }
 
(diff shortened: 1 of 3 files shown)


eanders-MS:
Thanks @mgbennet !

PR #368, opened 2017-12-16
Fix for log panel being minimized by default.

mgbennet:
Casting the settings as a Number resulted in dropping the percentage split. Numerous users have been confused by the log panel apparently "disappearing", though it is actually just minimized at the bottom of the right panel.

Setting the default values to numbers insures that on a fresh install the split panels are at reasonable positions. Changing the minimum size of the log panel insures two log entries will just be visible, helping existing users who might not realize their log is hidden.

Another thing we should consider doing is designing the draggable dividers so it is more obvious that they can be dragged, or just by increasing the clickable area so the mouse change on mouse over is more obvious.

### DIFF
--- a/src/client/mainView.tsx
+++ b/src/client/mainView.tsx
@@ -304,18 +304,20 @@ export class MainView extends React.Component<{}, {}> {
 
     render() {
         const settings = getSettings();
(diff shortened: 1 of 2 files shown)


eanders-MS:
Thanks @mgbennet !

Issue #314, opened 2017-10-23
Locale field cannot be cleared, will repopulate when erased


eanders-MS:
Try to delete the value in the locale field in the bot connect UI. Notice that immediately after the last character is deleted, the value is reset to the default ("en-US" on my machine)

eanders-MS:
Fixed by https://github.com/Microsoft/BotFramework-Emulator/pull/375

PR #374, opened 2017-12-19
Fixed bug #314 - locale field was resetting once empty.

Fignewton5:
- Bug was due to `locale` input field setting value based on an OR operator
  - so anytime the field reached an empty state, the left side of the OR operator would evaluate to `false` and default to the app's locale
- Added `en-US` as the default locale for the `default-bot`


eanders-MS:
Thanks @Fignewton5 !

PR #375, opened 2017-12-20
Added test for payment window to integration test.

mgbennet:
Added a way to bring up the payment window with the testbot, and added integration tests to test this functionality. Also update the BotBuilder package.


eanders-MS:
Thanks @mgbennet !

PR #377, opened 2017-12-22
Foundation for studio client v2

compulim:
* Add `.vscode/settings.json` for: 4 spaces tabs, empty space at EOF, no space at EOL
* `/src/client/` moved to `/client/v1/`
  * Built by Facebook's [`create-react-app`](https://npmjs.com/package/create-react-app) with [`react-scripts-ts`](https://www.npmjs.com/package/react-scripts-ts)
  * Client now compiled using TypeScript 2.5.3
* Host client using Webpack
  * Support hot reloading, file watching, etc (it is `create-react-app`/`react-scripts-ts` behind the scene)
  * Under `/client-v2/`, run `npm start`
  * Modify `/src/server/main.tsx` to load [http://localhost:3000/](http://localhost:3000/), instead of file:/.../index.html
* Isolating dependencies between Electron main and React
  * `redux`, `request`, and `rxjs` are loaded on both side

Caveats:
* Gulp to copy `/src/[shared,types]/**/*` and `*.css` to `/client-v2/src/external/`
  * `npm run build`/`npm start` at root still works, will run the new gulp scripts
  * To copy manually, under `/client-v2/`, run `gulp copy`
* `react-split-pane` seems broken on TypeScript >= 2.4.1
* Loading Electron-only package in React need to use `window['require']('electron')`, instead of `require('electron')`, this is to skip Webpack bundling
  * We also load `fs`, `mkdirp`, `path`, and `url` this way


compulim:
* Added "dev build" story: faster dev builds and simplified setup
  * Takes ~25s on first run, ~2s on code changes on React
  * `npm start` will run both Electron and React app [concurrently](https://npmjs.com/package/concurrently)
    * Electron app will load page code from [http://localhost:3000/](http://localhost:3000/)
  * `npm run build` will still produce production code on both Electron and React side, will still output to `/app/`

If you need to modify Electron code frequently, you can skip React code bundling by:
1. Run `npm run start:react-app`
2. Set `ELECTRON_TARGET_URL=http://localhost:3000/`
3. Run `gulp build-app`
4. Run `npm run start:electron`

> Repeat steps 3 and 4 after you made code changes to Electron side (~10s on each run).

> You still need `npm run build` for the very first time to bundle your React code.

eanders-MS:
Looks good, thanks!

PR #378, opened 2017-12-23
Studio shell

compulim:
Add shell and enclose the emulator

PR #380, opened 2017-12-27
Studio shell

compulim:
Resubmit of #378 with changes and base from `studio`.

Add shell and enclose the emulator


client-v2/package.json
```diff
@@ -5,17 +5,22 @@
...
     "react-dom": "^16.2.0",
-    "react-scripts-ts": "2.8.0",
+    "react-redux": "^5.0.6",
     "react-split-pane": "^0.1.74",
     "redux": "^3.7.2",
+    "redux-promise-middleware": "^5.0.0",
     "request": "^2.83.0",
-    "rxjs": "^5.5.6"
+    "rxjs": "^5.5.6",
+    "sockjs-client": "^1.1.4"
```
eanders-MS: We need to lock these versions. Is that what package-lock.json does? I've been locking them by pinning version numbers.

eanders-MS:
Did you mean to include client-v2/packages/react-dev-utils-4.2.1@ed5c48c.tgz?

eanders-MS:
This is looking really good! Would you expect the emulator to work right now? It seems I cannot connect to a bot. I get the following error:
> XMLHttpRequest cannot load http://localhost:5291/v3/directline/conversations. Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:3000' is therefore not allowed access. The response had HTTP status code 405.

compulim:
Yes, we need both `react-dev-utils-4.2.1@ed5c48c.tgz` and `react-scripts-ts-2.9.0@e404a66.tgz` for now because we are running on the edge (`@master`) now. When `react-scripts-ts@2.9.0` is out, we can drop both packages.

It should works now. Try add CORS to your bot to allow `http://localhost:3000`. We will be packaging everything using `file://` in production build.

You can also run Electron without `ELECTRON_TARGET_URL` env so it will run using bundles at `file://` (don't forget to run `npm run build` to build those static bundles).

eanders-MS:
It's the restify server in server/restServer.ts that needs CORS. You can repro by trying to connect to any bot endpoint (just need to get webchat to try to make a connection to the direct line server in the emulator's node process). We can merge this in, then I will add CORS support to the emulator side.

eanders-MS:
(All our unit tests are going to be invalid. That's fine for the time being)

eanders-MS:
@compulim and @Fignewton5, I've added you to the Bot Framework Emulator team. You should have merge permissions now.

client-v2/package.json
compulim: Yes, `package-lock.json` will lockdown versions for development, it technically same as `npm-shrinkwrap.json`.

When bumping version of packages, use `npm install` and commit `package-lock.json`. Don't modify `package.json/dependencies` by hand followed by `npm install`, this combo may break `package-lock.json`.

Also, don't forget to exclude our private NPM feed, otherwise, `package-lock.json` will instruct `npm` to get it from our private NPM. People without access to our NPM feed will break, even it is a public package. Private NPM passthru is a thing of the past, we should use @scoped packages by now.

eanders-MS:
@compulim F5 to run in vscode is not working for me in this branch. The app opens to a blank window. No error in the console. window.location is not set.

PR #381, opened 2017-12-28
Add CORS for localhost debugging

eanders-MS:
Also support attach to existing emulator process


eanders-MS:
More accurately: Add CORS for localhost debugging with webpack dev server

PR #385, opened 2017-12-29
Added new splitter control. Setup basic botchat layout.

eanders-MS:
* I inlined the react-splitter-layout package into our codebase and made some modifications. We'll keep a dependency on the package so that the license appears in our credits, and we'll credit the package in source comments. It will replace react-split-pane.

* Added new basic layout for botchat editor.


client-v2/public/css/splitter.css
```diff
@@ -0,0 +1,53 @@
+.splitter-layout {
+    display: flex;
+    flex-direction: row;
+    height: 100%;
+    overflow: hidden;
+    position: absolute;
+    width: 100%;
+}
+
+.splitter-layout .layout-pane {
```
compulim: Consider using `>` (direct descendant) to prevent leaking CSS styles.

client-v2/src/ui/layout/splitter/pane.js
```diff
@@ -0,0 +1,43 @@
...
+        let classes = 'layout-pane';
+        const style = {};
+        if (!this.props.primary) {
+            if (this.props.vertical) {
+                style.height = `${size}${unit}`;
+            } else {
+                style.width = `${size}${unit}`;
+            }
+        } else {
+            classes += ' layout-pane-primary';
```
compulim: Use [classnames](npmjs.com/packages/classnames) instead of string concat. We have already included the package in `/client-v2/`.

```js
import classNames from 'classnames';

<div className={ classNames('layout-pane', { 'layout-pane-primary': this.props.primary }) }>
```

client-v2/src/ui/layout/splitter/pane.js
```diff
@@ -0,0 +1,43 @@
...
+        );
+    }
+}
+
+Pane.propTypes = {
+    vertical: PropTypes.bool,
+    primary: PropTypes.bool,
+    size: PropTypes.number,
+    percentage: PropTypes.bool,
+    children: PropTypes.oneOfType([
```
compulim: Do we need to add `children` to `propTypes`?

client-v2/src/ui/layout/splitter/index.js
```diff
@@ -0,0 +1,235 @@
+// This is an inlined version of the react-splitter-layout npm package
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Pane from './pane';
+
+
+function clearSelection() {
```
compulim: Can we use CSS [`user-select`](https://caniuse.com/#search=user-select) instead?

client-v2/public/css/splitter.css
eanders-MS: This is code from react-splitter-layout package. I will be incrementally changing it in future PRs. But your point stands. I've already been bitten by a CSS leak from this very class!

client-v2/src/ui/layout/splitter/index.js
eanders-MS: This is react-splitter-layout code. I'll be incrementally changing it in future PRs.

client-v2/src/ui/layout/splitter/pane.js
eanders-MS: More react-splitter-layout code.

client-v2/src/ui/layout/splitter/pane.js
eanders-MS: More react-splitter-layout code. We can fix it later.

client-v2/src/ui/editor/botChatEditor/detailPanel.js
```diff
@@ -47,7 +47,7 @@ export default class DetailPanel extends React.Component {
         return (
             <div className={ CSS }>
                 <Panel>
-                    <Detail/>
+                    <Detail />
```
eanders-MS: It's better to communicate these changes to me rather than make them yourself. By agreeing to the change, we establish a coding standard. Plus, you probably don't want to run around behind me fixing everything!

client-v2/public/css/splitter.css
eanders-MS: I would prefer that you rolled back these changes. I want to keep this file as close to its original as possible in order to minimize diffs, in case I need to update to take a fix from the original project.

client-v2/src/ui/editor/botChatEditor/panel.js
```diff
@@ -36,11 +36,11 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 const CSS = css({
-    flex: 1,
-    position: 'relative',
     display: 'flex',
+    flex: 1,
```
eanders-MS: When you make changes like this, please explain your reason. Also, better to just state what should happen here and why, and let me take care of it.

PR #388, opened 2018-01-03
Toanzian/studio merge

Fignewton5:
- Added card editor components
  - Monaco JSON editor
  - Live card preview
  - Card template entities window (WIP)
  - Output window that displays card preview actions

**NOTE**: The card editor currently uses a card reducer that controls the state for the current card being edited. I assume this will change to probably just loading the card state from the selected document out of state.editor.documents, but it's just temporary. The structure of the card state that I'm using is also a WIP, as it will probably need to hold a path key as well that points to the card's location in the file system.


client-v2/src/data/reducer/card.js
```diff
@@ -0,0 +1,129 @@
+import {
```
eanders-MS: This should appear below the copyright notice

client-v2/src/data/reducer/editor.js
```diff
@@ -36,9 +36,11 @@ import * as constants from '../../constants';
 const DEFAULT_STATE = {
     documents: [{
         contentType: constants.ContentType_BotChat,
+        title: "My bot",
```
eanders-MS: `title` should not appear here. The source of the title depends on the asset type. For cards, cardTab.js will have knowledge of how to read and display the card's title.

client-v2/src/index.tsx
```diff
@@ -61,4 +61,4 @@ ReactDOM.render(
     document.getElementById('root')
 );
 
-registerServiceWorker();
+// registerServiceWorker();
```
eanders-MS: Unrelated change?

client-v2/src/v1/mainView.tsx
```diff
@@ -325,11 +325,11 @@ export class MainView extends React.Component<{}, {}> {
                                 onChange: (size: number) => this.verticalSplitChange(size)
                             },
                             [
-                                <div className='fill-parent'>
+                                <div key={ "botchat" } className='fill-parent'>
```
eanders-MS: Unrelated change?

.vscode/settings.json
```diff
@@ -5,7 +5,7 @@
   "files.trimTrailingWhitespace": true,
   "files.exclude": {
     "client-v2/build": true,
-    "client-v2/public/external": true,
+    "client-v2/public/external": false,
```
eanders-MS: Unrelated change?

client-v2/src/data/action/cardActions.js
```diff
@@ -0,0 +1,23 @@
+export const CARD_UPDATE_JSON = "CARD_UPDATE_JSON";
+export const CARD_ADD_OUTPUT_MSG = "CARD_ADD_OUTPUT_MSG";
+export const CARD_CLEAR_OUTPUT_WINDOW = "CARD_CLEAR_OUTPUT_WINDOW";
+
+export function updateCardJson(json) {
+    return {
+        type: CARD_UPDATE_JSON,
+        state: { json: json }
```
eanders-MS: Rename `state` to `payload` to make it a standard [FSA](https://github.com/acdlite/flux-standard-action). (True for each action here)

client-v2/src/data/action/cardActions.js
```diff
@@ -0,0 +1,23 @@
+export const CARD_UPDATE_JSON = "CARD_UPDATE_JSON";
```
eanders-MS: This file needs the copyright notice.

client-v2/src/data/reducer/card.js
```diff
@@ -0,0 +1,129 @@
...
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+const DEFAULT_STATE = {
+    title: "My card",
```
eanders-MS: CardTab component should read it's tab label from this field (rather than from the document object in the store).

client-v2/src/index.tsx
eanders-MS: Probably a fine change to make though. @compulim thoughts?

client-v2/src/ui/editor/cardEditor/cardJsonEditor/index.js
```diff
@@ -0,0 +1,236 @@
...
+
+import { connect } from 'react-redux'
+import { css } from 'glamor'
+import React from 'react'
+import ReactDOM from 'react-dom';
+import * as CardActions from '../../../../data/action/cardActions';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { updateCardJson } from '../../../../data/action/cardActions';
+const fs = typeof window !== 'undefined' ? window.require('fs') : require('fs');
```
eanders-MS: I think a better pattern here is to break this into two lines:
```js
if (typeof window !== 'undefined') { require = window['require']; }
const fs = require('fs');
```
This way, it fixes `require` for all calls to it within the file.

client-v2/src/ui/editor/cardEditor/cardJsonEditor/index.js
```diff
@@ -0,0 +1,236 @@
...
+
+const CSS = css({
```
eanders-MS: File needs copyright header.

client-v2/src/v1/mainView.tsx
eanders-MS: (probably fine.. it's all getting reimplemented anyway)

client-v2/src/index.tsx
Fignewton5: This was suggested by @compulim 👍

client-v2/src/v1/mainView.tsx
Fignewton5: These were just to suppress the annoying React warnings for arrayed elements being rendered without unique keys. I can remove if necessary

client-v2/src/data/action/cardActions.js
Fignewton5: Was completely unaware of this standard. Good call 👍

client-v2/src/data/reducer/editor.js
Fignewton5: Ok, so will the asset type's structure populate the `content` object? I'm just trying to think about how the `cardExplorer.js` will know about the card's title. Will it look for the title in `document.content.title` in the case of a card?

client-v2/src/ui/shell/explorer/cardExplorer.js
```diff
@@ -54,19 +55,34 @@ const BOTS_CSS = css({
...
+                        </ExpandCollapseContent>
+                    </ExpandCollapse>
+                </li>
+            </ul>
+        );
+    }
+}
+
+export default connect(state => ({
+    cards: state.editor.documents.filter(doc => doc.contentType === constants.ContentType_Card)
```
eanders-MS: Issue identified by @compulim:
> This call to Array.filter() will create a new documents array every time connect is called. Two things to do, either:
> 1.	Just pass documents inside the React component and process it there (in render or in willReceiveProps -> setState)
> 2.	Use reselect or any memorize function, n=1 memoizer is okay

I chose to just pass the documents, but not sure it isn't still making a copy in my render function.

PR #390, opened 2018-01-04
Address @compulim's PR feedback

eanders-MS:


client-v2/src/ui/shell/explorer/conversationExplorer.js
```diff
@@ -91,7 +91,7 @@ class ConversationExplorer extends React.Component {
                         <ExpandCollapseContent>
                             <ul className={ BOTS_CSS }>
                             {
-                                this.props.documents.map(doc => {
+                                this.props.documents.filter(doc => doc.contentType === constants.ContentType_Converation).map(doc => {
```
eanders-MS: Not sure this is sufficient either. @compulim what do you think?

PR #392, opened 2018-01-04
Two Redux store

compulim:
> Will add README.md on how to add code, still playing around with the idea and looks pretty good

Two Redux store, one on each side (main/renderer). Their connection is thru Action-over-WS-over-IPC. On one side, when we dispatch an action with `{ meta: { send: true } }`. Reducer/saga on the other side will receive this action for further processing.

The code may looks comprehensive at first, but it promotes two things:
* Separate native calls (`require('electron')` or `require('fs')`) into main, makes native boundary clear and the app portable
* Less logic code resides in React components

We are planning for two communication models:
* Challenge/response
  * Renderer ask a question, e.g. `SERVER/PING`
  * Main response with an answer, e.g. `SERVER/ALIVE`
* Publisher-subscriber
  * One side keep pushing updates (as action) to the other side
  * Action could be delta or the full copy of data, we still need to explore in this region

Notes:
* I see conflicting `restify@5.1.0` on `dependencies` and `restify@^4.3.2` on `devDependencies`
  * I am taking `restify@^4.3.2` on `devDependencies` because the commit is newer


.vscode/launch.json
```diff
@@ -32,10 +32,27 @@
             "protocol": "inspector"
         },
         {
+            "name": "Launch with localhost:3000",
             "type": "node",
-            "request": "attach",
-            "name": "Attach",
-            "port": 7777
```
eanders-MS: I need the Attach configuration in order to attach to the NodeJS process when run from `npm start`

client-v2/package-lock.json
```diff
@@ -95,6 +95,11 @@
         "acorn": "4.0.13"
       }
     },
+    "adaptivecards": {
```
eanders-MS: How did you regenerate package-lock.json???

.vscode/launch.json
compulim: I saw it appeared twice in [`launch.json`](https://github.com/Microsoft/BotFramework-Emulator/blob/studio/.vscode/launch.json) and removed one of them.

eanders-MS: Oh I see. GitHub's diff view hid that. Expanding the whole file, I see it at the top.

client-v2/package-lock.json
compulim: Run `npm install` and it will try to fix up stuff (e.g. add `dev: true` to dev packages). But I can't say if it works 100% of the time. I try:

* Not to delete `package-lock.json` if possible
* Only use `npm install --save` to change `package.json/d[evD]ependencies` section

But I agree merging `package-lock.json` is not an easy task. Yarn still have the very same story `yarn.lock`, but they are on YAML, slightly easier for eyes.

PR #393, opened 2018-01-04
Enable incremental Electron build

compulim:
# Why we want this
Every time we change anything under `/src/`, we need to run `gulp build-app` manually. And it is not incremental build so it is slow to develop the Electron "main" side of the app.

# What changed
* `gulp-tsc` does not support incremental build, moving away
* Moving `/src/` build into `npm run build:electron`, while keeping `gulp build-app` on-par

# How
It is same as what you did before: `npm run build` for clean build, `gulp build-app` for Electron "main" only. But now you can enable incremental build on "main" side.

To build stuff in lightning speed (~4 seconds), run `npm run build:electron -- --watch`. Then run the Electron app. Your "main" side of code will be watching and building incrementally.

> Tips: don't forget `build:electron` or `gulp build-app` will only build the Electron "main" side, you need to run `gulp build-react` for the "renderer" side, which output under `/app/client/` folder. You can use `ELECTRON_TARGET_URL` to retarget to http://localhost:3000/ for Webpack dev server and LiveReload.

> Tips: our original `gulp-tsc`/`build-app` task does not clean the target folder at `/app/`, so as the new one. So you may have stale code under `/app/` and it still "works on my machine".

# What else
Don't know why there is a type error in `ngrok.ts` that is not caught by `gulp-tsc` but `tsc`. Tried few things but still not able to fix it.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,12 +9,11 @@ gulp.task('clean', function () {
 });
 
 gulp.task('build-app', function () {
-    const tsc = require('gulp-tsc');
-    const tsconfig = require('./tsconfig.json');
-
-    return gulp.src(['src/**/*.ts', 'src/**/*.tsx'])
-        .pipe(tsc(tsconfig.compilerOptions))
-        .pipe(gulp.dest('app/'));
+    return gulp
+        .src('./package.json', { read: false })
+        .pipe(shell([
+            'npm run build:electron'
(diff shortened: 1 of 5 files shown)


eanders-MS:
Re: ngrok error. We have a work item to remove ngrok.ts and replace with the ngrok npm package.

PR #394, opened 2018-01-05
Shakeup the Redux store for MDI

compulim:
Shakeup on store to support multiple types of documents.

# Supported types

We only have 3 types of document today:

* Bot chat
* Card
* Conversation

# Using the MDI actions

To create and open a bot connection. You need to specifically ask the MDI to open the document for you.

```js
const connectAction = BotActions.connect(`http://localhost:4000/`);

this.props.dispatch(connectAction);
this.props.dispatch(EditorActions.open(constants.ContentType_BotChat, connectAction.payload.botId));
```

# Store dump

```js
{
  bot: {
    bots: {
      'bot:1': { url: '…' },
      'bot:2': { url: '…' }
    }
  },
  card: {
    cards: {
      'card:1': { … }
    }
  },
  conversation: {
    conversations: { … }
  },
  editor: {
    activeDocumentId: 'bot:1',
    documents: [{
      contentType: 'application/vnd.microsoft.botstudio.document.botchat',
      documentId: 'bot:1'
    }, {
      contentType: 'application/vnd.microsoft.botstudio.document.card',
      documentId: 'card:1'
    }]
  }
}
```

### DIFF
--- a/client-v2/src/ui/shell/explorer/botExplorer.js
+++ b/client-v2/src/ui/shell/explorer/botExplorer.js
@@ -31,11 +31,14 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
+import { connect } from 'react-redux';
 import { css } from 'glamor';
 import React from 'react';
-import { connect } from 'react-redux';
-import ExpandCollapse, { Controls as ExpandCollapseControls, Content as ExpandCollapseContent } from '../../layout/expandCollapse';
+
 import * as BotActions from '../../../data/action/botActions';
+import * as constants from '../../../constants';
+import * as EditorActions from '../../../data/action/editorActions';
+import ExpandCollapse, { Controls as ExpandCollapseControls, Content as ExpandCollapseContent } from '../../layout/expandCollapse';
 
 const CSS = css({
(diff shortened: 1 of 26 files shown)


client-v2/src/ui/shell/explorer/botExplorer.js
```diff
@@ -63,8 +66,10 @@ class BotExplorer extends React.Component {
     }
 
     handleAddClick(e) {
-        e.stopPropagation();
```
eanders-MS: Button click was propagating to the parent, causing the control to toggle expand/collapse. How is this inhibited now that the call to e.stopPropagation() is removed?

client-v2/src/utils.js
```diff
@@ -33,6 +33,5 @@
 
 
 export function uniqueId() {
-    Math.random().toString(36).substr(2);
+    return Math.random().toString(36).substr(2);
```
eanders-MS: Good catch!

client-v2/src/ui/shell/explorer/botExplorer.js
compulim: Because we separated the `onClick` for the expand/collapse title and its accessories, thus, we can get rid of the `stopPropagation()` call.

PR #396, opened 2018-01-08
Hooked up Card Editor with New Store Structure

Fignewton5:
## Major Changes 
- Card unique id is now created on the store-side instead of in the `CardExplorer`
- Opening a root assets folder through the `AssetExplorer` will now look for possible card files (any `.json` file in the directory `{rootFolder}/cards/*.json`) and populate the explorer with their filenames
- Can now open cards from the `CardExplorer` by clicking on the card name
- Can have multiple cards open simultaneously, switch between them, and save them to disk
- Output log is now specific to each card and can be cleared
- Can no longer open duplicates of the same document in the editor
- Fixed odd styling of card editor (splitter is still having some strange side effects)


client-v2/src/data/reducer/editor.js
```diff
@@ -74,3 +77,15 @@ export default function documents(state = DEFAULT_STATE, action) {
 
     return state;
 }
+
+function documentExists(id, documents) {
```
compulim: You can make it shorter like this:

```js
function documentExists(id, documents = []) {
  return documents.some(doc => doc.documentId === id);
}
```

client-v2/src/ui/editor/cardEditor/cardJsonEditor.js
```diff
@@ -134,13 +134,13 @@ class CardJsonEditor extends React.Component {
...
         const { loaderReady: prevLoaderReady } = prevState || {};
-        const { editorWidth: prevEditorWidth } = prevProps || null;
+        const { editorWidth: prevEditorWidth, cardId: prevCardId, cardJson: prevCardJson } = prevProps || null;
 
         if (prevEditorWidth && this._editor) {
             this._editor.layout();
         }
 
-        // if we have the monaco loader loaded, then let's create the editor
+        // if we switched to another asset tab and back, then we need to load the loader again
```
compulim: Woot! I thought we loaded the loader into `window.amdLoader` and it won't be unloaded if we remove the `<script>` tag.

client-v2/src/ui/editor/cardEditor/cardPreviewWindow.js
```diff
@@ -95,72 +99,79 @@ class CardPreview extends React.Component {
...
-        this.cardContainer = element;
+        this.cardContainer = ReactDOM.findDOMNode(element);
     }
 
     render() {
         return(
-            <div {...CSS} {...debug}>
+            <div className={ CSS } {...debug}>
                 <span className="preview-header">Preview</span>
                 <div className="preview-content" ref={ this.saveCardContainer }></div>
```
compulim: (Not in this PR).

You can self-close HTML tag in React and it will handle it correctly. Like `<div />` will be rendered as `<div></div>`. No need to recite on which can be self-closed or not.

client-v2/src/ui/editor/cardEditor/cardPreviewWindow.js
```diff
@@ -95,72 +99,79 @@ class CardPreview extends React.Component {
...
-// determines a card action type, formats the message, then saves it to the store
-function processAction(action, dispatch) {
-    let actionType;
-    if (action.url) {
-        actionType = "OpenUrl";
-    } else {
-        actionType = "Submit";
+    // determines a card action type, formats the message, then saves it to the store
+    processAction(action, dispatch) {
+        let actionType;
```
compulim: Cleaner and quicker to read.

```js
const actionType = action.url ? 'OpenUrl' : 'Submit';
```

(And single-quote for all JS strings)

client-v2/src/ui/editor/cardEditor/cardPreviewWindow.js
```diff
@@ -95,72 +99,79 @@ class CardPreview extends React.Component {
...
 
-    const actionMsg = formatActionMessage(action, actionType);
-    dispatch(addCardOutputMessage(actionMsg));
-}
+    // formats a card action output message based on the action type
+    formatActionMessage(action, actionType) {
+        const timeStamp = new Date().toTimeString().substring(0, 8);
+        const msgPrefix = `${timeStamp} [Card Action] - ${actionType}`;
+
+        let msgBody = "\n\tTitle: " + action.title;
```
compulim: Single quote and consider string templating like `\n\tTitle: ${ action.title }`.

client-v2/src/ui/editor/cardEditor/cardPreviewWindow.js
```diff
@@ -95,72 +99,79 @@ class CardPreview extends React.Component {
...
+            default:
+                // format hidden data as:
+                // Data: {
+                //     key1: val1,
+                //     key2: val2,
+                //     ...
+                // }
+                if (action.data && Object.keys(action.data).length) {
+                    msgBody += "\n\tData: {";
+                    let keys = Object.keys(action.data);
```
compulim: Will `JSON.stringify` help here?

client-v2/src/ui/editor/cardEditor/cardTemplatorWindow.js
```diff
@@ -71,16 +73,25 @@ class CardTemplator extends React.Component {
 
     render() {
         return (
-            <div {...CSS} {...debug}>
+            <div className={ CSS } {...debug}>
```
compulim: (Not in this PR)

Space inbetween contents of `{}`, like `{ ...debug }`, instead of `{...debug}`.

client-v2/src/ui/editor/cardEditor/cardPreviewWindow.js
compulim: Oh, if it's not for production but debugging, forget it.

client-v2/src/ui/editor/cardEditor/cardTemplatorWindow.js
```diff
@@ -71,16 +73,25 @@ class CardTemplator extends React.Component {
...
                 <span className="template-header">Template Editor</span>
                 <div className="template-content">
-                    { this.props.entities.map(ent => <CardTemplateRow key={ ent } entityName={ ent } />) }
+                    {
+                        this.props.length ?
+                        this.props.entities.map(ent => <CardTemplateRow key={ ent } entityName={ ent } />)
```
compulim: Indent like this, or if you could find a even-shorter way. The idea is to make both true/false case have same status (on same indent), while letting reader to quickly recognize this is ternary operator.

```js
<p>
  {
    predicate ?
      <div>True</div>
    :
      <div>False</div>
  }
</p>
```

client-v2/src/ui/shell/explorer/cardExplorer.js
```diff
@@ -56,6 +61,44 @@ const BOTS_CSS = css({
...
 export class CardExplorer extends React.Component {
+    constructor(props, context) {
+        super(props, context);
+
+        this.handleCardClick = this.handleCardClick.bind(this);
+    }
+
+    componentWillMount() {
+        // look at the current folder and check for cards under it
+        const cardsPath = this.props.folder + "/cards";
```
compulim: Single quote and consider string templating.

FYI, we will move native code out to main thread. I will read the content and push new changes to Redux from time to time.

client-v2/src/ui/shell/explorer/cardExplorer.js
```diff
@@ -56,6 +61,44 @@ const BOTS_CSS = css({
...
+                    };
+
+                    this.props.dispatch(CardActions.createCard(cardContent));
+                }
+            }
+        }
+    }
+
+    handleCardClick(e, id) {
+        e.stopPropagation();
```
compulim: We should not need `stopPropagation` on every `onClick` if we built the DOM correctly.

client-v2/src/ui/shell/explorer/cardExplorer.js
```diff
@@ -69,7 +112,7 @@ export class CardExplorer extends React.Component {
                                 {
                                     Object.keys(this.props.cards).length ?
                                         Object.keys(this.props.cards).map(id =>
-                                            <li key={ id }>{ this.props.cards[id].title }</li>
+                                            <li onClick={ (event) => this.handleCardClick(event, id) } key={ id }>{ this.props.cards[id].title }</li>
```
compulim: No need parenthesis if there is only one single argument in the lambda function.

Consider componentizing this `<li>` for perf reason (no need to build lambda function on `render()`).

client-v2/src/ui/shell/multiTabs/index.js
```diff
@@ -44,7 +44,9 @@ const CSS = css({
     backgroundColor: 'orange',
     display: 'flex',
     flex: 1,
-    flexDirection: 'column'
+    flexDirection: 'column',
+    height: "100%",
```
compulim: Single quote.

client-v2/src/ui/shell/multiTabs/index.js
```diff
@@ -70,7 +72,7 @@ export default class MultiTabs extends React.Component {
                         )
                     }
                 </TabBar>
-                { filterChildren(React.Children.toArray(this.props.children)[this.props.value].props.children, child => child.type === TabbedDocumentContent) }
+                { this.props.children.length ? filterChildren(React.Children.toArray(this.props.children)[this.props.value].props.children, child => child.type === TabbedDocumentContent) : false }
```
compulim: Consider this, reader doesn't need to find/read the negative case for the ternary operator.

```js
{ !!this.props.children.length && filterChildren(React.Children.toArray(this.props.children)[this.props.value].props.children, child => child.type === TabbedDocumentContent) }
```

client-v2/src/ui/utils.js
```diff
@@ -32,7 +32,43 @@
 //
 
 import React from 'react';
+if (typeof window !== 'undefined') { require = window['require']; }
```
compulim: FYI, we will move native calls out.

client-v2/src/data/reducer/editor.js
Fignewton5: Much nicer!

client-v2/src/ui/editor/cardEditor/cardPreviewWindow.js
Fignewton5: Can do. Also, what's the benefit of using single quotes? I used to use all single quotes, then switched back to double quotes just as a personal preference.

client-v2/src/ui/editor/cardEditor/cardPreviewWindow.js
Fignewton5: It's really placeholder functionality as of now, but I can try `JSON.stringify` :)

client-v2/src/ui/editor/cardEditor/cardPreviewWindow.js
compulim: The old story: free up double quotes so you can use it in HTML inside JavaScript code.

```js
document.body.innerHTML = '<div id="root"></div>';
```

client-v2/src/ui/editor/cardEditor/cardOutputWindow.js
```diff
@@ -83,16 +83,15 @@ class CardOutput extends React.Component {
...
         this.props.dispatch(CardActions.clearCardOutputWindow(this.props.cardId));
     }
 
     render() {
         return (
-            <div className={ CSS } {...debug}>
-                <span className="output-header">Output <span onClick={ this.clearOutput }>X</span></span>
-                <div className="output-content">
+            <div className={ CSS } { ...debug }>
+                <span className='output-header'>Output <span onClick={ this.clearOutput }>X</span></span>
```
compulim: You would leave this as double-quotes.

client-v2/src/ui/editor/cardEditor/cardPreviewWindow.js
eanders-MS: I like to use double quotes for strings the user will see. Single quotes for everything else, unless like above where it makes embedding a quoted string in another string. This helps with finding strings that need to be localized later.

client-v2/src/ui/utils.js
```diff
@@ -32,7 +32,43 @@
...
+const fs = require('fs');
```
eanders-MS: FYI, all 'fs' operations are moving to the Node side (echoing what William said above)

client-v2/src/ui/editor/cardEditor/cardPreviewWindow.js
Fignewton5: I see, thanks guys. 👍

PR #397, opened 2018-01-09
Tree view component

compulim:
It lives in test bed right now.

It is working for now but need further improvements:

* Separate UI testbeds into their own page for better testing
* Find a better DOM model for indentation
  * Currently, it indent `padding-left` of `<button>`
  * And it requires injecting `props.depth` into `<Branch>` and `<Leaf>`
* Add accessories button (on hover or static)
  * It should still keep whole row clickable
  * Some accessories only visible on hover/focus
* Accessibility (up/down/left/right arrow)

image


compulim:
Should PR against `studio`.

PR #398, opened 2018-01-09
Tree view component

compulim:
It lives in test bed right now.

It is working for now but need further improvements:

* Turns `treeViewFactory` into a `<FlatTreeView>` component
* Separate UI testbeds into their own page for better testing
* Find a better DOM model for indentation
  * Currently, it indent `padding-left` of `<button>`
  * And it requires injecting `props.depth` into `<Branch>` and `<Leaf>`
* Add accessories button (on hover or static)
  * It should still keep whole row clickable
  * Some accessories only visible on hover/focus
* Accessibility (up/down/left/right arrow)
* Better event handling at `<Leaf onClick>`, no `.bind`

image


compulim:
Let me work on some more stuff, will reopen soon.

client-v2/src/ui/widget/treeView.js
```diff
@@ -0,0 +1,218 @@
...
+// THE SOFTWARE IS PROVIDED ""AS IS"", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+import { css }    from 'glamor';
```
eanders-MS: I recommend against this kind of columnar formatting. It introduces a maintenance task for other devs that will invariably not happen sometimes, leading to degraded code consistency across the project.

client-v2/src/ui/utils/expandFlatTree.js
```diff
@@ -0,0 +1,52 @@
...
+export default function expandFlatTree(flattened, delimiter = '/') {
+    if (Array.isArray(flattened)) {
+        flattened = flattened.reduce((map, path) => {
+            map[path] = path;
+
+            return map;
+        }, {});
+    }
+
+    return Object.keys(flattened).reduce((expanded, path) => {
```
Fignewton5: This is awesome! 👍

client-v2/src/ui/widget/treeView.js
compulim: Yea, I was only using this columnar formatting in my toy projects. Probably accidentally hitting my own key combos to format this in columns. Will fix it.

client-v2/src/ui/utils/expandFlatTree.js
compulim: I learnt a new way to make `map`/`reduce` more readable by using spread/rest operator, at a cost of some performance:

```js
function arrayToMap(array) {
  return array.reduce((map, value) => {
    map[value.id] = value;

    return map;
  }, {});
}
```

Into

```js
function arrayToMap(array) {
  return array.reduce((map, value) => ({
    ...map,
    [value.id]: value
  }), {});
}
```

PR #411, opened 2018-02-14
storing status info as return from bot framework and giving it back t…

amitkumarshukla:
…o the bot framework with next request


eanders-MS:
I'm sorry we cannot accept your changes. `CurrentState` isn't a reserved entity type in the protocol, and therefore not something we'd make generally available to all users.

Can you describe what you're trying to accomplish? Maybe there's another way to accommodate it.

Issue #416, opened 2018-02-18
Support custom events to test custom Webchat clients.


Crazedpeanut:
Provide a mechanism to post custom events to a bot.

A form should be included to insert the following event details:

- Name; The name of the event to be handled by bot
- Value; The event data to be consumed by bot.

The event sent to the bot
<img width="550" alt="screen shot 2018-02-17 at 10 19 47 pm" src="https://user-images.githubusercontent.com/7700874/36349227-a82d0344-14d5-11e8-9ead-db0f71659c3e.png">

Custom event form to specify name and value
<img width="527" alt="screen shot 2018-02-18 at 6 00 20 pm" src="https://user-images.githubusercontent.com/7700874/36349228-a86373a2-14d5-11e8-8ee7-7eafd4cff7f5.png">

The customEvent button in the address bar menu
<img width="289" alt="screen shot 2018-02-18 at 6 00 29 pm" src="https://user-images.githubusercontent.com/7700874/36349229-a893df1a-14d5-11e8-8ae8-bf0bf983a718.png">

eanders-MS:
This is great. I think this is very reasonable request. Help me understand: Are you developing your own chat client that sends custom event activities?

negativeeddy:
I have a similar need. I use the webchat control, but when the control loads, the page it is on sends custom user information to the bot via the backchannel in order to customize the user's experience (or to use as tokens for backend APIs).

Crazedpeanut:
@eanders-MS exactly, I am developing a custom chat client. 

 @negativeeddy, I have basically the same use-case.

I created a PR, hopefully it gets merged in soon :)

petrformanek:
I also need such feature. I'm sending custom event with initialization data such as URL of external Web API's.
Thanks

Crazedpeanut:
Hi @eanders-MS, I am developing a custom chat client that sends custom events.

Any suggestions or feedback for my PR?

eanders-MS:
Hi @Crazedpeanut, we aren't taking PRs against this repo right now, unfortunately. We will resume PR acceptance around mid-May. It is possible to test using your custom client and the emulator. Here are the instructions, in case this is an option for you:
1. When the emulator starts up, you'll see something like `Emulator listening on http://[::]:2032` in the log. That's the emulator's Direct Line endpoint. Make a note of this url, but replace `[::]` with `localhost`. Also, your port number will probably be different.
[...]

PR #417, opened 2018-02-18
Support custom events issue #416

Crazedpeanut:
Add custom event form and support sending event activities to the bot framework.

See Github Issue #416 for more information

Keen to hear back with any comments or feedback!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,9648 @@
+{
+  "name": "botframework-emulator",
+  "version": "3.5.35",
+  "lockfileVersion": 1,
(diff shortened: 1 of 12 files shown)


vishwacsena:
@Crazedpeanut, thanks for putting this together. As Eric had noted in #416 , we are not accepting any PRs at the moment that targets the V3 Emulator. That said, it would be fantastic if you can do a PR for the [V4 emulator](https://github.com/Microsoft/BotFramework-Emulator/blob/v4/CONTRIBUTING.md#how-to-build-from-source).

PR #445, opened 2018-05-01
Jwilaby/fix/22690 unable to create bot

justinwilaby:


packages/app/client/src/ui/shell/explorer/luisExplorer/luisExplorer.tsx
```diff
@@ -1,3 +1,4 @@
+import { LuisService } from 'msbot/bin/models';
 import { IQnAService } from 'msbot/bin/schema';
```
eanders-MS: IQnAService import needed here?

justinwilaby: I'll fix this on the next PR. Looks like an artifact from another developer commit.

PR #458, opened 2018-05-02
Jwilaby/fix/22725 transcript file names

justinwilaby:
This PR provides the transcript file name in the transcript tabs for files and deeplinks

### DIFF
--- a/packages/app/client/src/data/reducer/editor.ts
+++ b/packages/app/client/src/data/reducer/editor.ts
@@ -31,9 +31,9 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-import * as EditorActions from '../action/editorActions';
-import * as Constants from '../../constants';
 import { deepCopySlow } from '@bfemulator/app-shared';
+import * as Constants from '../../constants';
+import * as EditorActions from '../action/editorActions';
 import { EditorAction } from '../action/editorActions';
 import { getOtherTabGroup, tabGroupHasDocuments } from '../editorHelpers';
 
@@ -60,6 +60,7 @@ export interface IDocument {
   contentType?: string;
   dirty?: boolean;
   documentId?: string;
+  fileName?: string;
   isGlobal?: boolean;
   meta?: any;
 }
@@ -199,7 +200,7 @@ export default function editor(state: IEditorState = DEFAULT_STATE, action: Edit
       // if the document is already in another tab group, focus that one
       if (tabGroupHasDocuments(state.editors[otherTabGroup]) && state.editors[otherTabGroup].documents[action.payload.documentId]) {
         const recentTabs = [...state.editors[otherTabGroup].recentTabs].filter(docId => docId !== action.payload.documentId);
-        recentTabs.unshift(action.payload.documentId)
+        recentTabs.unshift(action.payload.documentId);
         const tabGroupState: IEditor = {
           ...state.editors[otherTabGroup],
           activeDocumentId: action.payload.documentId,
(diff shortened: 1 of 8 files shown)


packages/app/client/src/data/reducer/editor.ts
```diff
@@ -244,6 +245,28 @@ export default function editor(state: IEditorState = DEFAULT_STATE, action: Edit
...
+      let i = editorKeys.length;
+      outer:while (i--) {
+        const documents = editors[editorKeys[i]].documents;
+        const documentKeys = Object.keys(documents);
+        let j = documentKeys.length;
+        while (j--) {
+          const document = documents[documentKeys[j]];
+          if (document.documentId === updatedDocument.documentId) {
+            documents[documentKeys[j]] = { ...document, ...updatedDocument };
+            break outer;
```
ToddN: Never seen this before. Cool.

packages/app/client/src/ui/editor/emulator/index.tsx
```diff
@@ -381,4 +367,15 @@ function mapStateToProps(state: IRootState, { documentId }: { documentId: string
   };
 }
 
-export default connect(mapStateToProps, null)(Emulator as any);
+function mapDispatchToProps(dispatch) {
```
ToddN: Personally, I wouldn't make a change like this after we're supposed to be in code freeze and when we don't have another bug bash coming up.  This is cleaner.  But it seems like it would be easy to accidentally break something in the transition. Why not make the change needed to fix the bug and do this after //Build?

justinwilaby: Unfortunately, this was required in order to get proper access to `state` and `dispatch`. I went though the calls after this change and tried to break it but wasn't able to. It should be a 1:1 in functionality. Without unit tests however, we'll never truly know ;)

packages/app/client/src/ui/editor/emulator/index.tsx
```diff
@@ -263,21 +253,22 @@ class Emulator extends React.Component<IEmulatorProps, {}> {
...
+            await CommandService.remoteCall('emulator:feed-transcript:deep-link', conversation.conversationId, props.document.activities);
           } catch (err) {
             throw new Error(`Error while feeding deep-linked transcript to conversation: ${err}`);
           }
         } else {
           try {
             // the transcript is on disk, so its activities need to be read on the main side and fed in
-            await CommandService.remoteCall('emulator:feed-transcript:disk', conversation.conversationId, props.document.documentId)
+            const fileInfo: { fileName: string, filePath: string } = await CommandService.remoteCall('emulator:feed-transcript:disk', conversation.conversationId, props.document.documentId);
+            this.props.updateDocument(this.props.documentId, {fileName: fileInfo.fileName});
```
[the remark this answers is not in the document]
justinwilaby: +1

packages/app/main/src/protocolHandler.ts
```diff
@@ -219,9 +220,10 @@ export const ProtocolHandler = new class ProtocolHandler implements IProtocolHan
               const conversationActivities = JSON.parse(transcriptString);
               if (!Array.isArray(conversationActivities))
                 throw new Error('Invalid transcript file contents; should be an array of conversation activities.');
-
+              const {name, ext} = Path.parse(url);
```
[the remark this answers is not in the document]
justinwilaby: +1

packages/app/client/src/data/reducer/editor.ts
justinwilaby: This is an alternative syntax to `Object.assign({}, document, updatedDocument)`

ToddN: I meant the break to a label.

PR #463, opened 2018-05-03
Changed bot creation dialog to use new flow.

tonyanziano:
- Removed bot path location field
- Changed title to 'New bot configuration'
- Changed endpoint field placeholder text
- Changed 'Save' button to 'Save and connect'
- After hitting 'Save and connect', the user is then prompted with the save .bot file dialog that is filled with a sanitized filename derived from the Bot Name field

### DIFF
--- a/packages/app/client/src/ui/dialogs/botCreationDialog.tsx
+++ b/packages/app/client/src/ui/dialogs/botCreationDialog.tsx
@@ -257,7 +257,7 @@ export default class BotCreationDialog extends React.Component<{}, BotCreationDi
       <div { ...CSS }>
         <Column>
           <MediumHeader className="bot-create-header">New bot configuration</MediumHeader>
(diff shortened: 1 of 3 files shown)


packages/app/client/src/ui/dialogs/botCreationDialog.tsx
```diff
@@ -236,27 +241,25 @@ export default class BotCreationDialog extends React.Component<{}, BotCreationDi
     this.setState({ secretConfirmation: e.target.value, secretsMatch: e.target.value === this.state.secret });
   };
 
+  private getSafeBotFileName = (name: string): Promise<string> => {
```
ToddN: Nit: I'd rename botFileName to botFilePath and getSafeBotFileName to getSafeBotFilePath

tonyanziano: I considered this, but it is actually just the name of the file and not the entire path. So it seems misleading to me to call it path.

PR #465, opened 2018-05-03
#22733 - removed fields designed for just a single bot

justinwilaby:


packages/app/client/src/ui/editor/botSettingsEditor/index.tsx
```diff
@@ -203,13 +203,6 @@ class BotSettingsEditor extends React.Component<BotSettingsEditorProps, BotSetti
     return (
       <GenericDocument style={ CSS }>
         <Column>
-          <MediumHeader className="bot-settings-header">Bot Settings</MediumHeader>
-          <TextInputField label="Bot name" value={ this.state.bot.name } required={ true } onChange={ this.onChangeName } />
```
eanders-MS: I think we want to keep bot name.

packages/app/client/src/ui/editor/botSettingsEditor/index.tsx
```diff
@@ -203,13 +203,6 @@ class BotSettingsEditor extends React.Component<BotSettingsEditorProps, BotSetti
...
-          <TextInputField label="Endpoint URL" value={ this.state.endpoint.endpoint } required={ true } onChange={ this.onChangeEndpoint } />
-          <Row className="multiple-input-row">
-            <TextInputField label="MSA App Id" value={ this.state.endpoint.appId } onChange={ this.onChangeAppId } />
-            <TextInputField label="MSA App Password" value={ this.state.endpoint.appPassword } onChange={ this.onChangeAppPw } type="password" />
-          </Row>
```
tonyanziano: Remove the `onChange` handlers for these removed fields?

justinwilaby: +1
I wasn't sure if we'd need these once we start to support multiple bots in the settings

packages/app/client/src/ui/editor/botSettingsEditor/index.tsx
justinwilaby: Added back the bot name and also installed validation for the required field.

packages/app/client/src/ui/editor/botSettingsEditor/index.tsx
tonyanziano: We can always dredge them out of the commit history. 😃

PR #472, opened 2018-05-04
Fixed protocol-initiated livechat sessions

tonyanziano:
- Livechat sessions initiated via protocol were broken by the changes around endpoint services and the move to the emulator/core package
- `ProtocolHandler` will now check to see if ngrok is running before falling back to waiting on ngrok's `connect` event
  - sometimes ngrok was already connected and the `connect` event would fire before the `ProtocolHandler` even started, which would cause it to wait forever and never start a livechat
- `ProtocolHandler` now calls the `bot:restart-endpoint-service` command like we do when we set an active bot, which will allow the emulator package to pick up the new endpoint
- Also added a filter on App Menu most-recent bots list to filter out bad items


packages/app/main/src/protocolHandler.ts
```diff
@@ -249,15 +260,23 @@ export const ProtocolHandler = new class ProtocolHandler implements IProtocolHan
 
     const appSettings: IFrameworkSettings = getSettings().framework;
     if (appSettings.ngrokPath) {
-      // if ngrok is configured, wait for it to connect and load the bot
-      ngrokEmitter.once('connect', (...args: any[]): void => {
+      if (running())
```
eanders-MS: Braces around these code blocks please.

tonyanziano: +1

PR #473, opened 2018-05-04
Update QnAMaker lib version num, fix for updating exisiting question

mgbennet:


packages/extensions/qnamaker/client/src/App.tsx
```diff
@@ -286,12 +287,15 @@ class App extends React.Component<any, AppState> {
     let success = false;
     try {
       if (this.state.qnaService !== null) {
-        const response = await this.client.publish(this.state.traceInfo.knowledgeBaseId);
-        success = response.status === 204;
+        $host.logger.log('Publishing... this could take a while.');
```
eanders-MS: Remove "this could take a while."

Issue #482, opened 2018-05-07
The emulator should be able to open files (bot/transcript etc) of different encodings


emadelwany:

tonyanziano:
Addressed in #483

tonyanziano:
Fix has been verified. Closing the issue.

Issue #165, opened 2017-04-17
Bot Emulator Connectivity issue


jacobchikku:
I am trying to connect just a Hello world bot application and its getting the error POST connect ECONNREFUSED 127.0.0.1:3978, on the emulator, i have tried with valid MSA id and secret and also blank ones. but still receiving the same.

jacobchikku:
Resolved this issue by downloading and unzipping and copying the Visual Studio Bot Application template from https://docs.botframework.com/en-us/downloads/

eanders-MS:
I'm glad you found a solution!

PR #483, opened 2018-05-08
Added support for different Node-supported file encodings.

tonyanziano:
- Added Support for different file encodings:
  - ascii
  - utf-8
  - utf-16le & ucs-2
  - latin1 & iso-8859-1
  - base64
  - binary
  - hex

*NOTE*: Have tested loading iso-8859-1 `.json` files and utf-16le `.transcript` files. I think it's possible to also read utf-16**be** encoded files as well using `Buffer.swap16()` but I haven't tried it yet.

### DIFF
--- a/packages/app/main/package-lock.json
+++ b/packages/app/main/package-lock.json
@@ -1509,6 +1509,11 @@
 				}
 			}
 		},
+		"chardet": {
+			"version": "0.5.0",
+			"resolved": "https://fuselabs.pkgs.visualstudio.com/_packaging/FuseNPM/npm/registry/chardet/-/chardet-0.5.0.tgz",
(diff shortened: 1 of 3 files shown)


eanders-MS:
Does this fix loading the transcripts saved by @szul?

tonyanziano:
Yeah, it wasn't very well documented either.

I don't think I can confirm the fix, because I was able to open his transcripts originally. We would have to figure out what encoding the transcripts he was generating were in, and then try to open one.

eanders-MS:
The transcript I downloaded from the share was UTF-16 LE, so seems this change is good on that front.

tonyanziano:
Oh his transcripts were utf-16le? Then I did actually test opening that encoding and was successful as well.

eanders-MS:
Looks good. Once you've merged it I'll kick off new builds. Thank you!

szul:
This update fixed my issue mentioned in #165 and seems to satisfy #482. The `.transcript` file is generating on my machine as UTF-16 LE, which previously was not being loaded by the Bot Emulator. Thanks, everyone!

tonyanziano:
Glad to hear it. I'm going to close #482 then.

PR #489, opened 2018-05-11
Switched form chardet package to read-text-file for different file encodings

tonyanziano:
- Swapped out `chardet` for `read-text-file` package which wraps `jschardet` and `iconv` to strip byte order marks and decode text files


packages/app/main/src/utils.ts
```diff
@@ -170,31 +170,9 @@ export const getFilesInDir = (path) => {
...
-      // encoding not supported by Node
-      throw new Error(`Error reading file ${path}: Encoding ${encoding} not supported by Node.`)
-    } else {
-      // bad encoding
-      throw new Error(`Error reading file ${path}: Couldn't detect encoding: ${encoding}`);
-    }
+    return readTextFile.readSync(path);
   } catch (e) {
+    console.error(`Error reading file ${path}: ${e}`);
     return '';
```
eanders-MS: This location is too low level to consume the error. I would let this propagate. Callers should catch and either handle or propagate further.

PR #492, opened 2018-05-14
Implemented a focus trap for all dialogs / modals

tonyanziano:
- Added focus trap to `host.tsx` component on the client side
  - Uses invisible, tab-able spans (I called them sentinels) on each side of the modal host's content div, and they catch focus events and pass them to the next available tab-able element within the modal host either at the beginning or end of the content

- Added optional prop of `tabIndex` to different input components in `ui-react/widgets` so that they can be tabbed to and work within the focus trap

- Added functionality to `host.tsx` that will also automatically focus the first tab-able element within the modal when it is shown

**Tested within bot-creation-style modal and services-style modal**


packages/app/client/src/ui/dialogs/host.tsx
```diff
@@ -85,17 +94,86 @@ class DialogHost extends React.Component<IDialogHostProps, IDialogHostState> {
...
+    if (this._hostRef) {
+      return this._hostRef.querySelectorAll('[tabIndex]:not([tabIndex="-1"])');
+    }
+    return new NodeList();
+  }
+
+  private initFocusTrap = () => {
+    const allFocusableElements = this.getFocusableElementsInModal();
+    if (allFocusableElements.length) {
+      // focus the first child after a small delay (issue with react where child hasn't yet rendered fully)
```
emadelwany: For my own edification (I'm new to react), shouldn't it be safe to do this in componentDidMount instead of the delay hack which seems a bit finicky.

tonyanziano: I actually just looked it up and `componentDidMount` does run after the component's children are mounted as well, which I was unaware of. I thought it fired down the DOM tree (parent -> child -> child's child) not upwards. That makes more sense. I'll make the change! 👍

tonyanziano:
Turns out that since the dialog host component is always mounted, `componentDidMount` is not a good indicator if when the child dialog component is finished rendering.

My solution was for the `DialogService` to use the `ReactDom.render()` callback to notify the dialog host element of when the child dialog has finished rendering, which the host can then react to by initializing the focus trap.

Another issue was that the dialog host was still `display: none` when the child dialog finished rendering, which then caused the focus trap to initialize and focus on the correct input field within the dialog. However, because its parent -- the dialog host -- was still `display: none`, the first input field, and the entire child dialog component was still hidden in the DOM, which ate the focus event but didn't display it. Then when the dialog host changed to become visible, the DOM refresh would lose the input's focus state.

Issue #570, opened 2018-05-16
Convert remaining JavaScript files to TypeScript


tonyanziano:

tonyanziano:
Addressed in #581. Closing

PR #581, opened 2018-05-17
Converted all JS files to TS/X and Refactored Code

tonyanziano:
# Converted all JS files to TS/X and Refactored Code

#570 

**Also:**
- Also cleaned up unused files
- Fixed explorer toggle


packages/app/client/src/ui/shell/mdi/tab.tsx
```diff
@@ -31,69 +31,72 @@
...
+    super(props);
 
     this.state = {
+      draggedOver: false,
       owningEditor: getTabGroupForDocument(props.documentId)
     };
   }
 
-  onDragStart(e) {
+  private onDragStart = (e) => {
```
emadelwany: Would be nice to strongly type events all over this PR, react has types for all sorts of events under React.*Event, or we can use any type definition library that defines browser events...

emadelwany:
This is a fantastic change, thanks for doing it!

packages/app/client/src/ui/shell/mdi/tab.tsx
justinwilaby: Since this is a bound function, it would be quite a lot more than just typing the events. e.g.
```
private onDragStart: DragEventHandler<HTMLElement> = (e: SyntheticEvent<HTMLElement> => {
  ....
}
```

packages/app/client/src/ui/editor/index.tsx
```diff
@@ -31,41 +31,46 @@
...
-        <Emulator mode="transcript" documentId={ document.documentId } dirty={ this.props.document.dirty } />
-        : contentType === Constants.ContentType_BotSettings ?
-        <BotSettingsEditor documentId={ document.documentId } dirty={ this.props.document.dirty } />
-        : contentType === Constants.ContentType_AppSettings ?
-        <AppSettingsEditor documentId={ document.documentId } dirty={ this.props.document.dirty } />
-        : contentType === Constants.ContentType_WelcomePage ?
-        <WelcomePage documentId={ document.documentId } />
-        : false
-    );
+    switch (contentType) {
```
justinwilaby: Thank you so much for this!

packages/app/client/src/ui/editor/index.tsx
```diff
@@ -31,41 +31,46 @@
...
+      case Constants.ContentType_LiveChat:
+        return (<Emulator mode="livechat" documentId={ document.documentId } dirty={ this.props.document.dirty } />);
```
justinwilaby: an optimization would be to reference `this.props.document.dirty` locally. The reason is minification will not reduce multiple calls to properties on `this` where it will reduce local declarations. e.g.
`const { dirty } = this.props.document;`

packages/app/client/src/ui/shell/mdi/tab.tsx
```diff
@@ -31,69 +31,72 @@
...
+}
 
-    this.onDragStart = this.onDragStart.bind(this);
-    this.onDragOver = this.onDragOver.bind(this);
-    this.onDragEnter = this.onDragEnter.bind(this);
-    this.onDragLeave = this.onDragLeave.bind(this);
-    this.onDrop = this.onDrop.bind(this);
-    this.onDragEnd = this.onDragEnd.bind(this);
+class Tab extends React.Component<TabProps, TabState> {
+  constructor(props: TabProps) {
```
justinwilaby: Nice! I love clean constructors!

packages/app/client/src/ui/shell/mdi/tabFactory.tsx
```diff
@@ -31,24 +31,38 @@
...
-  : props.document.contentType === Constants.ContentType_AppSettings ?
-    <GenericTab documentId={ props.document.documentId } title={ "Emulator Settings" } dirty={ props.document.dirty } />
-  : false
+import { IDocument } from '../../../data/reducer/editor';
+
+interface TabFactoryProps {
+  document?: IDocument;
+}
+
+export default (props: TabFactoryProps) => {
```
justinwilaby: So much more readable!

PR #604, opened 2018-05-23
Fix LUIS extension loading and refactor global CSS

emadelwany:


packages/sdk/ui-react/src/styles/globalCss.ts
```diff
@@ -0,0 +1,73 @@
...
+      cursor: 'default',
+      fontSize: '13px',
+      fontFamily: Fonts.FONT_FAMILY_DEFAULT,
+      height: '100%',
+      margin: 0,
+      minHeight: '100%',
+      overflow: 'hidden',
+      userSelect: 'none',
+    };
+    Object.assign(rootCss, rootCssAdditions);
```
tonyanziano: Just a suggestion, could use the spread operator here instead of `Object.assign()`:

    let rootCss = {
        backgroundColor: ...
        cursor: ...
    
        ...rootCssAdditions
    };

emadelwany: Good suggestion! Will move to that model in the next change (and replace all other assignments)

PR #623, opened 2018-05-31
Merge production Web Chat

compulim:
Now, we can use production Web Chat without modifications in Emulator:

* No custom build of Web Chat and Direct Line JS, we are using production version from NPM directly
* Chat history preservation/restoration is done thru Emulator Core, not hacking into Web Chat redux store
* Chat tabs behind the current tab will no longer unmount, but hidden and kept in DOM until close

Caveats:

After merging this PR, we will show send box and interactivity even we load transcript, this is because the PR for hiding send box is pending approval on Web Chat. Once it's in Web Chat, we can bump version and get that fixed without code change. Will track in another bug.

Some design considerations:

* Bot ID and user ID are now fixed to "bot" and "default-user" respectively
   * Otherwise, we will need to load physical transcript files before instantiating the document editor (need context switching: client -> main -> client)
   * Web Chat is not reactive to bot ID and user ID change
   * No one complain about random bot ID and user ID for Chatdown-generated transcript, seems no one really care about bot ID and user ID yet
* Tabs that hide behind active tab are now hidden, instead of unmounted
   * Otherwise, we will need to preserve Web Chat redux store, which means
      1. Preserving history in Web Chat is controversy and we probably need to keep custom Web Chat, instead of production
      2. Emulator redux store will need to keep a copy of another redux store (as of today), seems weird, and Web Chat redux store is not plain JSON too


packages/app/client/src/data/action/chatActions.ts
```diff
@@ -148,7 +147,7 @@ export function newDocument(documentId: string, mode: ChatMode, additionalData?:
       mode,
       documentId,
       conversationId: null,
-      webChatStore: createWebChatStore(),
+      // webChatStore: createWebChatStore(),
```
emadelwany: Is this no longer required? If so can we just delete the line?

emadelwany:
@compulim  so good to see so much custom code get deleted! Awesome PR 👍

PR #627, opened 2018-06-01
Bump Web Chat version to fix build break

compulim:


compulim:
Version bump, pass unit test, fix build break.

I am self-approving this.

Issue #597, opened 2018-05-22
Ability to open a .chat file and convert to transcript file within emulator


vishwacsena:

emadelwany:
We also want to be able to detect changes to the imported chat file, so that a conversation design can update the chat file, and see the changes reflected directly in the emulator.

emadelwany:
@tomlm - this is the feature you requested.
@vishwacsena  - FYI I pulled this to 6B as we want to run some usability studies using it.

vishwacsena:
File explorer shows .chat and .transcript files
Automatically convert .chat to .transcript (in memory) and render

Design: How do we notify on errors with conversion/ rendering?

tonyanziano:
Addressed in #660

Issue #542, opened 2018-05-16
Add refresh button to the emulator for transcript files or file watcher 


emadelwany:
Feature ask - Add refresh button to the emulator for transcript files or file watcher 

You have to close down transcript view in the emulator to "refresh" the view of the transcript.

emadelwany:
Related to #557

tonyanziano:
Addressed in #660

PR #660, opened 2018-06-27
Support for .chat files and transcript reloading

tonyanziano:
Fix for #597 & #542 & #668 

---

- `.chat` files are now shown in the explorer pane
  - can click on `.chat` files and they will be converted to a `.transcript` in-memory using the `chatdown` package
- any open `.transcript` and `.chat` files that have been changed on disk will ask the user if they want to reload the transcript when focusing on that tab in the Emulator


justinwilaby:
:shipit:

tonyanziano:
I believe #668 was being caused by the `takeLatest` saga effect which was cancelling previous sagas that already were executing code that was prompting users with a native dialog box.

Issue #648, opened 2018-06-20
Restarting the main side of the application wipes the bots list


tonyanziano:
When building / developing locally, restarting the main process will wipe the most recently used bots list.

tonyanziano:
Addressed in #661

PR #661, opened 2018-06-28
Fixed vanishing bots list, and separated utils functions.

tonyanziano:
Fix for #648 

---

- Fixed issue where restarting the app was wiping the `bots.json` file of recently used bots
  - Basically, every time the main side was started up, there was a state change that invoked the `store.subscribe()` listener before the app had a chance to load the bots list from disk, so the listener would get the empty array from the state and use it to overwrite `bots.json`
  - My solution is to get a reference of the bot list in state as soon as the app initializes, and when the store listener is invoked, compare the current bot list against that ref and only write to disk if they are different

- Also moved all the functions in `/main/utils.ts` to their own separate files to make it easier to test `/main/utils/{ utilFunctionName }.ts`
  - was trying to write a test for a utility function I wrote and it was trying to initialize the main window and electron because of all the dependencies inside of `utils.ts`


justinwilaby:
:shipit: 💩

PR #667, opened 2018-07-02
Add wide card layout

### LINKED ISSUE #543: Receipt Cards facts.value properties are cut off
Regardless of having the Files explorer sidebar on the left side open, the values of a receipt card's facts are cut off. See attached screenshots.

image

compulim:
Fix #543, add `wc-wide` to the Web Chat so it will make Adaptive Cards looks wider and not cutting off some digits.

### DIFF
--- a/packages/app/client/src/ui/editor/emulator/parts/chat.tsx
+++ b/packages/app/client/src/ui/editor/emulator/parts/chat.tsx
@@ -241,7 +241,7 @@ class Chat extends React.Component<Props> {
       );
 
       return (
-        <div id="webchat-container" className="wc-app" { ...CSS }>
+        <div id="webchat-container" className="wc-app wc-wide" { ...CSS }>
           <WebChat.Chat
             key={ document.directLine.token }
             { ...webChatProps }


compulim:
### Before
image

### After
image

mecampbellsoup:
@tonyanziano 	@compulim hey guys, I'm struggling with a similar issue - in some channels (e.g. Skype, Telegram, but not web chat) our hero cards are rendered with some of their contents cut off/truncated, and I cannot find any information about changing how these cards are rendered (so that they include all the details we specify when assembling the card). Here is an example of what I mean:

image

That image is a Skype Mac OS client. As you can see, the `Value:` key is cut off. 

Here is how the same card is rendered in Skype iOS (more is cut off!):

image

We're really hoping you guys can point us in the direction of some style/CSS configuration options. For example, if we use Adaptive Cards, are there some styling options in the [host config](https://github.com/Microsoft/AdaptiveCards/blob/master/samples/HostConfig/sample.json) that I can use to control the size such that all a card's contents are rendered all the time?

PR #679, opened 2018-07-09
Jwilaby/theming

justinwilaby:
This PR provides Themes for light, dark and high contrast.


packages/app/client/src/ui/shell/multiTabs/multiTabs.scss
```diff
@@ -0,0 +1,21 @@
+.multiTabs {
+  display: flex;
+  flex: 1;
+  flex-direction: column;
+  height: 100%;
+  box-sizing: border-box;
+  .tab {
+    height: 100%;
```
tonyanziano: Just a nitpick, but I've noticed that some scss files leave a space between each selector and some don't. We might want to consider deciding on a format going forward, but it's not something urgent we need to address. Personally, I prefer a space between each selector, but I'm indifferent as long as it's consistent.

packages/app/main/src/protocolHandler.ts
```diff
@@ -244,8 +244,10 @@ export const ProtocolHandler = new class ProtocolHandlerImpl implements Protocol
               const fileName = `${name}${ext}`;
               // open a transcript on the client side and pass in some
               // extra info to differentiate it from a transcript on disk
-              mainWindow.commandService.remoteCall('transcript:open', 'deepLinkedTranscript',
-                { activities: conversationActivities, inMemory: true, fileName });
+              mainWindow.commandService.remoteCall(
+                SharedConstants.Commands.Emulator.OpenTranscript, 'deepLinkedTranscript',
+                { activities: conversationActivities, deepLink: true, fileName }
```
tonyanziano: `deepLink` property should be `inMemory`

tonyanziano: `deepLinkedTranscript` remains the same

packages/extensions/luis/client/src/Controls/Header/Header.scss
```diff
@@ -0,0 +1,38 @@
...
+
+  #appId {
+    padding-left: 16px;
+    color: var(--neutral-3);
+  }
+
+  #left {
+    display: block;
+    float: left;
+
```
tonyanziano: whitespace

packages/extensions/qnamaker/client/src/media/ic_add_thin.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="14px" height="20px" viewBox="0 0 14 20" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 49.3 (51167) - http://www.bohemiancoding.com/sketch -->
+    <title>dark/qna/plus button/rest</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="dark/qna/plus-button/rest" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <polygon id="Add-icon" fill="#C8C8C8" points="13.25 9.09375 13.25 9.90625 7.15625 9.90625 7.15625 16 6.34375 16 6.34375 9.90625 0.25 9.90625 0.25 9.09375 6.34375 9.09375 6.34375 3 7.15625 3 7.15625 9.09375"></polygon>
+    </g>
+</svg>
```
tonyanziano: newline

packages/app/client/src/ui/dialogs/botCreationDialog/botCreationDialog.scss
```diff
@@ -0,0 +1,29 @@
...
+      display: none;
+    }
+  }
+
+  .botCreateForm .botCreateForm > * {
+    margin-bottom: 16px;
+  }
+}
+
+
```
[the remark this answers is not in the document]
justinwilaby: +1

packages/app/client/src/ui/editor/emulator/emulator.tsx
```diff
@@ -230,12 +149,12 @@ class EmulatorComponent extends React.Component<EmulatorProps, {}> {
       try {
         const conversation = await CommandServiceImpl.remoteCall('transcript:new', conversationId);
 
-        if (props.document && props.document.inMemory && props.document.activities) {
```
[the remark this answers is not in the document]
justinwilaby: +1

packages/app/main/src/protocolHandler.ts
justinwilaby: Looks like this matches up  with the `v4` branch. Are you sure this property needs to change?

packages/extensions/luis/client/src/Controls/Header/Header.scss
justinwilaby: +1

packages/extensions/qnamaker/client/src/media/ic_add_thin.svg
justinwilaby: +1

packages/app/client/src/ui/shell/multiTabs/multiTabs.scss
justinwilaby: I do as well. This was an artifact from the wholesale conversion from `glamor` to `scss`.  it would be nice to implement a formatting util as a pre-commit hook and fail the commit if formatting isn't to spec.

packages/app/main/src/protocolHandler.ts
tonyanziano: Ah, yeah. I forgot to change it in this file. Looks like I forgot to change it in `clientInitCommands` as well to `inMemory`

Don't worry about it, I'll put out a small PR to fix it

packages/app/client/src/ui/shell/multiTabs/multiTabs.scss
tonyanziano: Sounds good to me 👍

packages/app/main/src/settingsData/reducers/usersReducer.ts
```diff
@@ -31,51 +31,36 @@
...
-        newUsersById[user.id] = user;
+        const user1 = action.state.users[key];
+        newUsersById[user1.id] = user1;
       }
       return Object.assign({}, state, { usersById: newUsersById });
     }
-    case 'Users_RemoveUsers': {
+
+    case REMOVE_USERS: {
       // Object.assign({}, state, { }
```
tonyanziano: remove?

justinwilaby: I think the entire user reducer can be removed. I'm not seeing it used anywhere.

packages/app/main/src/utils/readFileSync.ts
```diff
@@ -31,11 +31,11 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-import * as readTextFile from 'read-text-file';
+import * as fse from 'fs-extra';
 
 export const readFileSync = (path4: string): string => {
   try {
-    return readTextFile.readSync(path4);
```
tonyanziano: This might cause an issue. We were using `read-text-file` because it strips encoding byte order marks from files. A user was running into an issue where using chatdown through the VSCode integrated terminal was encoding files in something other than `utf-8` and we were silently erroring when trying to read the file

tonyanziano: We were previously using `fs.readSync`

tonyanziano: https://github.com/Microsoft/BotFramework-Emulator/pull/483 for reference

justinwilaby: `BOM` is just `\uFEFF' so why use an library to strip it when you can just do this?
```js
const text = fs.readFileSync(path, 'utf-8');
return text.replace('\uFEFF', '');
```

tonyanziano: Can't it change depending on the encoding? I thought the representation depends on utf-8 / 16 / 32 etc.

tonyanziano: I think the issue was that it wasn't being saved as a utf-8, but something else like utf-16.

packages/app/main/src/settingsData/reducers/usersReducer.ts
tonyanziano: 👍

packages/app/main/src/utils/readFileSync.ts
tonyanziano: So I looked at the issue and the user's transcripts were being saved with utf-18le encoding, that's why we were using the library, because it takes care of multiple encodings.

packages/app/main/src/settingsData/reducers/usersReducer.ts
justinwilaby: I'll speak with @compulim before I remove this. It looks like he has commented code in [createConversation.ts](https://github.com/Microsoft/BotFramework-Emulator/blob/v4/packages/emulator/core/src/conversations/middleware/createConversation.ts#L80) that dispatches. I'd like to know if this was meant to be iterated on later and should stay in the codebase.

PR #686, opened 2018-07-14
Fix speech token exchange

### LINKED ISSUE #646: Get speech recognition and TTS support for V4 Emulator
At the moment, the new emulator does not support speech in/ out. V3 does. we need to add support for this.

compulim:
Fix #646

### DIFF
--- a/packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
+++ b/packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
@@ -128,16 +128,16 @@ class Chat extends Component<Props> {
       );
 
       return (
-        <div id="webchat-container" className={`${styles.chat} wc-app wc-wide`}>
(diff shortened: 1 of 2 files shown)


packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
```diff
@@ -147,16 +128,16 @@ class Chat extends Component<Props> {
       );
 
       return (
-        <div id="webchat-container" className={ `${styles.chat} wc-app wc-wide` }>
+        <div id="webchat-container" className={`${styles.chat} wc-app wc-wide`}>
           <WebChat
-            key={ document.directLine.token }
-            { ...webChatProps }
+            key={document.directLine.token}
+            {...webChatProps}
```
tonyanziano: We have been using whitespace padding between brackets on JSX properties:

**Ex.**

`key={ document.directLine.token }`

Issue #528, opened 2018-05-16
Implement notifications


emadelwany:

tonyanziano:
Addressed in #701

PR #701, opened 2018-07-25
Implemented Notifications

tonyanziano:
Feature for #528 

---

## Added notifications system

- Notifications are now displayed in the **Notifications Explorer** accessible via the "Bell" icon in the nav bar
- Notifications consist of:
  - `type`: (Info / Warn / Error) Doesn't do anything as of now
  - `message`: Text do display
  - `timestamp`: automatically generated timestamp at the time of creation
  - `buttons`: optional buttons that can fire handlers when clicked
- Since notifications will inevitably be generated on the `main` side and need to be sent to the `client` side, and functions can't be sent over IPC, I created some utility functions to help facilitate passing notifications from `main` to `client` using the Electron `remote` object
  - `sendNotificationToClient()`: attaches a notification to the `global` object on the `main` side, calls a command on the `client` side that will use the `remote` object to retrieve that notification, and then deletes the notification from the `global` object

- Also due to the fact that we aren't storing functions in the redux store on the `client` side, notifications will be managed by a singleton class on the `client` called `NotificationManager` 

**OTHER:**

- Brought back a trimmed-down version of the `<InsetShadow />` component because notifications in the explorer were clipping into the box-shadow and making it look terrible


packages/app/client/src/data/sagas/notificationSagas.ts
```diff
@@ -0,0 +1,78 @@
...
+  finishRemove
+} from '../action/notificationActions';
+import { ForkEffect, takeEvery, call, put } from 'redux-saga/effects';
+
+/** Adds a notification to the notification manager's store then
+ *  adds it to the state store
+ */
+export function* addNotification(action: BeginAddNotificationAction): IterableIterator<any> {
+  const { notification } = action.payload;
+  const instance = yield call([NotificationManager, NotificationManager.getInstance]);
```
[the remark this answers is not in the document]
tonyanziano: I assume you are referring to the `yield` calls that call `NotificationManager.getInstance()` ? I was trying to make the sagas testable, but now I realize that I shouldn't have to test that `.getInstance()` was called since I can just check if `.addNotification` was called and the `put` was carried out.

packages/app/client/src/notificationManager.ts
```diff
@@ -0,0 +1,70 @@
...
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+import { Notification } from '@bfemulator/app-shared';
+
+/** Singleton that will keep track of all active notifications that we can't put in the store */
+export class NotificationManager {
```
justinwilaby: It seems like this entire class can be replaced by: 
```js
export const NotificationManager = new Map<Notification>();
```

packages/app/client/src/data/sagas/notificationSagas.ts
```diff
@@ -0,0 +1,81 @@
...
+  yield put(finishClear());
+}
+
+/** Removes a single notification from the notification manager's store then
+ *  removes it from the state store
+ */
+export function* removeNotification(action: BeginRemoveNotificationAction): IterableIterator<any> {
+  const { id: notificationId } = action.payload;
+  const instance = NotificationManager.getInstance();
+  // yield call([NotificationManager, NotificationManager.getInstance]);
```
justinwilaby: Remove?

packages/app/client/src/data/sagas/notificationSagas.ts
```diff
@@ -0,0 +1,81 @@
...
+  yield call([instance, instance.removeNotification], notificationId);
```
justinwilaby: Since `instance.removeNotification()` returns `void`, using `yield call()` is superflous

packages/app/client/src/data/reducer/notification.ts
```diff
@@ -0,0 +1,112 @@
...
+        },
+        allIds
+      };
+      break;
+    }
+
+    case NotificationActions.finishRemove: {
+      const { id: idToRemove } = action.payload;
+      const byId = {};
+      Object.keys(state.byId).forEach(notifId => {
```
justinwilaby: This can be shortcut by the following:
```js
const { [idToRemove]: toRemove, ...rest } = state.byId;
state = {
  ...rest
  // others...
}
```
```

File: packages/app/client/src/data/reducer/notification.ts [hunk 6]
```diff
@@ -0,0 +1,112 @@
...
+  };
+  allIds: string[];
+}
+
+const DEFAULT_STATE: NotificationState = {
+  byId: {},
+  allIds: []
+};
+
+export function notification(state: NotificationState = DEFAULT_STATE, action: NotificationAction): NotificationState {
```
justinwilaby: Would it make sense that the `Notification` Object Model contains a the field: `public read: boolean;` instead of maintaining a store of id's? Is there a reason for keeping this separate?

packages/app/client/src/data/reducer/notification.ts
tonyanziano: woah, awesome!

packages/app/client/src/data/sagas/notificationSagas.ts
tonyanziano: +1

packages/app/client/src/notificationManager.ts
tonyanziano: +1

packages/app/client/src/data/sagas/notificationSagas.ts
tonyanziano: I was mainly using `call` so that I could unit test the saga. Is it not worth it to test that `removeNotification` was called with the proper id? Or should I only be testing the `put` at the end?

packages/app/client/src/data/reducer/notification.ts
tonyanziano: Yes, that does make more sense. I was trying to follow the Redux docs for normalizing state and must have tunnel visioned their approach 👍 .

tonyanziano: I believe my thinking was that if I had the `read` status searchable by an associative array then I would only have to perform a re-render on the UI associated with the affected notification when it was updated, but I now see that `NotificationManager` provides that same associative array.

PR #709, opened 2018-07-30
Jwilaby/aad auth for abs

justinwilaby:
This PR provides AAD auth to retrieve the user's Arm Token to query LUIS and QnAMaker for models and for the Authoring key.


packages/app/main/src/services/azureAuthWorkflowService.ts
```diff
@@ -0,0 +1,164 @@
...
+    return new Promise(resolve => {
+      browserWindow.addListener('close', () => resolve(false));
+      browserWindow.addListener('page-title-updated', event => {
+        const uri: string = (event.sender as any).history.pop() || '';
+        if (!uri.toLowerCase().includes('localhost')) {
+          return;
+        }
+        const idx = uri.indexOf('#');
+        const values = uri.substring(idx).split('&');
+        let len = values.length;
```
tonyanziano: `const len`

Issue #705, opened 2018-07-26
Add Hotkey for "Start Over" button during livechat session


tonyanziano:

tonyanziano:
Addressed in #754

PR #754, opened 2018-08-08
Added hotkey (Ctrl/Cmd + Shift + R) for starting a conversation over.

tonyanziano:
Fix for #705 

===========

- Pressing `Ctrl` + `Shift+ `R` or `Cmd` + `Shift+ `R` with a chat session open will now restart that conversation
- Also fixed the chat panel header. We were trying to read the endpoint from an outdated property, and the endpoint should now once again be visible in the header


tonyanziano:
**Note:** This doesn't work with multiple chat sessions open, so I'm working on a fix.

tonyanziano:
Fixed the multiple chat sessions issue. Also got rid of `chat.pingId` which was being incorrectly used as a React `key` everywhere where there weren't any elements in an array.

PR #755, opened 2018-08-08
moved back to TS 2.7.2 and optimized build

justinwilaby:
reverts back to TS 2.7.2, removes prop-types from @types directory with each build and other optimizations.


packages/app/client/package.json
```diff
@@ -12,7 +12,7 @@
     "build:shared:dev": "webpack --mode development --progress --colors",
     "build:app:dev": "webpack --mode development --progress --colors",
     "build": "run-s lint build:vendors build:shared build:app",
-    "clean": "rimraf build",
+    "clean": "rimraf ../../../node_modules/@types/prop-types",
```
tonyanziano: should this still clean build?

justinwilaby: Client builds to `/public` which is managed by webpack

tonyanziano: 👍

Issue #536, opened 2018-05-16
In-app Help System (callouts, etc)


emadelwany:

tonyanziano:
Needs design @mewa1024

vishwacsena:
Addressed in current v4 release.

PR #764, opened 2018-08-14
Change label for LUIS version to avoid ambiguity

n01d:
With the label "Version" developers might expect this to be the LUIS api version, as that information is required when doing API calls.


justinwilaby:
Thank you for the contribution!

I have discussed this change with our designer who would like to maintain the labels as-is at this time so we best align with the luis.ai portal's naming. This is a great suggestion for future work should we revisit the input labeling in this form.

tonyanziano:
@n01d we have plans to include in-app callouts and help tips (#536) which I think this issue would be a perfect use case for. This would allow us to place a small question mark icon next to the LUIS version label that when hovered over / clicked could display more detailed information about what version we are asking for.

Issue #619, opened 2018-05-30
UI needed to edit bot list (bots.json)


mewa1024:

vishwacsena:
Let's do the clear part for my bots MRU

tonyanziano:
Confirmation prompt has been added. Just need final UI design.

mewa1024:
@tonyanziano Let's follow the VS Code pattern: show remove button on hover, use the X (same as the file tab X) as the icon:

<img width="608" alt="image" src="https://user-images.githubusercontent.com/39317542/44073746-4a9a353a-9f4a-11e8-90b8-a67ff74ca960.png">

tonyanziano:
Sounds good, will do. 👍

tonyanziano:
Addressed in #767

PR #767, opened 2018-08-14
Final design for bot list delete ui.

tonyanziano:
Fix for #619 

==========

- Styled according to final design spec
- Added theming

### DIFF
--- a/packages/app/client/src/ui/editor/welcomePage/welcomePage.scss
+++ b/packages/app/client/src/ui/editor/welcomePage/welcomePage.scss
@@ -14,32 +14,30 @@
     overflow: hidden;
     text-decoration: none;
     color: var(--focused-selected-list-item-bg);
-
-    :hover {
-      color: var(--focused-selected-list-item-bg);
-    }
   }
 
   ul {
     margin: 0;
     list-style: none;
     padding: 0;
+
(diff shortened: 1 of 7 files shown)


packages/app/client/src/ui/editor/welcomePage/welcomePage.scss
```diff
@@ -79,24 +86,32 @@
...
-    text-align: center;
+    height: 0;
+    width: 0;
     cursor: pointer;
+
+    &:after {
+      height: 16px;
+      width: 8px;
+      margin-right: 8px;
+      content: "\2716";
```
justinwilaby: Looks like this is missing the "u".  Should be `\u2716`

tonyanziano: I tried `\u2716` and it just printed `u2716` instead of the icon.

I also looked around online and it seems that you don't need to put the '\u" before the character codes to get them to appear.

[Source 1](https://brajeshwar.github.io/entities/)
[Source 2](https://stackoverflow.com/questions/10393462/placing-unicode-character-in-css-content-value)
[Source 3](https://css-tricks.com/snippets/html/glyphs/)

Issue #505, opened 2018-05-15
Migrate v3 endpoints to v4 bots


emadelwany:

emadelwany:
Needs a spec

vishwacsena:
[Spec](https://raw.githubusercontent.com/wiki/Microsoft/BotFramework-Emulator/content/Specs/v3-to-v4-migration.md)

mewa1024:
@vishwacsena For this part of the spec in bold:
On V4 first launch, present Emulator V4 overview doc page that includes
        - Walkthrough of new features in the V4 Emulator
        - Provide an overview of the .bot files that were automatically created and ### **how developers can consolidate the endpoints into meaningful bot file structure**

Users need to be able to move their .bot files and not break the MRU. Can we move files through the UI?

sgellock:
we need V3 emulator packages built so we can set SxS

mewa1024:
@tonyanziano Comps of the entire flow are on Invision here: [https://microsoft-fuse-design.invisionapp.com/d/main#/projects/prototypes/15122715](https://microsoft-fuse-design.invisionapp.com/d/main#/projects/prototypes/15122715)

Sorry--I thought I'd linked to them earlier. I opened a new ticket to add the context menus to the Welcome page: #986

tonyanziano:
Improved by #1020 #1021

PR #780, opened 2018-08-17
V3 - > V4 Migration [V4 SIDE] 

tonyanziano:
Fix for #505 

===========

- Migration on the V4 side will check if the bots list is empty, and then if that is true, it will check the app folder for a file called `migration_marker.txt` to see if the V4 side of migration has been performed
- if the marker is not there, all the bots in `/migration/` will be added to the MRU and the post-migration dialog will be displayed


tonyanziano:
Sorry, didn't have time to write tests. I can do that when I get back. 👍

PR #788, opened 2018-08-23
Enable Travis deployment releases

JuanAr:
- Enabled Travis Deployments using GitHub Releases
- after_success: check if it is a Cron Job and generates a new tag name.
- before_deploy: run script to generate binaries (this is work in progress) and set the TRAVIS_TAG env var.
- deploy: creates a new Release including binaries files, name and description. It marks it as Pre-Release.
- condition: Release will be created for branch `v4` only and for Cron Jobs only.
- additionally, devDependencies wehre updated with those installed globally in before_install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,27 @@ install:
 script:
 - npm run build && npm run test:coveralls
 
+after_success:
+- echo "Cheking TRAVIS_EVENT_TYPE is cron -> TRAVIS_EVENT_TYPE = $TRAVIS_EVENT_TYPE"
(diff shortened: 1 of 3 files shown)


JuanAr:
The `GITHUB_OAUTH_TOKEN` environment variable is required in Travis for this to work. Also, a new Cron Job must be created since it is configure to create releases only for Cron Jobs.
The `GITHUB_OAUTH_TOKEN`  can be generated using the Travis commandline tool.

cleemullins:
I've updated the Travis build with my travis-ci.org API / OAuth key, added the GitHub_OAuth_Token enviornment variable, and setup the cron job. 

oauth

PR #795, opened 2018-08-28
Ensure that timestamp and recipient are not altered for historic activities.

ckkashyap:
Ensure that timestamp and recipient are not altered for historic activities.


packages/emulator/core/src/facility/conversation.ts
```diff
@@ -196,9 +196,9 @@ export default class Conversation extends EventEmitter {
   /**
    * Queues activity for delivery to user.
    */
-  public postActivityToUser(activity: Activity): ResourceResponse {
+  public postActivityToUser(activity: Activity, isHistoric: boolean = false): ResourceResponse {
```
cleemullins: This is new code, but there are no unit tests. 

Please add a unit test (or 5!) for this new behavior.

packages/emulator/core/src/facility/conversation.ts
```diff
@@ -654,17 +654,20 @@ export default class Conversation extends EventEmitter {
     return response;
   }
 
-  private postage(recipientId: string, activity: Activity): Activity {
+  private postage(recipientId: string, activity: Activity, isHistoric: boolean = false): Activity {
```
cleemullins: Please add unit tests for this new behavior.

justinwilaby:
The infrastructure for unit tests have not yet been set up for this package. I will need to set this up but do not have cycles until later in the week. Although I do not generally like to create more technical debt, I think  we should let this one go in and circle back once a unit test runner can be installed

packages/emulator/core/src/facility/conversation.ts
justinwilaby: Circling back on this once unit test infrastructure is setup for the `app\main` package

packages/emulator/core/src/facility/conversation.ts
justinwilaby: Circling back on this once unit test infrastructure is setup for the `app\main` package

Issue #772, opened 2018-08-16
Update welcome page to match design (align with portal, F5)


vishwacsena:

mewa1024:
Design is here:
https://microsoft-fuse-design.invisionapp.com/d/main#/console/15242220/316902430/preview

Issue #737, opened 2018-08-07
[MAS1.4.1] Links present in 'Welcome' pane are identifiable as link by their color only.


priysn:
**Actual Result:**
The only way to know the difference between links and normal text present on 'Welcome' screen is Blue color of those links.

**Expected Result:**
There should be some other way as well to know the difference between links and normal text present on 'Welcome' screen beside Blue color of those links. All links must be underlined.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Welcome screen will open up.
3. Notice links present on this page (In Blue).

**User Impact:**
Users with low vision or color-blindness will find it difficult to identify these links.

**Test Environment:**
OS: Windows 10
OS Build: 17134.167
OS Version: 1803
Browser: Edge

**Impacted MAS Link:**
[MAS 1.4.1 – Use of Color (16)](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={9f1f0271-42a1-45b1-be00-31473ba2e579})

**Attachments:**
![mas1 4 1_botframeworkemulator_welcomepage_links](https://user-images.githubusercontent.com/41668935/43773318-faa9e4b2-9a62-11e8-8e3a-7176f503e0f0.png)

corinagum:
@mewa1024 Could you let me know how we want to update this one? Should the links have underlines, or something else? Thank you.

mewa1024:
@corinagum Underline for now.

PR #827, opened 2018-09-05
Update welcome page to match design

corinagum:
Resolves [#772:](https://github.com/Microsoft/BotFramework-Emulator/issues/772)
Resolves [#737](https://github.com/Microsoft/BotFramework-Emulator/issues/737)

- New right side panel on welcome page
- Copy & styling updates
- Adds ability to login / logout of Azure on Welcome page


packages/app/client/src/ui/editor/welcomePage/welcomePage.tsx
```diff
@@ -74,65 +107,207 @@ export class WelcomePage extends React.Component<WelcomePageProps, {}> {
...
+              </dd>
+            </dl>
+          </div>
+        </div>
+      </div>
     );
   }
 }
+
+const mapStateToProps = (state: RootState): WelcomePageProps => ({
```
justinwilaby: Isn't there already a `welcomPageContainer.ts` ?

packages/app/client/src/ui/editor/welcomePage/welcomePage.tsx
```diff
@@ -74,65 +107,207 @@ export class WelcomePage extends React.Component<WelcomePageProps, {}> {
...
               this.props.recentBots.slice(0, 10).map(bot => bot &&
-                <li className={ styles.recentBot } key={ bot.path }>
-                  <a href="javascript:void(0);" onClick={ ev => onBotClick(ev, bot.path) }
-                      title={ bot.path }><TruncateText>{ bot.displayName }</TruncateText></a>
-                  <TruncateText className={ styles.recentBotPath }
-                                title={ bot.path }>{ bot.path }</TruncateText>
-                <div className={ styles.recentBotActionBar }>
-                  <button onClick={ ev => onDeleteBotClick(ev, bot.path) }></button>
+                <li className={styles.recentBot} key={bot.path}>
+                  <button onClick={ev => onBotClick(ev, bot.path)}
```
justinwilaby: Creating a pointer function with each render will cause React to redraw because a new reference is created and the in-memory VDOM will not be equivalent.  e.g.
```js
(t => {}) === (t => {}) // false
```
All handlers should be class members.

packages/app/client/src/ui/editor/welcomePage/welcomePage.tsx
```diff
@@ -74,65 +107,207 @@ export class WelcomePage extends React.Component<WelcomePageProps, {}> {
...
+              Sign in with your Azure account.
+            </button>
+        }</div>
     );
   }
 
-  public render(): JSX.Element {
-    const { startSection, myBotsSection, helpSection } = this;
+  private signInWithAzure() {
+    CommandServiceImpl.call(UI.SignInToAzure);
```
justinwilaby: Calls to external dependencies should be managed in the `Container` to increase encapsulation and make the component more testable.  Also, it looks like the `Promise` returned from this call is not handled.

packages/app/client/src/ui/editor/welcomePage/welcomePage.tsx
```diff
@@ -74,65 +107,207 @@ export class WelcomePage extends React.Component<WelcomePageProps, {}> {
...
+  }
+
+  private signOutWithAzure() {
+    CommandServiceImpl.call(Azure.SignUserOutOfAzure);
```
justinwilaby: Same here. External service dependencies should be contained within the `Container` and the container should be its own file.

packages/app/client/src/ui/media/build_plan/bp_ic_01.svg
```diff
@@ -0,0 +1 @@
...-2h7V49H49Z"/><path class="cls-1" d="M17.14,42.15l2.78-10.62L48.18,3.47a5.6,5.6,0,0,1,7.76,0,5.56,5.56,0,0,1,0,7.73L28,39.28Zm4.57-9.59-1.77,6.78,7-1.86L54.53,9.81a3.55,3.55,0,0,0,0-4.92,3.59,3.59,0,0,0-5,0Z"/><circle class="cls-1" cx="41.49" cy="52.5" r="1.5"/><circle class="cls-1" cx="33.82" cy="52.5" r="1.5"/><circle class="cls-1" cx="26.16" cy="52.5" r="1.5"/><circle class="cls-1" cx="18.49" cy="52.5" r="1.5"/><circle class="cls-1" cx="33.82" cy="7.49" r="1.5"/><circle class="cls-1" cx="26.16" cy="7.49" r="1.5"/><circle class="cls-1" cx="18.49" cy="7.49" r="1.5"/><rect class="cls-1" x="48.67" y="5.44" width="2" height="8.89" transform="translate(7.81 38.39) rotate(-45.51)"/><polygon class="cls-2" points="21.43 37.8 26.14 36.54 49.85 13 46.3 9.46 22.65 33.31 21.43 37.8"/><circle class="cls-1" cx="7.52" cy="41.57" r="1.5"/><circle class="cls-1" cx="7.52" cy="33.9" r="1.5"/><circle class="cls-1" cx="7.52" cy="26.24" r="1.5"/><circle class="cls-1" cx="7.52" cy="18.57" r="1.5"/><circle class="cls-1" cx="52.51" cy="41.57" r="1.5"/><circle class="cls-1" cx="52.51" cy="33.9" r="1.5"/><circle class="cls-1" cx="52.51" cy="26.24" r="1.5"/><rect class="cls-3" width="60" height="60"/></svg>
```
justinwilaby: these SVG's contain fill color values which might make them hard to apply themes to. Consider using `-webkit-mask` and `background-color` to apply themes to these svgs.
Also, it looks like these files lack the xml declaration . Probably a non-issue but we may want to add them so they are valid XML documents

packages/app/main/src/appMenuBuilder.ts
```diff
@@ -189,7 +189,10 @@ export const AppMenuBuilder = new class AppMenuBuilderImpl implements AppMenuBui
       label: azureMenuItemLabel,
       click: () => {
         if (signedInUser) {
-          return mainWindow.commandService.call(Azure.SignUserOutOfAzure);
+          return Promise.all([
+            mainWindow.commandService.call(Azure.SignUserOutOfAzure),
+            mainWindow.commandService.remoteCall(UI.InvalidateAzureArmToken)
```
justinwilaby: The call to invalidate the arm token should probably me made as a `yield` call in the `AzureAuthWorkflowService.enterSignOutWorkflow` or as the last step in the  `Azure.SignUserOutOfAzure` command generator since they are tightly coupled.

packages/app/client/src/ui/media/build_plan/bp_ic_01.svg
corinagum: Added the xml declaration. As far as applying themes, Marc mentioned that he designed these specifically to work with all themes, so I think that would be unnecessary here.

Issue #839, opened 2018-09-10
App ID and password not working on Emulator Core


ckkashyap:
This commit broke it, https://github.com/Microsoft/BotFramework-Emulator/commit/e915ec1dd2bab1750d18bfa1ade2c27586d7c08e#diff-c02a923b1d3cf43295b9353b9c20e2fe.

# Repro
1. Run a bot with MSA ID and password
2. Run the Emulator CLI with the same ID and password
3. Run Web Chat, pointing to the Emulator Core
4. Send "hi"

## Expected
It should works, the bot should response

## Actual
When Emulator Core verify the signature from POST /directline/conversations/…/activities, the Emulator Core `jsonwebtoken` think it is wrong and return 401 back to the bot.

PR #840, opened 2018-09-10
Fix for bug #839

ckkashyap:
Issue: https://github.com/Microsoft/BotFramework-Emulator/issues/839

Fix: Undid the part of the [change](https://github.com/Microsoft/BotFramework-Emulator/commit/e915ec1dd2bab1750d18bfa1ade2c27586d7c08e#diff-c02a923b1d3cf43295b9353b9c20e2fe) that introduced the faulty extra base64 encoding.


packages/emulator/core/src/utils/openIdMetadata.ts
```diff
@@ -90,7 +90,7 @@ export default class OpenIdMetadata {
           return null;
         }
 
-        return getPem(Buffer.from(key.n).toString('base64'), key.e);
+        return getPem(base64url.toBase64(key.n), key.e);
```
justinwilaby: Node has full support for Base64 encoding. How are these different?

justinwilaby: Also, are you able to use the `rsa-pem-from-mod-exp` library as in [this example](https://github.com/Microsoft/BotFramework-Emulator/blob/v4/packages/app/main/src/services/azureAuthWorkflowService.ts#L257)?

compulim: Base64 has [10+ variants](https://en.wikipedia.org/wiki/Base64#Variants_summary_table). Node.js support the standard one (using `+`, `/`, and `=`). But OpenID is using the Base64 URL-variant.

The original code will do the following:
- Replace `-` with `+`
- Replace `_` with `/`
- Pad `=` as needed

The original code will not do the actual Base64 encoding. It just do simple string replacement as mentioned above. That's why we are calling `base64url.toBase64`, instead of `base64url.encode`.

PR #841, opened 2018-09-10
Fix #606 Use endpoint ID

### LINKED ISSUE #606: Microphone is not working in bot Emulator
my microphone is not working in bot emulator 

I have registered it , and also have given AppID and password . but microphone doesnt work 

need your input

compulim:
This will fix #606. We should use `endpoint.id` to identify an endpoint instead of `endpoint.endpoint`.

### DIFF
--- a/packages/app/client/src/ui/editor/emulator/parts/chat/chat.spec.ts
+++ b/packages/app/client/src/ui/editor/emulator/parts/chat/chat.spec.ts
@@ -1,0 +1,31 @@
+import { CommandServiceImpl } from '../../../../../platform/commands/commandServiceImpl';
+import { getSpeechToken } from './chat';
+
+jest.mock('../../../../dialogs', () => ({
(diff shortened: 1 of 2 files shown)


cleemullins:
@compulim Please put descriptions into the PRs. This should include the link to the Issue. 

Note also that if the description is correct, GitHub will autoclose associated issues. 
https://blog.github.com/2013-05-14-closing-issues-via-pull-requests/

cleemullins:
@compulim Also, code changes mean a unit should have changed. Please add a test to validate this behavior.

compulim:
Added unit tests

compulim:
Merged even though the coverage is dropped because:
- Our code change should not drop coverage (it's a property name change)
- We introduced more unit tests in the code path, it should keep the coverage or increase it

So I think the coverage report is bugged for this PR.

Issue #798, opened 2018-08-29
Not clear error message on first run post install


yochay:
Loading the emulator with a bot file shows the following message. 
1. the text is not properly wrapped
2. not clear what the error is and how/what it is related to my bot

Previously I had the Build version.

![image](https://user-images.githubusercontent.com/896149/44759058-b7f39d80-aaec-11e8-907a-ee92a2bcb21f.png)

tonyanziano:
This issue has to do with post-v4 upgrade migration logic. I can look into this one.

tonyanziano:
Addressed in #844

justinwilaby:
Closing this issue since the fix was merged.

PR #844, opened 2018-09-12
Disabled migration on fresh v4 installs; fixed notification text wrapping

tonyanziano:
Fix for #798 

=========

Migration will no longer run for fresh V4 installs since it will now first check if the `%emulator_data%/migration/` folder exists. (which should only ever exist post-v3-migration)

Also fixed text wrapping inside of notifications so that text can longer bleed out of the notification element

### DIFF
--- a/packages/app/main/src/migrator.ts
+++ b/packages/app/main/src/migrator.ts
@@ -59,7 +59,7 @@ export class Migrator {
   /** Adds the bot files in the /migration/ dir
    *  to the MRU bots list and displays an overview page
    */
-  private static async migrateBots(): Promise<boolean> {
+  public static async migrateBots(): Promise<boolean> {
     const botFilesDirectory = Path.join(ensureStoragePath(), 'migration');
     // if the /migration/ directory does not exist then abort migration
     if (!Fs.existsSync(botFilesDirectory)) {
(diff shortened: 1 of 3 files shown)


packages/app/main/src/migrator.ts
```diff
@@ -49,17 +49,24 @@ export class Migrator {
...
   /** Adds the bot files in the /migration/ dir
    *  to the MRU bots list and displays an overview page
    */
-  private static async migrateBots(): Promise<void> {
-    // read bots from directory
-    const botFiles = (getFilesInDir(Path.join(ensureStoragePath(), 'migration')) || []) as string[];
+  public static async migrateBots(): Promise<boolean> {
+    const botFilesDirectory = Path.join(ensureStoragePath(), 'migration');
+    // if the /migration/ directory does not exist then abort migration
+    if (!Fs.existsSync(botFilesDirectory)) {
```
justinwilaby: F.Y.I Using `fs-extra` automatically promisifies all async FS operations:
```js
import * as fs from 'fs-extra';
// ...
if (!await Fs.exists(botFilesDirectory)) {
```
It's fine like this though too!

tonyanziano: Good to know!

Issue #716, opened 2018-07-31
[MAS4.2.1] Narrator is reading "heavy multiplication x" at 'Welcome' button which is irrelevant. 


priysn:
**Actual Result:**
Narrator reads "heavy multiplication x" when focus reach at Welcome' button which is irrelevant.

**Expected Result:**
Narrator should not read "heavy multiplication x" when focus reach at Welcome' button because this text is irrelevant and unnecessary.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Start Narrator using Ctrl + Windows Logo Key +Enter.
3. Navigate to 'Welcome' tab button.
4. Notice what Narrator reads for this button.

**User Impact:**
Screen Reader dependent users may get confuse due to any irrelevant content being announced at any element.

**Test Environment:**
OS: Windows 10
OS Build: 17134.167
OS Version: 1803
Browser: Edge

**Impacted MAS Link:**
[MAS 4.2.1 – Object Information (33A)](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={a12f3779-bdf1-40c2-bbcc-0e314ea94858})

**Attachment:**
[MAS4.2.1_BotFrameworkEmulator_Welcome_IrrelevantContent.zip](https://github.com/Microsoft/BotFramework-Emulator/files/2245349/MAS4.2.1_BotFrameworkEmulator_Welcome_IrrelevantContent.zip)

PR #859, opened 2018-09-13
Toanzian/#716 - Fixed irrelevant narration on tab close icon by screen reader

tonyanziano:
Fix for #716 

==========

Tab close icons now use a tintable svg mask instead of the `:after` pseudo element. Will now read 'Close' when tab is focused by the narrator.


packages/app/client/src/ui/shell/mdi/tab/tab.scss
```diff
@@ -44,37 +47,32 @@
 
   &:hover {
     > .editorTabClose {
-      opacity: 1;
+      visibility: visible;
```
corinagum: Does this not break the keyboard accessibility of the close button?

packages/app/client/src/ui/shell/mdi/tab/tab.tsx
```diff
@@ -77,9 +77,12 @@ class TabComponent extends React.Component<TabProps, TabState> {
         { this.props.dirty ? <span>*</span> : null }
         <a
           href="javascript:void(0)"
+          title="Close"
           className={ styles.editorTabClose }
           onKeyPress={ this.onCloseButtonKeyPress }
-          onClick={ this.props.onCloseClick }>
+          onClick={ this.props.onCloseClick }
+          >
+          <span style={{ '-webkit-mask': `url(${closeIcon})` } as any }></span>
```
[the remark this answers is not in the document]
tonyanziano: Wasn't sure if the url loader would work if it was in the SCSS file. I'll try that.

packages/app/client/src/ui/shell/mdi/tab/tab.scss
tonyanziano: I just tested it with the narrator and it works fine. It's actually better than opacity, because with opacity the close button was still focusable even on an inactive tab.

Issue #728, opened 2018-08-03
[NARRATOR] Narrator's behavior is confusing in Right side pane of Bot Framework Emulator while using either scan mode or item mode.


priysn:
**Actual Result:**
Narrator reads each element present in this section separately, does not announce anything at some elements, read some hidden confusing content as well.

**Expected Result:**
Narrator should be able to read all the content logically and efficiently.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Start Narrator using Ctrl + Windows Logo Key +Enter.
3. Navigate in Right side pane of Bot Framework Emulator using Scan mode (Up/Down arrow keys) or item mode (Caps lock + Right/ Left arrow keys).
4. Notice Narrator's behavior at elements present in this section.

**User Impact:**
Screen Reader dependent users may get confuse if Narrator read any hidden confusing information, or just don't announce anything in any keystroke.

**Test Environment:**
OS: Windows 10
OS Build: 17134.167
OS Version: 1803
Browser: Edge
Tool: Narrator

**Attachment:**
[NARRATOR_BotEmulatorFramework_Endpoint_TranscriptExplorerControls.zip](https://github.com/Microsoft/BotFramework-Emulator/files/2257532/NARRATOR_BotEmulatorFramework_Endpoint_TranscriptExplorerControls.zip)

tonyanziano:
Seems to be outdated with the new explorer panels, but there are still some lingering bugs that should be resolved:

 - explorer panel list items should be focusable and be narrated clearly
 - explorer headers should be narrated clearly as well

tonyanziano:
Fixed in #864

PR #864, opened 2018-09-14
Improved screen reader narration for explorer panes and accessory buttons

tonyanziano:
Fix for #728 

=============

Added ARIA tags to a lot of the explorer panel components, and changed markup for `ExpandCollapse` component.


packages/sdk/ui-react/src/layout/expandCollapse/expandCollapse.tsx
```diff
@@ -107,13 +111,13 @@ export class ExpandCollapse extends React.Component<ExpandCollapseProps, ExpandC
     }
   }
 
-  private onActuatorClick = () => {
+  private toggleExpandedState = () => {
```
justinwilaby: Event handler names in JS start with `on` by convention and generally follows this:
`on<elementName | elementPurpose + elementName ><eventType>` or `on<noun | verb+noun><eventType>`
so `onButtonClick` or `onToggleExpandButtonClick` would be acceptable. Using the latter is recommended when there is more than 1 button .

tonyanziano: 👍

PR #904, opened 2018-09-19
Fixed bug where padlock wasn't being generated on newly-encrypted bots.

tonyanziano:
Fixed several encryption-related issues:

1. If an old bot was being encrypted via the bot settings dialog, the `bots.json` wasn't being updated before the main side was trying to read it for the secret, causing the secret not to be validated, and therefore, not generating a `padlock` within the bot file.

2. If the bot didn't have a secret when the bot settings window was opened, a secret was being generated automatically via the `get generatedSecret()` function. The correct behavior is that the `encryptKey` state should default to false, and that the secret input was blank with placeholder text.

3. When checking the encryption checkbox, the input should default to plain text so that the newly generated secret is shown

### DIFF
--- a/packages/app/client/src/ui/dialogs/botSettingsEditor/botSettingsEditor.tsx
+++ b/packages/app/client/src/ui/dialogs/botSettingsEditor/botSettingsEditor.tsx
@@ -67,7 +67,7 @@ export class BotSettingsEditor extends React.Component<BotSettingsEditorProps, B
     const secret = (botInfo && botInfo.secret);
     this.state = {
       ...bot,
-      secret: secret || this.generatedSecret,
+      secret: secret,
       revealSecret: false,
       encryptKey: !!secret
     };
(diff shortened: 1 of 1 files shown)


packages/app/client/src/ui/dialogs/botSettingsEditor/botSettingsEditor.tsx
```diff
@@ -203,13 +209,15 @@ export class BotSettingsEditor extends React.Component<BotSettingsEditorProps, B
   /** Saves a bot config of a bot loaded from disk */
   private saveBotFromDisk = async (bot: BotConfigWithPath): Promise<void> => {
     const { Save, PatchBotList } = SharedConstants.Commands.Bot;
+    // write updated bot entry to bots.json so main side can pick up possible changes to secret
     const botInfo: BotInfo = getBotInfoByPath(bot.path) || {};
     botInfo.secret = this.state.secret;
-    // write updated bot entry to bots.json
+    await CommandServiceImpl.remoteCall(PatchBotList, bot.path, botInfo);
```
justinwilaby: We should revisit this post ignite. Business logic should belong in a container or better yet, a saga.

tonyanziano: Sounds good to me. 👍

PR #905, opened 2018-09-19
Explorer pane empty content update & splitter border

### LINKED ISSUE #892: Emulator strings/ style scrub
* The development and production "buttons" don't look like buttons and their icon should be a + because they create a new tab. On first coming into the tool it looks like you switch between the two environments.
* "You have not saved any Services apps to this bot"... what does that mean.
* Start Over and Save Transcript As are buttons (do something when you click them) but they look like text. In the lower "Log" section the text looks like links (ie POST and 202) but they aren't links/buttons.
* Fuzzy icons
* Missing border between chat window and log/inspector... the explorer has them.
* Two different buttons that do different things use the same gear icon

image

corinagum:
Resolves #892

### DIFF
--- a/packages/app/client/src/ui/shell/explorer/servicesExplorer/servicesExplorer.tsx
+++ b/packages/app/client/src/ui/shell/explorer/servicesExplorer/servicesExplorer.tsx
@@ -51,6 +51,7 @@ import { serviceTypeLabels } from '../../../../utils/serviceTypeLables';
 export interface ServicesExplorerProps extends ServicePaneProps {
   services?: IConnectedService[];
   toAnimate?: { [serviceId: string]: boolean };
+  onAnchorClick: (url: string) => void;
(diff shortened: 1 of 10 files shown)


packages/app/client/src/ui/shell/explorer/servicesExplorer/servicesExplorer.tsx
```diff
@@ -91,6 +92,42 @@ export class ServicesExplorer extends ServicePane<ServicesExplorerProps> {
...
+      <div>
+        <p className={ styles.emptyContent }>None.</p>
+        <p className={ styles.emptyContent }>
+          You can connect your bot to services such as &nbsp;
+          <a
+            href="javascript:void(0);"
+            onClick={ this.onLearnMoreLUISAnchor }
+          >
+            Language Understanding (LUIS)
+            </a>, &nbsp;
```
tonyanziano: align with opening `<a> tag`

PR #918, opened 2018-09-20
Jwilaby/#696 blank menu item mac

justinwilaby:


packages/app/main/src/appMenuBuilder.ts
```diff
@@ -254,9 +254,13 @@ export const AppMenuBuilder = new class AppMenuBuilderImpl implements AppMenuBui
...
         { role: 'paste' },
+        process.platform === 'darwin' ? {
+          role: 'pasteAndMatchStyle'
+        } : null,
         { role: 'delete' },
-        { role: 'selectall' },
-      ]
+        process.platform === 'win32' ? { type: 'separator' } : null,
+        { role: 'selectall' }
+      ].filter(item => item) as any[]
```
tonyanziano: So where was the blank item coming from? Was it just an unfiltered item? Or was it the selectall?

justinwilaby: I think it was the select all `selectAll` does not work while `selectall` does. also, it looks like separators are conditional based on the platform and where they fall in the menu item ordering.

tonyanziano: Ah I see. 👍

justinwilaby: Full disclosure, this hasn't been tested on a Mac per se, but the interwebs say this is on fleek!

justinwilaby: image

PR #954, opened 2018-10-02
Fixed bundling issue caused by Webpack 4.19.1 & HardSourceWebpackPlugin.

tonyanziano:
Locked webpack to `4.16.4`

===================

webpack `>=4.19.1` breaks default exports when using `hard-source-webpack-plugin`

Reference: https://github.com/mzgoddard/hard-source-webpack-plugin/issues/443


compulim:
Thanks for fixing it. It must have took you a long time for the investigation.

Someone just mentioned `@4.19.0` also works. But I am fine to keep it to `@4.16.4` for now.

Can you keep track of the [original thread](https://github.com/mzgoddard/hard-source-webpack-plugin/issues/443) so we know when to unlock it?

tonyanziano:
@compulim I've subscribed to the PR request for hard source to fix it (https://github.com/mzgoddard/hard-source-webpack-plugin/pull/444) as well as the issue over on webpack (https://github.com/webpack/webpack/issues/8052)

PR #956, opened 2018-10-03
Jwilaby/emulator core unit tests

justinwilaby:
This adds 37 new unit tests to the emulator/core package


justinwilaby:
This PR also corrects errant reporting for built packages which inflates the coverage percentages

Issue #863, opened 2018-09-14
Emulator: sidebar items, change the action for opening items from single click to double click


mewa1024:
Launching deep links on single-click makes for a twitchy experience.

justinwilaby:
Fixed with #968

PR #968, opened 2018-10-09
#863 - Change deep links to require a double click

justinwilaby:


packages/app/main/src/services/contextMenuService.ts
```diff
@@ -53,7 +53,7 @@ export class ContextMenuService {
       });
       const menu = ContextMenuService.currentMenu = Menu.buildFromTemplate(template);
 
-      menu.popup();
+      menu.popup({});
```
tonyanziano: You beat me to it. I'll close my PR #969.

Can you add the changes I had in this file as well?

1. Remove `MenuItem` from the `import()` statement
2. Provide default value of `[]` for `options` argument

justinwilaby: 👍

PR #969, opened 2018-10-09
Fixed service context menu broken by electron v2.x.x

tonyanziano:
`menu.popup()` now requires an options argument


tonyanziano:
Dupe of #863

Issue #976, opened 2018-10-11
Sign out should sign me out and not remember me if I don't explicitly ask to be remembered


carlosscastro:
Steps:

- Sign in to Azure
- Do NOT choose "keep me signed in"
- Sign out
- Sign in again: my credentials are remembered and I'm automatically signed in, which I don't want. 

The "keep me signed in" checkbox is ignored and assumed always true perhaps?

justinwilaby:
Thank for opening this discussion. 

The purpose of remember me is not to control whether or not the user is prompted for credentials the login tool. Its purpose is to prevent the user from having to login again if the emulator is closed and reopened at a later time. To implement what you describe, we would need two checkboxes, one to remember credentials in the sign in tool and another to remember the user's azure login at the emulator. The two are mutually exclusive and we currently implement only the latter.

carlosscastro:
The current state doesn't feel right because if I sign in with my work account by mistake, then I sign out with the intention of signing in with my personal account, I cannot - the emulator simply chooses to sign me in to an account that i did not ask to be remembered and I don't want to sign in to.

I got frustrated because of this and I imagined our users might feel equally frustrated. Whether it meets the bar to get it fixed or not, I'm not sure! Lets dicuss in standup :)

carlosscastro:
If we leave the checkboxes as they are, but sign out reprompts for which account you want to sign in, I would be fine with that. That is a correct behavior, and seems easy enough to fix.

justinwilaby:
Yes, setting the `prompt=consent` seems to work. I can implement this

carlosscastro:
@justinwilaby Sure! However when I find these things, I can help fix them too, no need for you to jump in, we can help :) I file it as issue so we get to discuss what the right thing is 👍

justinwilaby:
Fixed with #977

carlosscastro:
Reopening since our fix for this unfortunately was not what we need. We need to figure out what to do (probably will take a bit) and do it (probably a few lines of code, jsut hard to know which!).

carlosscastro:
after we make ours a first party app, we may be able to have consent work and re-apply

PR #977, opened 2018-10-11
#976 - azure sign in tool now prompts for username with each login

justinwilaby:
#976 

Fixed

### DIFF
--- a/packages/app/main/src/services/azureAuthWorkflowService.ts
+++ b/packages/app/main/src/services/azureAuthWorkflowService.ts
@@ -47,7 +47,7 @@ export class AzureAuthWorkflowService {
   private static config: Config;
   private static jwks: Jwks;
 
-  public static * retrieveAuthToken(renew: boolean = false, redirectUri: string): IterableIterator<any> {
+  public static* retrieveAuthToken(renew: boolean = false, redirectUri: string): IterableIterator<any> {
     const authWindow = yield this.launchAuthWindow(renew, redirectUri);
     authWindow.show();
     const result = yield this.waitForAuthResult(authWindow, redirectUri);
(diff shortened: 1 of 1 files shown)


packages/app/main/src/services/azureAuthWorkflowService.ts
```diff
@@ -92,8 +92,8 @@ export class AzureAuthWorkflowService {
         const values = uri.substring(idx + 1).split('&');
         const len = values.length;
         for (let i = 0; i < len; i++) {
-          const [key, value] = values[i].split(/[=]/);
-          result[key] = value;
+          const [ key, value ] = values[ i ].split(/[=]/);
```
carlosscastro: This feels weird, I prefer values[i]. Sure you want this edit?

packages/app/main/src/services/azureAuthWorkflowService.ts
```diff
@@ -92,8 +92,8 @@ export class AzureAuthWorkflowService {
...
+          result[ key ] = value;
```
carlosscastro: Same here, not a fan of this, not sure if it is a standard in js but it looks like an editor automated thing

packages/app/main/src/services/azureAuthWorkflowService.ts
```diff
@@ -247,7 +246,7 @@ export class AzureAuthWorkflowService {
   }
 
   private static async validateJWT(token: string): Promise<boolean> {
-    const [header] = token.split('.');
+    const [ header ] = token.split('.');
```
carlosscastro: same here, is this on purpose?

justinwilaby: Auto format - rules might have changed unintentionally.

Issue #546, opened 2018-05-16
Auto-update isn't working on Mac


emadelwany:

tonyanziano:
Addressed in #996

PR #996, opened 2018-10-17
Fixed signing & auto update on Mac build.

tonyanziano:
Fix for #546 

====

- Fixed signing for Mac.
- Minor changes to auto update UI & flow
- Removed redundant / unused code in mac and windows build processes
- Fixed some issues with nightly versioning

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-	"name": "BotFramework-Emulator",
+	"name": "EmuGH",
 	"requires": true,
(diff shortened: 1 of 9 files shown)


tonyanziano:
There is a currently known issue on Mac with `electron-updater@3.1.x` that prevents electron from quitting and installing the update after downloading. So I downgraded to `3.0.3`.

https://github.com/electron-userland/electron-builder/issues/3269

PR #1005, opened 2018-10-19
Auth: fix consent blocking MSFt employees sign in

carlosscastro:


carlosscastro:
We can find another fix for the issue where the user is remembered and you cannot sign in to another acct. I'll reopen that one since this was part of its fix, which had this side effect. Glad we found it!

PR #1015, opened 2018-10-23
New V3 Build Infrastructure

tonyanziano:
Overhaul of V3 build infrastructure.


build/scripts/utils.js
```diff
@@ -0,0 +1,11 @@
+function getEnvVar(varName, defaultValue) {
```
compulim: Since `getEnvVar` is always used with `defaultValue` (in other 3 script files), it is impossible to throw an error.

What's more, the variable is named "default value", which means: "if it's not set in environment variable, use this value." And it will be strange to throw the error "oh, no value found, you should have set it!"

Then, this code would become `process.env[varName]`, which defeated its purpose (of existence).

build/scripts/linux.js
```diff
@@ -0,0 +1,43 @@
+const { getEnvVar } = require('./utils');
```
compulim: Looks like this file, `mac.js`, and `windows.js` are almost the same with very little change.

Unlike production code, I don't recommend to merge them together. But it could be a good idea to refactor helper function `hashFileAsync` and `writeYamlMetadataFile` out to a separate file, because they are "stateless".

build/scripts/utils.js
tonyanziano: Good point! I'll just go with `process.env[varName] || 'mac / linux / windows'`

build/scripts/linux.js
tonyanziano: 👍

PR #1017, opened 2018-10-23
styling for client cert select dialog

corinagum:


packages/app/client/src/ui/dialogs/clientCertSelectDialog/clientCertSelectDialog.scss
```diff
@@ -1,3 +1,72 @@
...
+        background-color: var(--neutral-16);
+    }
+}
+
+:global {
+    ::-webkit-scrollbar-corner {
+        background-color: transparent; 
+        // TO DO: confirm invisible scrollbar with design
+    } 
 }
```
tonyanziano: newline at end of document

packages/app/client/src/ui/dialogs/clientCertSelectDialog/clientCertSelectDialog.tsx
```diff
@@ -31,23 +31,26 @@ export class ClientCertSelectDialog extends Component<ClientCertSelectDialogProp
               this.props.certs.map((cert, index) => {
                 return (
                   <li 
-                    className={ selectedIndex === index ? 'SelectedClassName' : 'nonSelectedClassName'}
+                    className={ selectedIndex === index ? styles.selectedCert : ''}
```
justinwilaby: This list probably needs accessibility.
An option would be to use `aria-selected` for the selected `<li>` tag and `li[aria-selected="true"]` on the css side.

corinagum: Roger that, thanks for catching this. I'm also adding role="listbox" and "option" to the ul and li respectively. 

I have a clarification question: is there specific styling I should be doing for the selected certificate (that I forgot), or do you in this case mean to change selected-cert (the class name) to li[aria-selected="true]?

corinagum: (Got Tony to check this out. Let me know if you want me to further fix later when I do unit tests!)

PR #1020, opened 2018-10-24
Fixed v3 -> v4 migration issues

tonyanziano:
Fix for #505 

=======

Found out that since the package.json `name` differs between Emulator v3 and v4:

**v3**: `botframework-emulator`
**v4**: `@bfemulator/main`

then the app data folder where we store emulator files is also different because it corresponds to the electron app name which defaults to this property:

**v3**: `%APPDATA%/botframework-emulator/botframework-emulator/...`
**v4**: `%APPDATA%/@bfemulator/main/botframework-emulator/...`

This means that the v4 version needs to look for the bots to be migrated in the v3 app data folder.

So this PR adds that mapping.

### DIFF
--- a/packages/app/main/src/migrator.ts
+++ b/packages/app/main/src/migrator.ts
@@ -31,6 +31,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
+import { app } from 'electron';
 import * as Fs from 'fs';
 import * as Path from 'path';
 import * as BotActions from './botData/actions/botActions';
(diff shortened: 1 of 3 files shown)


tonyanziano:
Unit tests are WIP

packages/app/main/src/migrator.ts
```diff
@@ -73,10 +88,23 @@ export class Migrator {
         const botFile = botFiles[i];
         // read the bot file and create a bot info item from it
         try {
-          const path = Path.join(ensureStoragePath(), 'migration', botFile);
-          const bot = await BotConfiguration.load(path);
+          // v3 path
+          const oldPath = Path.join(botFilesDirectory, botFile);
+          // v4 path
+          const newPathBase = Path.join(ensureStoragePath(), 'migration');
+          if (!Fs.existsSync(newPathBase)) {
```
justinwilaby: it's a good idea to get into the habit of async operations for disk IO. It's not likely to have a huge impact here but you can use `fs-extra` as an `await`able alternative to `fs`
```js
import * as fs from 'fs-extra';
if (await fs.exists(newPathBase)) {}
```

packages/app/main/src/migrator.ts
```diff
@@ -107,6 +135,6 @@ export class Migrator {
 
   /** Checks for the migration marker to determine if it has already been performed */
   private static get migrationHasBeenPerformed(): boolean {
-    return Fs.existsSync(Path.join(ensureStoragePath(), 'migration_marker.txt'));
+    return Fs.existsSync(Path.join(ensureStoragePath(), this._migrationMarkerName));
```
justinwilaby: Here as well. Consider using async disc IO.

packages/app/main/src/migrator.ts
tonyanziano: Is there a benefit to awaiting an async operation over calling the sync version? Does the sync version completely lock the main thread and awaiting the async doesn't?

justinwilaby: Yes, that's correct. Disk reads are slow and systems under load are susceptible to periods of nonresponsivness when using sync. Again, this isn't a big deal here but if you were building a middleware for a high traffic service, sync reads are a big problem.

tonyanziano: Got it 👍

PR #1095, opened 2018-11-15
Refresh conversation menu when bot opens/closes

### LINKED ISSUE #801: Conversation -> send system activity should be disabled when no bot is active
Right now, the menu is always enabled. We should have this disabled when no bot is active. 

image

corinagum:
Fixes #801

### DIFF
--- a/packages/app/main/src/main.ts
+++ b/packages/app/main/src/main.ts
@@ -80,12 +80,13 @@ if (app) {
 // -----------------------------------------------------------------------------
 // App-Updater events
 
-AppUpdater.on('checking-for-update', (...args) => {
-  AppMenuBuilder.refreshAppUpdateMenu();
+AppUpdater.on('checking-for-update', async (...args) => {
+  await AppMenuBuilder.refreshAppUpdateMenu();
 });
 
-AppUpdater.on('update-available', (update: UpdateInfo) => {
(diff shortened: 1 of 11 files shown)


packages/app/main/src/main.ts
```diff
@@ -306,8 +308,9 @@ const createMainWindow = async () => {
   // Start auto-updater
   AppUpdater.startup();
 
-  const template: Electron.MenuItemConstructorOptions[] = AppMenuBuilder.getAppMenuTemplate();
-  Menu.setApplicationMenu(Menu.buildFromTemplate(template));
+  AppMenuBuilder.getMenuTemplate().then((template: Electron.MenuItemConstructorOptions[]) => {
+    Menu.setApplicationMenu(Menu.buildFromTemplate(template));
+  });
```
tonyanziano: convert this to an await?

packages/app/shared/src/constants/sharedConstants.ts
```diff
@@ -161,5 +162,8 @@ export const SharedConstants = {
       UpdateProgressIndicator: 'shell:updateProgressIndicator',
       InvalidateAzureArmToken: 'shell:invalidateAzureArmToken',
     }
-  }
+  },
+    Content: {
```
tonyanziano: nitpick: but I would probably call this `ContentTypes`

PR #1107, opened 2018-11-19
Fix accessibility of links on Welcome Page by converting anchors to buttons

### LINKED ISSUE #1096: Accessibility: "How to build a bot" section of Welcome Page [Regression]
## Version
Latest master

## Describe the bug
"How to build a bot" section on the welcome page is no longer keyboard accessible. This is a regression from previous functionality

## To Reproduce
1. Open Emulator
2. Tab through the different buttons and links on the Welcome Page
3. Note that after "Sign in with your Azure account", the keyboard focus jumps back to the Bot Explorer button instead of going to "How to built a bot" section

## Expected behavior
After tabbing to "Sign in with your azure account", the next keyboard focus item should be "design guidelines".

[bug]

corinagum:
Fixes #1096 
This changes the buttons-that-look-like-anchors to be buttons since there is no href tag, and therefore these are technically not links. This fixes the non-keyboard accessible issue that I came across.

### DIFF
--- a/packages/app/client/src/ui/editor/welcomePage/welcomePage.tsx
+++ b/packages/app/client/src/ui/editor/welcomePage/welcomePage.tsx
@@ -41,7 +41,7 @@ import { GenericDocument } from '../../layout';
 export interface WelcomePageProps {
   accessToken?: string;
   documentId?: string;
-  onAnchorClick: (url: string) => void;
+  onLinkClick: (url: string) => void;
   onNewBotClick?: () => void;
   onOpenBotClick?: () => void;
   onBotClick?: (_e: any, path: string) => void;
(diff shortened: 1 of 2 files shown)


tonyanziano:
Based on what @justinwilaby said in #1108 , I would attempt converting one of the links to 

`<a href="https://my.link.com" target="_blank">my link</a>`

without a handler and see if the link successfully opens in an external window. Then we won't need to use the handle external link method and we get the expected accessibility of an anchor tag for free.

justinwilaby:
@corinagum -

I'd like to approach this in another way. Anchors should remain anchors with no click handlers and no `data-href`bs...just normal HTML links. 

In the `index.tsx`, I'd like to add this code:

```js
document.addEventListener('click', (event: MouseEvent): void => {
  const { target } = event;
  const { href } = target;
  if (target.tagName !== 'A' || !href || href.includes('javascript:void(0)')) {
    return;
  }
  event.preventDefault();
  CommandService.remoteCall(SharedConstants.Commands.Electron.OpenExternal, href);
}
```

This should fix our issue and allow us to use semantically correct html

Issue #1026, opened 2018-10-25
[Feature Request] More Control Over Auto Update


tonyanziano:
## Version
4+

## Describe the bug
Now that we have nightly builds going and auto update functioning, we need to consider adding UI that gives the user more control over when and to what version they update.

Currently, we plan to enable auto update, but only have it target the latest release and not pre-releases (nightly). However, we do not provide any control over this option through the UI, which means that people who download the nightlies will be auto-updated **backwards** to the latest release.

**Example:**
`Nightly: 4.0.45678` (newer) == auto updates to ==> `Stable: 4.0.0` (older)

## To Reproduce
1. Install one of the nightly builds on the Releases page, (https://github.com/Microsoft/BotFramework-Emulator/releases/tag/v4.0.42244)
2. Click `Check for Updates` in the Help menu and install the update
3. Restart the Emulator
4. Go to Help -> About and check the version. It should be at the latest stable release (https://github.com/Microsoft/BotFramework-Emulator/releases/tag/v4.0.0-preview.40025)

## Expected behavior
There should be UI, probably in the App Settings page that allows the user to opt in or out of Auto Update (opt out would still be able to explicitly update). If the user opts in, there should also be an extra option to opt in or out of the pre-release updates.

## Screenshots
Here is a rough mock up of what it could look like:

![image](https://user-images.githubusercontent.com/3452012/47519731-14501080-d843-11e8-9fec-6877fbb79d36.png)

tonyanziano:
I just totally realized that I used the wrong template for this. 🤦‍♂️

PR #1110, opened 2018-11-19
Hooked up auto update download progress UI.

tonyanziano:
#1026 

=====

Forgot to hook up the progress dialog 😅


packages/app/main/src/main.ts
```diff
@@ -98,26 +101,35 @@ AppUpdater.on('update-available', async (update: UpdateInfo) => {
...
-      } catch (e) {
-        sendNotificationToClient(newNotification(e), mainWindow.commandService);
-      }
-    });
-    sendNotificationToClient(notification, mainWindow.commandService);
+    // TODO - localization
+    if (AppUpdater.userInitiated) {
+      // update the progress indicator
+      const { UpdateProgressIndicator } = SharedConstants.Commands.UI;
+      const progressPayload = { label: 'Download finished.', progress: 100 };
```
justinwilaby: The progress indicator goes from 0% to 100% but nothing in-between? It seems like we could show the actual progress of the download `bytesDownloaded / bytesTotal`

tonyanziano: I actually added the in between progress to the initial PR [(Code here)](https://github.com/Microsoft/BotFramework-Emulator/blob/77e65f6bd42e162a86831b37fcfb1633e0c1ca8d/packages/app/main/src/main.ts#L147) but I forgot to add the starting and end states, so the progress dialog never actually showed even though the state's progress indicator slice was being updated. 😅

justinwilaby: Nice!

Issue #1109, opened 2018-11-19
App Settings Editor UI incorrectly reflects internal state


tonyanziano:
## Version
4.1.0

## Describe the bug
The default app settings page state remains displayed by the UI even after the async call has been made to fetch the settings from `server.json` and a subsequent `setState()` is called with the updated settings.

## To Reproduce
1. In your `%emulator%/server.json` file, set the `usePrereleases` property to `true

![image](https://user-images.githubusercontent.com/3452012/48739120-cf33ba00-ec07-11e8-88c5-ff9ad0a88863.png)

2. Launch the emulator

3. Click on the App Settings gear icon in the bottom left
4. Observe how the checkbox for "Use pre-release versions" remains unchecked

![image](https://user-images.githubusercontent.com/3452012/48739192-1326bf00-ec08-11e8-9352-c69017f41d9e.png)

## Expected behavior
The component should initialize with default settings and then after retrieving the settings from disk, redraw with the updated settings.

## Additional context
This issue occurs regardless of the property. The default settings will always be persisted on the first draw. However, if you debug the render method, the state actually reflects the expected behavior.

[bug]

tonyanziano:
Turns out that it's caused by the way `<Checkbox/>` works

tonyanziano:
Addressed in #1115

PR #1115, opened 2018-11-20
Fixed incorrect UI state on App Settings page

tonyanziano:
Fix for #1109


packages/sdk/ui-react/src/widget/checkbox/checkbox.tsx
```diff
@@ -22,13 +54,24 @@ export class Checkbox extends Component<CheckboxProps, CheckboxState> {
 
   constructor(props: CheckboxProps) {
     super(props);
-    this.state = { checked: props.checked, focused: false };
+    const { checked } = props;
+    if (checked === undefined) {
+      console.error(
```
justinwilaby: We should treat undefined as false for the `checked` value, no?

packages/sdk/ui-react/src/widget/checkbox/checkbox.tsx
```diff
@@ -22,13 +54,24 @@ export class Checkbox extends Component<CheckboxProps, CheckboxState> {
...
+    this.state = { focused: false };
   }
 
   public render(): ReactNode {
     // Trim off what we don't want to send to the input tag
-    const { checked: _, className, label = '', ...inputProps } = this.props;
-    const { checked, focused } = this.state;
+    const { checked, className, label = '', ...inputProps } = this.props;
+    const { focused } = this.state;
+    if (checked === undefined) {
```
justinwilaby: Maybe we should `console.warn` on this condition so the checkbox does't mysteriously disappear given the case where the checked value in the constructor is 'false' and an update later on sets this to `undefined`

justinwilaby:
Let's give `getDerivedStateFromProps` a shot here. I think it will provide more flexibility when implementing and extending this component. Leave the `change` handler off like you have now.

```js
public static getDerivedStateFromProps(newProps: CheckboxProps, prevState: CheckboxState = {}): CheckboxState {

const { checked = false, indeterminate = false } = newProps;
const { checked: prevChecked, indeterminate: prevIndeterminate } = prevState;
if (  prevChecked !== checked || preIndeterminate !== indeterminate) {
  return {checked, indeterminate };
}
return prevState;
}
```

PR #1119, opened 2018-11-20
Committing updated qna.js file in debug extension.

tonyanziano:
Tired of ignoring this file.

All it is is the bundled react sample app under `extensions/debug/client`.

https://github.com/Microsoft/BotFramework-Emulator/blob/master/packages/extensions/debug/client/webpack.config.js#L4

I believe some of the dependencies got updated and generated a new bundled / minified version which resulted in a diff.


tonyanziano:
Even better, we could probably just gut the debug extension in the future once we have more time to look at it.

Issue #1167, opened 2018-12-13
"Migration complete!" popup strong html tags


artemious7:
## Version
v4.2.0

## Describe the bug
I have just upgraded from Emulator v3 to v4.2.0 and "Migration complete!" window pops up. It says: 

    ... saved them as <strong>.bot files</strong>...

The "strong" tag isn't handled properly and gets displayed instead of applying <strong>strong</strong> formatting.

## To Reproduce
1. Migrate from v3 to v4 (I didn't check if this happens after clean install)
2. Open Emulator
3. See the "Migration complete!" pops up

## Expected behavior
Proper formatting of the &lt;strong&gt; html tag

## Screenshots
![image](https://user-images.githubusercontent.com/16724889/49938231-f2d5e180-fee1-11e8-9472-3a8f316cfe01.png)

## Additional context
none

[bug]

tonyanziano:
Addressed in #1173

PR #1173, opened 2018-12-14
Fixed bold text in post migration dialog.

tonyanziano:
Fix for #1167


packages/app/client/src/ui/dialogs/postMigrationDialog/postMigrationDialog.tsx
```diff
@@ -47,8 +48,11 @@ export class PostMigrationDialog extends React.Component<PostMigrationDialogProp
     return (
       <Dialog cancel={ this.onClose } title="Migration complete!">
         <p>
-          { 'We’ve copied your bot endpoints from Emulator v3 and saved them as <strong>.bot files</strong>. ' +
-          'A <strong>.bot file</strong> stores metadata about different services your bot consumes and enables you ' +
+          { 'We’ve copied your bot endpoints from Emulator v3 and saved them as ' }
+          <strong className={ styles.boldText }>.bot files</strong> 
+          { '. A ' }
+          <strong className={ styles.boldText }>.bot file</strong>
+          { ' stores metadata about different services your bot consumes and enables you ' +
```
cwhitten: why the need to concatenate two different strings?

packages/app/client/src/ui/dialogs/postMigrationDialog/postMigrationDialog.tsx
```diff
@@ -47,8 +48,11 @@ export class PostMigrationDialog extends React.Component<PostMigrationDialogProp
     return (
       <Dialog cancel={ this.onClose } title="Migration complete!">
         <p>
-          { 'We’ve copied your bot endpoints from Emulator v3 and saved them as <strong>.bot files</strong>. ' +
-          'A <strong>.bot file</strong> stores metadata about different services your bot consumes and enables you ' +
+          { 'We’ve copied your bot endpoints from Emulator v3 and saved them as ' }
```
cwhitten: what is the i18n story with the Emulator? English only for now?

packages/app/client/src/ui/dialogs/host/host.tsx
```diff
@@ -34,25 +34,28 @@
...
     super(props);
   }
 
   public componentDidMount() {
     this._hostRef.addEventListener('dialogRendered', this.initFocusTrap);
   }
 
+  public componentWillUnmount() {
+    this._hostRef.removeEventListener('dialogRendered', this.initFocusTrap);
+  }
```
a-b-r-o-w-n: This appears to be unrelated to #1167. Can you provide some context about this change?

tonyanziano: Noticed that the component was never removing the event listener so I added it. I can split it into a different PR if necessary.

justinwilaby: @a-b-r-o-w-n - We often address tech debt in this way. The "see a weed, pull it now" strategy...

PR #1174, opened 2018-12-14
set node version to latest LTS

a-b-r-o-w-n:


.gitattributes
```diff
@@ -9,3 +9,4 @@
 *.json text eol=lf
 *.html text eol=lf
 *.map text eol=lf
+.node-version eol=lf
```
a-b-r-o-w-n: I'm not sure if this is necessary or not. I can remove it if it's not needed.

Issue #1157, opened 2018-12-07
Emulator - switching between chat tabs is really slow 


mewa1024:
## Version
nightly 4.1.45039 (but was also slow in previous versions)

## Describe the bug
Give a clear and concise description of what the bug is.

## To Reproduce
1. Open a bot.
2. Open a chat with an endpoint.
3. Open another chat with endpoint.
3. Click on the Welcome tab.
3. Click back on the endpoint tab.
4. It can take more than 1 second for Emulator to react.

## Expected behavior
The tab should load more quickly. If it is a loading problem, there should be some feedback to the user that tab is refreshing so they don't keep clicking on the tab.

## Additional context
I was using the Basic bot template which starts with an adaptive card.

[bug]

tonyanziano:
We know that this is related to the number of Webchat (livechat / transcript) tabs open. We have a hunch that it is due to excessive re-rendering of the Webchat component.

justinwilaby:
Another note, this seems to be exacerbated when Adaptive Cards are being rendered in WebChat.

PR #1177, opened 2018-12-17
Fixed tab switching performance issue

tonyanziano:
Fix for #1157 

===

Switching between tabs when a livechat session was open was having a major performance issue because of sending the entire client state over IPC multiple times.

This PR fixes that performance issue, rewrites `appMenuBuilder.ts` to allow us to change parts of the native app menu without reconstructing the entire thing, and adds some much needed tests.


packages/app/main/src/appMenuBuilder.ts
```diff
@@ -44,38 +43,124 @@ import { getActiveBot } from './botHelpers';
...
+      }
+    }
   }
 
-  /** Constructs a file menu template. If recentBots is passed in, will add recent bots list to menu */
-  async getFileMenu(recentBots?: BotInfo[]): Promise<MenuOpts> {
+  /** Creates a file menu item for each bot that will set the bot as active when clicked */
+  private static getRecentBotsList(bots: BotInfo[] = []): Electron.MenuItem[] {
+    // only list 9 most-recent bots
+    return bots.filter(bot => !!bot).slice(0, 9).map(bot => new Electron.MenuItem({
```
a-b-r-o-w-n: A slightly more idiomatic way to do this filter would be `bots.filter(Boolean)`

packages/app/main/src/main.ts
```diff
@@ -330,9 +330,7 @@ const createMainWindow = async () => {
   mainWindow.browserWindow.setTitle(app.getName());
   windowManager = new WindowManager();
 
-  AppMenuBuilder.getMenuTemplate().then((template: Electron.MenuItemConstructorOptions[]) => {
-    Menu.setApplicationMenu(Menu.buildFromTemplate(template));
-  });
+  AppMenuBuilder.initAppMenu().catch();
```
a-b-r-o-w-n: Is this just so that there are no uncaught errors?

tonyanziano:
Going to address PR comments and resubmit this against branch `v4.2.1

packages/app/main/src/appMenuBuilder.ts
tonyanziano: Awesome! 👍

packages/app/client/src/data/sagas/botSagas.ts
```diff
@@ -51,13 +56,14 @@ export function* generateHashForActiveBot(action: SetActiveBotAction): IterableI
 }
 
 export function* refreshConversationMenu(): IterableIterator<any> {
```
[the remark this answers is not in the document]
justinwilaby: `refreshConversationMenu()` can optionally be represented as a closure:
```js
yield takeEvery(
    [
      BotActions.setActive, 
      BotActions.load, 
      BotActions.close
    ],
    * () => yield CommandServiceImpl.remoteCall(SharedConstants.Commands.Electron.UpdateConversationMenu);
  );
```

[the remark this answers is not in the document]
tonyanziano: @justinwilaby I think I would take that approach if it was a one-liner, but now it's a 2 line function because of the `select` action. I'm leaning towards refactoring it into a `common.ts` or `shared.ts` inside of `/sagas/`

packages/app/main/src/main.ts
tonyanziano: Correct, but now I'm thinking it should be surfaced either through an Electron message box or at the very least through the console.

a-b-r-o-w-n: That sounds good.

PR #1178, opened 2018-12-17
Revert "Added missing https-proxy-agent dependency."

justinwilaby:
Reverts Microsoft/BotFramework-Emulator#1176 as this should have been merged into master


tonyanziano:
@justinwilaby I believe this should actually be merged into a 4.2.1 branch cut from 4.2. Then we can merge it into master from there. This should still be reverted from 4.2.

PR #1180, opened 2018-12-17
Fixed tab switching performance issue

tonyanziano:
Fix for #1157 

===

**Same PR as before with suggested changes integrated**

--- 

Switching between tabs when a livechat session was open was having a major performance issue because of sending the entire client state over IPC multiple times.

This PR fixes that performance issue, rewrites `appMenuBuilder.ts` to allow us to change parts of the native app menu without reconstructing the entire thing, and adds some much needed tests.

### DIFF
--- a/packages/app/client/src/data/sagas/botSagas.ts
+++ b/packages/app/client/src/data/sagas/botSagas.ts
@@ -36,12 +36,7 @@ import { CommandServiceImpl } from '../../platform/commands/commandServiceImpl';
 import { SharedConstants } from '@bfemulator/app-shared';
 import { BotActions, botHashGenerated, SetActiveBotAction } from '../action/botActions';
 import { generateBotHash } from '../botHelpers';
-import { RootState } from '../store';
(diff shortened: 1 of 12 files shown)


packages/app/client/src/data/sagas/botSagas.ts
```diff
@@ -31,11 +31,17 @@
...
 import { CommandServiceImpl } from '../../platform/commands/commandServiceImpl';
 import { SharedConstants } from '@bfemulator/app-shared';
 import { BotActions, botHashGenerated, SetActiveBotAction } from '../action/botActions';
 import { generateBotHash } from '../botHelpers';
+import { RootState } from '../store';
+import { refreshConversationMenu } from './sharedSagas';
+
+export function editorSelector(state: RootState) {
+  return state.editor;
+}
```
a-b-r-o-w-n: Looks like this can be imported from the shared sagas as well.

tonyanziano: Oh whoops. Good catch

PR #1190, opened 2018-12-19
Fixes #1169 - Region missing from LUIS service editor

### LINKED ISSUE #1169: LUIS Service created by Emulator is invalid
## Version
4.2.0

## Describe the bug
I used the Services area to add a new LUIS service to my `.bot` file. When prompted, here is what I see:
image

**notice the lack of an area in which to put `Region`**. This results in a `NullReferenceException` when the Luis Service is attempted to be configured as per [the LUIS Sample](https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/12.nlp-with-luis/BotServices.cs#L39).

justinwilaby:


packages/app/main/src/appMenuBuilder.spec.ts
```diff
@@ -307,9 +307,9 @@ describe('AppMenuBuilder', () => {
     await AppMenuBuilder.initAppMenu();
 
     // verify that each section of the menu is the expected length
-    expect(appMenuTemplate).toHaveLength(5); // file, edit, view, convo, help
```
tonyanziano: Oh, I see what's happening here. This is supposed to test the menu construction on Windows / Linux, but when I wrote the test I didn't think about the test being run on a Mac machine, which is why the indices are off. Since when I run the test on my machine, `process.platform` is set to `'win32`, however it's set to `darwin` on yours.

If you look at the test case below this one, you can see that I am testing the menu construction on Mac, and I do that by this line:

`Object.defineProperty(process, 'platform', { value: 'darwin' });`

Can you change these indices back to what they were and instead add the line 

`Object.defineProperty(process, 'platform', { value: 'win32' });`

to this test case and try running it again?

PR #1192, opened 2018-12-20
Jwilaby/#1185 app insights

### LINKED ISSUE #1185: Create and manage App Insights via Login UI (vs modal)
Retrieve the list of App Insights services from Azure then display them for selection.

justinwilaby:
resolves #1185

### DIFF
--- a/packages/app/client/src/ui/shell/explorer/servicesExplorer/connectedServicePicker/connectedServicePicker.tsx
+++ b/packages/app/client/src/ui/shell/explorer/servicesExplorer/connectedServicePicker/connectedServicePicker.tsx
@@ -371,7 +371,7 @@ export class ConnectedServicePicker extends Component<ConnectedServicesPickerPro
     return (
       <>
         <a href="https://aka.ms/bot-framework-emulator-create-appinsights" className={ styles.paddedLink }>
-          Create a new Azure storage account
+          Create a new Application Insights Component
         </a>
         <p>
           { ` Signed in as ${ this.props.authenticatedUser }.` }
--- a/packages/app/main/src/services/appInsightsApiService.spec.ts
+++ b/packages/app/main/src/services/appInsightsApiService.spec.ts
@@ -1,0 +1,160 @@
+import '../fetchProxy';
+import { AppInsightsApiService } from './appInsightsApiService';
+
+const mockArmToken = 'bm90aGluZw.eyJ1cG4iOiJnbGFzZ293QHNjb3RsYW5kLmNvbSJ9.7gjdshgfdsk98458205jfds9843fjds';
+let mockArgsPassedToFetch = [];
+let mockResponses;
+jest.mock('node-fetch', () => {
+  const fetch = (url, headers) => {
+    mockArgsPassedToFetch.push({ url, headers });
+    return {
+      ok: true,
+      json: async () => mockResponses.shift(),
+      text: async () => mockResponses.shift()
+    };
+  };
+  (fetch as any).Headers = class {
+  };
+  (fetch as any).Response = class {
+  };
+  return fetch;
+});
+
+const mockResponseTemplate = [
+  {
+    // Subscriptions
+    value: [
+      {
+        'id': '/subscriptions/1234',
+        'subscriptionId': '1234',
+        'tenantId': '1234',
+        'displayName': 'Pretty nice Service',
+        'state': 'Enabled',
+        'subscriptionPolicies': {
+          'locationPlacementId': 'Public_2014-09-01',
+          'quotaId': 'MSDN_2014-09-01',
+          'spendingLimit': 'On'
+        },
+        'authorizationSource': 'Legacy'
+      },
+      {
+        'id': '/subscriptions/1234',
+        'subscriptionId': '1234',
+        'tenantId': '43211',
+        'displayName': 'A useful service',
+        'state': 'Enabled',
+        'subscriptionPolicies': {
+          'locationPlacementId': 'Internal_2014-09-01',
+          'quotaId': 'Internal_2014-09-01',
+          'spendingLimit': 'Off'
+        },
+        'authorizationSource': 'RoleBased'
+      }
+    ]
+  },
+  // Components
+  {
+    value: [
+      {
+        'id': '/subscriptions/08a9411c/resourceGroups/myResourceGroup/' +
+          'providers/microsoft.insights/components/TestAppInsights',
+        'name': 'TestAppInsights',
+        'type': 'microsoft.insights/components',
+        'location': 'westus2',
+        'kind': 'Node.JS',
+        'etag': '0000',
+        'properties': {
+          'ApplicationId': 'TestAppInsights',
+          'AppId': '70f773ca',
+          'Application_Type': 'Node.JS',
+          'Flow_Type': 'Redfield',
+          'Request_Source': 'IbizaAIExtension',
+          'InstrumentationKey': '2e1f4ec2',
+          'Name': 'TestAppInsights',
+          'CreationDate': '2018-11-20T17:29:18.0789365+00:00',
+          'PackageId': null,
+          'TenantId': '08a9411c',
+        }
+      }
+    ]
+  },
+  {
+    value: [
+      {
+        'id': '/subscriptions/08a9411c/resourceGroups/myResourceGroup2/' +
+          'providers/microsoft.insights/components/TestAppInsights',
+        'name': 'TestAppInsights',
+        'type': 'microsoft.insights/components',
+        'location': 'westus2',
+        'kind': 'Node.JS',
+        'etag': '0000',
+        'properties': {
+          'ApplicationId': 'TestAppInsights',
+          'AppId': '70f773ca2',
+          'Application_Type': 'Node.JS',
+          'Flow_Type': 'Redfield',
+          'Request_Source': 'IbizaAIExtension',
+          'InstrumentationKey': '2e1f4ec22',
+          'Name': 'TestAppInsights2',
+          'CreationDate': '2018-11-20T17:29:18.0789365+00:00',
+          'PackageId': null,
+          'TenantId': '08a9411c2',
+        }
+      }
+    ]
+  },
+  // api-keys
+  {
+    value: [
+      { id: '/subscriptions/1234/resourcegroups/container/providers/microsoft.insights/components/com/apikeys/53434' }
+    ]
+  },
+  {
+    value: [
+      { id: '/subscriptions/123456/resourcegroups/container/providers/microsoft.insights/components/com/apikeys/4532' }
+    ]
+  }
+];
+
+describe('The AppInsightsApiService', () => {
+  beforeEach(() => {
+    mockResponses = JSON.parse(JSON.stringify(mockResponseTemplate));
+    mockArgsPassedToFetch.length = 0;
+  });
+
+  it('should deliver AppInsightsService objects when the happy path is followed', async () => {
+    const result = await getResult();
+    expect(result.services.length).toBe(2);
+    expect(result.code).toBe(0);
+  });
+
+  it('should return an empty payload with an error if no subscriptions are found', async () => {
+    mockResponses = [{ value: [] }];
+    const result = await getResult();
+    expect(result).toEqual({ services: [], code: 1 });
+  });
+
+  it('should return an empty payload with an error if no components are found', async () => {
+    mockResponses[1] = mockResponses[2] = { value: [] };
+    const result = await getResult();
+    expect(result).toEqual({ services: [], code: 1 });
+  });
+});
+
+async function getResult() {
+  const it = AppInsightsApiService.getAppInsightsServices(mockArmToken);
+  let result = undefined;
+  while (true) {
+    const next = it.next(result);
+    if (next.done) {
+      result = next.value;
+      break;
+    }
+    try {
+      result = await next.value;
+    } catch (e) {
+      break;
+    }
+  }
+  return result;
+}
--- a/packages/app/main/src/services/appInsightsApiService.ts
+++ b/packages/app/main/src/services/appInsightsApiService.ts
@@ -1,0 +1,109 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license.
+//
+// Microsoft Bot Framework: http://botframework.com
+//
+// Bot Framework Emulator Github:
+// https://github.com/Microsoft/BotFramwork-Emulator
+//
+// Copyright (c) Microsoft Corporation
+// All rights reserved.
+//
+// MIT License:
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED ""AS IS"", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+import { ServiceCodes } from '@bfemulator/app-shared';
+import { AppInsightsService } from 'botframework-config/lib/models';
+import { AzureManagementApiService, AzureResource, baseUrl, Provider, Subscription } from './azureManagementApiService';
+
+export class AppInsightsApiService {
+
+  public static* getAppInsightsServices(armToken: string): IterableIterator<any> {
+    const payload = { services: [], code: ServiceCodes.OK };
+
+    // 1. get a list of subscriptions for the user
+    yield { label: 'Retrieving subscriptions from Azure…', progress: 25 };
+    const subs: Subscription[] = yield AzureManagementApiService.getSubscriptions(armToken);
+    if (!subs) {
+      payload.code = ServiceCodes.AccountNotFound;
+      return payload;
+    }
+
+    // 2. Retrieve the app insights components
+    yield { label: 'Retrieving Application Insights Components from Azure…', progress: 50 };
+    const req = AzureManagementApiService.getRequestInit(armToken);
+    const requests = subs.map(sub => {
+      const url = `${ baseUrl + sub.id }/providers/${ Provider.ApplicationInsights }/components?api-version=2015-05-01`;
+      return fetch(url, req);
+    });
+    const appInsightsResponses: Response[] = yield Promise.all(requests);
+    const appInsightsComponents: { component: AzureResource, subscription: Subscription }[] = [];
+    let i = appInsightsResponses.length;
+    while (i--) {
+      const response = appInsightsResponses[i];
+      if (!response.ok) {
+        continue;
+      }
+      const { value: components = [] }: { value: AzureResource[] } = yield response.json();
+      const subscription = subs[i];
+      components.forEach(component => appInsightsComponents.push({ component, subscription }));
+    }
+
+    // 3. Retrieve the api-keys for each component
+    yield { label: 'Retrieving Api Keys from Azure…', progress: 75 };
+    const apiKeysRequests = appInsightsComponents.map(info => {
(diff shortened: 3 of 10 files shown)


packages/app/main/src/services/appInsightsApiService.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+    };
+  };
+  (fetch as any).Headers = class {
+  };
+  (fetch as any).Response = class {
+  };
+  return fetch;
+});
+
+const mockResponseTemplate = [
```
cwhitten: is there a notion of a `fixtures` directory in the emulator codebase? It would be nice to have this exportable and possibly held somewhere general for re-use, if it's reasonable to possibly need it again in a different test file.

packages/app/main/src/services/appInsightsApiService.ts
```diff
@@ -0,0 +1,114 @@
...
+    // 1. get a list of subscriptions for the user
+    yield { label: 'Retrieving subscriptions from Azure…', progress: 25 };
+    const subs: Subscription[] = yield AzureManagementApiService.getSubscriptions(armToken);
+    if (!subs) {
+      payload.code = ServiceCodes.AccountNotFound;
+      return payload;
+    }
+
+    // 2. Retrieve the app insights components
+    yield { label: 'Retrieving Application Insights Components from Azure…', progress: 50 };
```
cwhitten: are these yielded only for the tests?

packages/app/main/src/services/appInsightsApiService.ts
```diff
@@ -0,0 +1,114 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+import { ServiceCodes } from '@bfemulator/app-shared';
+import { AppInsightsService } from 'botframework-config/lib/models';
+import { AzureManagementApiService, AzureResource, baseUrl, Provider, Subscription } from './azureManagementApiService';
+
+export class AppInsightsApiService {
+
+  public static* getAppInsightsServices(armToken: string): IterableIterator<any> {
```
cwhitten: this function is doing quite a lot. can it be factored into smaller pieces? the way you've numbered the sections of code could be a good place to start!

packages/app/client/src/ui/shell/explorer/servicesExplorer/connectedServicePicker/connectedServicePicker.tsx
```diff
@@ -371,7 +371,7 @@ export class ConnectedServicePicker extends Component<ConnectedServicesPickerPro
     return (
       <>
         <a href="https://aka.ms/bot-framework-emulator-create-appinsights" className={ styles.paddedLink }>
-          Create a new Azure storage account
+          Create a new Application Insights Component
```
cwhitten: conventionally this should evaluate a JS string, no?

packages/app/main/src/services/appInsightsApiService.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+    return payload;
+  }
+}
+
+function createAppInsightsService(component: AzureResource, sub: Subscription, keys: string[]): AppInsightsService {
+  const { id, name, properties } = component;
+  const { InstrumentationKey, ApplicationId, TenantId } = properties;
+  const service = new AppInsightsService();
+  service.id = id;
```
cwhitten: instead of mutating `AppInsightsService`'s prototype after initialization here would it make more sense to pass this data into the constructor and giving it a type?

packages/app/main/src/services/cosmosDbApiService.ts
```diff
@@ -97,8 +97,9 @@ export class CosmosDbApiService {
 }
 
 function buildServiceModel
-(account: AzureResource, cosmosDb: AzureResource, collection: { id: string }): CosmosDbService {
+(account: AzureResource, cosmosDb: AzureResource, collection: { id: string, _rid: string }): CosmosDbService {
   const service = new CosmosDbService();
```
cwhitten: Can CosmosDbService also take constructor arguments instead of adding to members its prototype after initialization? see L101 - L109

packages/app/main/src/services/appInsightsApiService.ts
justinwilaby: Take a look at the [consumer of this generator](https://github.com/Microsoft/BotFramework-Emulator/blob/master/packages/app/main/src/commands/connectedServiceCommands.ts#L14). Those yields are to indicate progress.

packages/app/main/src/services/appInsightsApiService.spec.ts
justinwilaby: We do not have fixtures in this codebase since there isn't a lot of opportunity to reuse mock data other than in the test it applies to. In this case, these are specific to the test suite and cannot be reused (just the variable name is reused).

packages/app/main/src/services/appInsightsApiService.ts
justinwilaby: Nah, I thought about it but the function signatures would have a lot of repetitive args and would make debugging more difficult since I would have to yield to another generator. I'm not sure there is a lot to gain from breaking it out into smaller functions.

packages/app/client/src/ui/shell/explorer/servicesExplorer/connectedServicePicker/connectedServicePicker.tsx
justinwilaby: I'm not sure what you mean here. JSX treats text node children as strings so yes, in that sense it does evaluate to a JS string then used as a TextNode in the dom once rendered. Is this what you mean?

packages/app/main/src/services/appInsightsApiService.ts
justinwilaby: Another note is that we are in the process of evaluating how this data is retrieved. It could change on the next iteration to include a hierarchical data structure to show the relationship between Subscription, Resource Group and Finally the Resource itself.

packages/app/main/src/services/appInsightsApiService.ts
justinwilaby: Either way works. I find this to be cleaner and easier to read versus:
```js
new AppInsightsService( {
  id,
  name,
  serviceName: name,
  tenantId: TenantId,
  // ...
}}
```
But this is generally preference IMO.

packages/app/main/src/services/cosmosDbApiService.ts
justinwilaby: Out of curiosity, is there an advantage to doing so?

packages/app/main/src/services/appInsightsApiService.ts
justinwilaby: Made the changes here for consistency.

packages/app/main/src/services/appInsightsApiService.ts
cwhitten: I see. Is it common for a service like this to have such presentational opinions on the state it emits? I could see a better pattern something like emitting status codes that the called can respond to, having more control.

packages/app/client/src/ui/shell/explorer/servicesExplorer/connectedServicePicker/connectedServicePicker.tsx
cwhitten: By that I mean most of the text rendered in the dom is evaluated like this:

```javascript
{ 'Create a new Application Insights Component' }
```

cwhitten: By that I mean most of the text rendered in the DOM is evaluated like this:

{ 'Create a new Application Insights Component' }

cwhitten: By that I mean most of the text rendered in the DOM is evaluated like this:

```javascript
{ 'Create a new Application Insights Component' }
```

packages/app/main/src/services/appInsightsApiService.ts
cwhitten: Passing a shape in the constructor is a technique to document the contract to the caller much better than expecting the caller to just know what values it needs to append to the prototype to have the instance work correctly. It also makes tooling work better (autocomplete/documented function signatures, for example)

packages/app/main/src/services/cosmosDbApiService.ts
cwhitten: Passing a defined shape in the constructor of a class is a technique that gives you a few things:

1. It documents a contract to the caller much better than expecting the caller to just know what values it needs to append to the prototype to have the instance work correctly.
2. It makes tooling work better (autocomplete/documented function signatures, for example).
3. It allows one to specify default values or validate inputs before the caller does something bad trying to use it down the road.
4. ..there are more merits that I could expand on if you would like

It's a pretty effective mechanism.

packages/app/main/src/services/appInsightsApiService.ts
justinwilaby: So instead of emitting text, emit status codes that map to text in the consumer? Seems like an extra step that moves the responsibility of interpreting state onto the consumer. The idea here is to keep the consumer simple and dumb so it can manage any number of iterators. The service then is burdened with managing its own state as yielded output from the generator.

packages/app/main/src/services/appInsightsApiService.ts
cwhitten: Hmm, it smells to me. I would think that would be _more_ reason to be thoughtful about the separation of concerns, to not throw it all away if there is a change is relation or organization. Why can this be broken into the most reasonable and smallest unit of work, so that down the road it can very easily be recombined or factored without much lift?

Tangentially but I forgot to ask, what happens if any of the network activity in the parts that need it fails?

packages/app/client/src/ui/shell/explorer/servicesExplorer/connectedServicePicker/connectedServicePicker.tsx
justinwilaby: Bracket notation is only necessary when you need to preserve whitespace at the end of linebreaks in code. e.g.
```js
<p>Create a new Application <- This space here will be lost when transpiling
Insights Component</p>
```
To fix this we use:
```js
<p> { 'Create a new Application Insights '+ <- space is preserved
'Component' } </p>
```

cwhitten: Understood, a lot of codebases would enforce brackets always for consistency. Once we move to internationalization this won't matter, as every string will be require it.

packages/app/main/src/services/cosmosDbApiService.ts
justinwilaby: In this case, a source config object is optional in the constructor and no default values are specified. I can see something like this being helpful: 
```js
class CosmosDbService {
  constructor({endpoint = '', key='', database='', collection=''}) {
    super({endpoint, key, database, collection});
}
```
But the author of the schema choose not to mandate default values.

packages/app/main/src/services/appInsightsApiService.ts
cwhitten: A scenario off to top of my head would be if the view needed to special case the label for any reason based off of some state that the view has (or knowledge of the user, their locale/language settings, etc).

justinwilaby: Since these use cases do not yet exist and the technical implementation isn't clear. Let's keep this as-is and revisit it once we have a strategy for localization. I'm skeptical about adding infrastructure when the value of doing so depends on a yet-to-be defined requirement. Let's define the requirement, then refactor to fit.

packages/app/main/src/services/appInsightsApiService.ts
justinwilaby: Reusable work units are sufficiently broken out in the `AzureManagementApiService` class. Each of the composed services provide a concrete implementation that aggregates various data to build the models that are delivered once the generator completes. As we iterate on design and a new use case requires a different approach, refactoring is trivial and a low LOE.

packages/app/main/src/services/appInsightsApiService.spec.ts
cwhitten: FWIW I still think it's plenty valuable to remove this type of data from the test file for readability if nothing else. Plus it gives other developers a place to look to see if they're working on some shape that may be used already.

packages/app/main/src/services/appInsightsApiService.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (!appInsightsComponents.length) {
+      payload.code = ServiceCodes.AccountNotFound;
+      return payload;
+    }
+
+    // 3. Retrieve the api-keys for each component
+    yield { label: 'Retrieving Api Keys from Azure…', progress: 75 };
+    const apiKeysRequests = appInsightsComponents.map(info => {
+      const { component } = info;
+      return fetch(`${ baseUrl + component.id }/apiKeys?api-version=2015-05-01`, req);
```
cwhitten: `apiKeys?api-version=2015-05-01` should probably be added to https://github.com/Microsoft/BotFramework-Emulator/blob/35e5eecd8a786949f7e3d21c7966e81982e6b88b/packages/app/main/src/services/azureManagementApiService.ts#L80 and referenced here.

packages/app/main/src/services/appInsightsApiService.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (!subs) {
+      payload.code = ServiceCodes.AccountNotFound;
+      return payload;
+    }
+
+    // 2. Retrieve the app insights components
+    yield { label: 'Retrieving Application Insights Components from Azure…', progress: 50 };
+    const req = AzureManagementApiService.getRequestInit(armToken);
+    const requests = subs.map(sub => {
+      const url = `${ baseUrl + sub.id }/providers/${ Provider.ApplicationInsights }/components?api-version=2015-05-01`;
```
cwhitten: `${ baseUrl + sub.id }/providers/${ Provider.ApplicationInsights }/${AccountIdentifier.ApplicationInsights}` ?

packages/app/main/src/services/cosmosDbApiService.ts
```diff
@@ -86,28 +86,32 @@ export class CosmosDbApiService {
...
-  service.resourceGroup = account.id.split('/')[4];
-  service.subscriptionId = account.subscriptionId;
-  service.tenantId = account.tenantId;
-
-  return service;
+function buildServiceModel(account: AzureResource,
+                           cosmosDb: AzureResource,
+                           key: string,
+                           collection: { id: string, _rid: string }): CosmosDbService {
+  return new CosmosDbService({
```
cwhitten: With an ARM token from the Resource Manager, what's keeping us from utilizing the published SDK's?

https://www.npmjs.com/package/azure-arm-cosmosdb

packages/app/main/src/services/appInsightsApiService.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function createAppInsightsService(component: AzureResource, sub: Subscription, keys: string[]): AppInsightsService {
+  const { id, name, properties, subscriptionId } = component;
+  const {
+    InstrumentationKey: instrumentationKey,
+    ApplicationId: applicationId,
+    TenantId: tenantId
+  } = properties;
+  return new AppInsightsService({
```
cwhitten: With an ARM token from the Resource Manager, what's keeping us from utilizing the published SDK's?

https://www.npmjs.com/package/azure-arm-appinsights

packages/app/main/src/services/appInsightsApiService.ts
justinwilaby: The apiKeys endpoint is not an AccountIdentifier since it does not return account related information.

packages/app/main/src/services/cosmosDbApiService.ts
justinwilaby: The [dependency list](https://github.com/Azure/azure-sdk-for-node/blob/master/package.json#L30) is why. It also uses ms-rest which has an insane number of dependencies itself. Pulling in very large libraries to accomplish trivial tasks will only serve to slow down an already slow build/startup and to increase the app's overall size. We need to be **really** careful what we choose to include as a dependency and live by the rule: *The best dependency is no dependency.*

packages/app/main/src/services/appInsightsApiService.ts
justinwilaby: I have taken this into consideration as part of the initial technical study and have chosen not to include this as a dependency. See my comments on cosmosDB as to why.

packages/app/main/src/services/cosmosDbApiService.ts
cwhitten: You mean _this_ ms-rest? The one already in our depedency tree?
```
"ms-rest-js": {
  "version": "1.0.455",
  "resolved": "https://registry.npmjs.org/ms-rest-js/-/ms-rest-js-1.0.455.tgz",
  "integrity": "sha512-RUDnFFNhk4ZdvOACg0yfaxmp5OzNwUcTIwgh/rVBeuNzgA7hOoVh5zFW06XmOtaBHXL2Bu/vWoQtzloEUlv9tw==",
  "requires": {
    "axios": "^0.18.0",
    "form-data": "^2.3.2",
    "tough-cookie": "^2.4.3",
    "tslib": "^1.9.2",
    "uuid": "^3.2.1",
    "xml2js": "^0.4.19"
  }
}
```
was there any benchmarking done on this decision? can we not treeshake any of the deadcode?

cwhitten: I did a quick size test, here is what these two sdk's produce:
```
~/src/cosmosdbtest $ cat package.json
{
  "name": "cosmosdbtest",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "azure-arm-appinsights": "^2.1.0",
    "azure-arm-cosmosdb": "^2.3.0",
    "install": "^0.12.2",
    "npm": "^6.5.0"
  }
}
```
```
~/src/cosmosdbtest $ ls -lah
total 192
drwxr-xr-x   5 chriswhitten  staff   160B Dec 31 14:12 .
drwxr-xr-x  12 chriswhitten  staff   384B Dec 31 14:11 ..
drwxr-xr-x  74 chriswhitten  staff   2.3K Dec 31 14:12 node_modules
-rw-r--r--   1 chriswhitten  staff    88K Dec 31 14:12 package-lock.json
-rw-r--r--   1 chriswhitten  staff   354B Dec 31 14:12 package.json
```

packages/app/main/src/services/appInsightsApiService.ts
justinwilaby: Different API versions perform different actions on the same endpoint. It is possible that 2 different API versions can be used on the same endpoint yielding different results so I am keeping them within the concrete implementation instead of an enumeration.

packages/app/main/src/services/cosmosDbApiService.ts
justinwilaby: ms-rest is being removed when we can get away with it. We already made the decision to use an isomorphic `fetch` implementation. Haphazardly including dependencies means we could end up with 4 different libraries that make rest calls. 

Also, ms-rest- [does not appear to support proxy settings](https://github.com/Azure/ms-rest-js/issues/169) so using it will introduce a regression that will not exist with the one in this PR.  

Regarding the dependencies, remember that they are fractal and NPM downloads all child, grandchild, great gandchild, etc. dependencies. This is a list of dependencies from `node_modules` when the only entry in the `package.json` is `azure-arm-cosmosdb`: 

image

As you can see, this is probably not something we should include just to gain a few trivial rest calls.

packages/app/main/src/services/appInsightsApiService.spec.ts
justinwilaby: Ok, should I do this for all the others (~25+)so we can be consistent throughout the app?

packages/app/main/src/services/cosmosDbApiService.ts
cwhitten: A few things:

- I understand how npm resolves dependency trees
- 62 of the 80 dependencies listed exist in the `main` package already (77.5%)
- Adding `azure-arm-cosmosdb` and `azure-arm-appinsights` appear to add very little in total bytes to the dependency directory
- As I see it, bundle size is important to consider, but less important than traditional webapps. Nonetheless, my approach to considering additional dependencies is definitely not "haphazard".
- The "trivial" REST calls are not what you should focus on. You should focus on how you've now added a burden of maintaining live code and test code around communication with these services as they age and improve. There is a litany of things they could change requiring us to change our implementation(s) and test code down the road that could have been as easy as a version bump.
- As new developers come into this codebase, you've locked them in to looking at your implementation to understand how to use your interface with these services instead of referencing the evergreen documentation published by the services themselves on the web.

There are scenarios where it would be appropriate to roll our own REST client but I'm not seeing it in this case beyond the missing support in using proxy which I would like to understand more, but generally speaking the next time we have a decision like this to make lets consider what I mentioned above.

packages/app/main/src/services/appInsightsApiService.ts
cwhitten: `"components?api-version=2015-05-01"` is the value currently assigned to `AccountIdentifier.ApplicationInsights`. You'd prefer to duplicate and hardcode it here?

justinwilaby:
Stale

Issue #1037, opened 2018-10-25
Get rid of Gulp & Simplify v4 packaging process


tonyanziano:
We don't need to use gulp anymore, and we can move to pure npm / node scripts. We should also try to simplify the packaging config for all platforms like in v3.

compulim:
Do it! 💪

I think it's hard, but also think about using Webpack as little as possible. It is because Electron doesn't quite need bundling as web browser do, there are no network traffic. And it could make the app launch faster (less things to parse/interpret at boot time).

tonyanziano:
- [x] Windows
- [x] Linux
- [ ] Mac

tonyanziano:
Pushing Mac work to 4.3

tonyanziano:
Addressed in #1200

PR #1200, opened 2018-12-28
Removed gulp from mac build process & project

tonyanziano:
This completes the remaining work items for #1037


CONTRIBUTING.md
```diff
@@ -40,7 +40,7 @@ cd BotFramework-Emulator
 > npm version 5.6.0 or greater is required.
 
 ```
-npm i -g lerna@3.4.0 gulp@4.0.0 gulp-cli@2.0.1 webpack@4.8.x jest
+npm i -g lerna@3.4.0 webpack@4.8.x jest
```
a-b-r-o-w-n: Are these globals still necessary? Lerna is listed as a project dependency, I'm not sure about jest or webpack.

tonyanziano: Lerna is listed as a dependency in the root, but nowhere in our docs do we tell the user to run an `npm install` in the root to install lerna. We actually tell them explicitly not to run `npm install` in any of the directories. So we can either add the step and modify the docs a bit, or we can leave it as it is and remove it from the root dependencies.

I don't think `jest` is necessary, because it's only needed if you are running the unit tests.

Webpack, I'm unsure of whether we still need that as a global since we use babel in most of the packages now, and the packages that we do use webpack have it listed as a dependency at `4.16.x`

PR #1201, opened 2018-12-28
Adds support for closing a tab, creating a bot, and opening a bot via keyboard bindings

cwhitten:
- (CMD/CTRL) + W now closes the active tab if there is one, currently the bindings close the application entirely.
- (CMD/CTRL) + O now prompts the user to select a .bot file to open. currently there is no behavior with this binding.
- (CMD/CTRL) + N now opens the new bot configuration modal. currently there is no behavior with this binding.

Bugfixes/Improvements:

When a conversation tab is closed and opened again via an open dialog(s), the user receives an error about the bot already being open. To achieve this behavior successfully it requires a click on the correct Endpoint item to re-open the conversation tab. This change makes it so that the tab is automatically re-opened without the error when the user opens an already open bot with an open dialog.

### DIFF
--- a/packages/app/client/src/index.tsx
+++ b/packages/app/client/src/index.tsx
@@ -44,6 +44,7 @@ import { showWelcomePage } from './data/editorHelpers';
 import { CommandRegistry, registerAllCommands } from './commands';
 import { SharedConstants, newNotification } from '@bfemulator/app-shared';
 import { beginAdd } from './data/action/notificationActions';
+import { globalHandlers } from './utils/eventHandlers';
 import 'botframework-webchat/botchat.css';
 import './ui/styles/globals.scss';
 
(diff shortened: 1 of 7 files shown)


packages/app/client/src/index.tsx
```diff
@@ -67,13 +67,32 @@ CommandServiceImpl.remoteCall(SharedConstants.Commands.ClientInit.Loaded)
...
     const notification = newNotification(errMsg);
+    window.removeEventListener('keydown', globalKeyboardEventListener);
     store.dispatch(beginAdd(notification));
   });
 
 if (module.hasOwnProperty('hot')) {
   (module as any).hot.accept();
 }
+
+const globalKeyboardEventListener: EventListener = (event: KeyboardEvent): void => {
```
tonyanziano: Would it be better to extract this listener into its own file for the sake of future-proofing? If we continue to add more keybindings down the road this listener could get rather large.

packages/app/client/src/index.tsx
```diff
@@ -67,13 +67,32 @@ CommandServiceImpl.remoteCall(SharedConstants.Commands.ClientInit.Loaded)
...
+  // Meta corresponds to 'Command' on Mac
+  const ctrlOrCmdPressed = event.getModifierState('Control') || event.getModifierState('Meta');
```
tonyanziano: `event.getModifierState()` is deprecated [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState#Browser_compatibility), you can instead use:

`event.ctrlKey || event.metaKey`

packages/app/client/src/ui/shell/mdi/tabBar/tabBar.tsx
```diff
@@ -121,6 +129,17 @@ export class TabBar extends React.Component<TabBarProps, TabBarState> {
     );
   }
 
+  private keyboardListener = (event: KeyboardEvent): void => {
+    // Meta corresponds to 'Command' on Mac
+    const ctrlOrCmdPressed = event.getModifierState('Control') || event.getModifierState('Meta');
```
tonyanziano: same as above, can use `event.ctrlKey || event.metaKey`

packages/app/client/src/index.tsx
cwhitten: ah, nice. I just grabbed that from some other `keydown` event handlers.

packages/app/client/src/index.tsx
cwhitten: Yeah! Most of the existing event handlers are peppered throughout the components that care. I think consolidating them would be good, and this handler would be a good place to start it.

packages/app/client/src/index.tsx
```diff
@@ -67,13 +67,32 @@ CommandServiceImpl.remoteCall(SharedConstants.Commands.ClientInit.Loaded)
...
+  if (ctrlOrCmdPressed && key ===  'o') {
+    const { Commands: { Bot: { OpenBrowse }} } = SharedConstants;
+    CommandServiceImpl.call(OpenBrowse).catch();
+  }
+
+  if (ctrlOrCmdPressed && key ===  'n') {
+    const { Commands: { UI: { ShowBotCreationDialog }} } = SharedConstants;
+    CommandServiceImpl.call(ShowBotCreationDialog).catch();
+  }
+};
```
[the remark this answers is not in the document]
a-b-r-o-w-n: I have an idea of how we might build in the keybinding context as well. We can discuss offline.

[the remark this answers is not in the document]
cwhitten: a `keybindings.json` input from users along with a defined list of commands that they can reference, spread over some defaults would be nice functionality, but we should gauge interest a bit before implementing something like that.

cwhitten:
closing in lieu of https://github.com/Microsoft/BotFramework-Emulator/pull/1202

Issue #560, opened 2018-05-16
Menu items need shortcut keys


emadelwany:
Standard windows short cuts for working with menus is to use Alt + letter to access menu items
for example:
ALT-F should show &File menu
ALT-F O should invoke &File->&Open... 

Likewise, standard shortcuts are
CTRL+O to open a file
CTRL+S to save a file
etc.

justinwilaby:
fixed by #1203

PR #1203, opened 2018-12-29
Add some additional keybindings, improve re-opening an already open bot

cwhitten:
(CMD/CTRL) + W now closes the active tab if there is one, currently the binding closes the application entirely.
(CMD/CTRL) + O now prompts the user to select a .bot file to open. currently there is no behavior with this binding.
(CMD/CTRL) + N now opens the new bot configuration modal. currently there is no behavior with this binding.
Bugfixes/Improvements:

When a conversation tab is closed and opened again via an open dialog, the user receives an error about the bot already being open. This is nonsensical. To achieve this behavior successfully it requires a click on the correct Endpoint item to re-open the conversation tab. This change makes it so that the tab is automatically re-opened without the error when the user opens an already open bot with an open dialog.


packages/app/client/src/ui/shell/mdi/tabBar/tabBar.tsx
```diff
@@ -82,6 +82,14 @@ export class TabBar extends React.Component<TabBarProps, TabBarState> {
     };
   }
 
+  public componentDidMount() {
+    window.addEventListener('keydown', this.keyboardListener);
```
justinwilaby: Event naming should follow JS best practices. in this case `onKeyDown` or `onWindowKeydown` is recommended.

packages/app/client/src/utils/eventHandlers.ts
```diff
@@ -0,0 +1,51 @@
...
+import { CommandServiceImpl } from '../platform/commands/commandServiceImpl';
+import { SharedConstants } from '@bfemulator/app-shared';
+
+export const globalHandlers: EventListener = (event: KeyboardEvent): void => {
+  // Meta corresponds to 'Command' on Mac
+  const ctrlOrCmdPressed = event.ctrlKey || event.metaKey;
+  const key = event.key.toLowerCase();
+
+  if (ctrlOrCmdPressed && key ===  'o') {
+    const { Commands: { Bot: { OpenBrowse }} } = SharedConstants;
```
justinwilaby: This `const` declaration is redundant since it is only ever used once. Consider just using `SharedConstants.commands.Bot.OpenBrowse`

packages/app/client/src/utils/eventHandlers.ts
```diff
@@ -0,0 +1,51 @@
...
+    CommandServiceImpl.call(OpenBrowse).catch();
+  }
+
+  if (ctrlOrCmdPressed && key ===  'n') {
+    const { Commands: { UI: { ShowBotCreationDialog }} } = SharedConstants;
```
justinwilaby: same here: redundant `const` declaration

packages/app/client/src/utils/eventHandlers.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+jest.mock('../platform/commands/commandServiceImpl', () => ({
+  CommandServiceImpl: {
+    call: async (commandName: string, ...args: any[]) => {
+      mockLocalCommandsCalled.push({ commandName, args: args });
+    },
+  }
+}));
+
+describe('#globalHandlers', () => {
```
justinwilaby: Consider making this test suite read like a sentence. Instead of `#globalHandlers handles CMD+O`, use `The globalHandlers should call OpenBrowse when CMD+O is pressed`

justinwilaby:
@cwhitten - 

There are a few things about this pull request that gives me pause. I strongly believe this PR should not be accepted in its current state for the following reasons:

1. It changes the functionality of how a bot it opened  - this PR eliminates the message to the user that a bot is already open and to click on an endpoint to open a new conversation. Instead, the user is presented with a dialog to switch bots (even though the bot they are switching to is already open). This has the side effect of irreversibly wiping out all previous conversations from that bot. The current implementation is intentional and was put in place to avoid this. This PR removes that intentional design and introduces a defect.

2. The keyboard shortcuts in this PR deviate from the accepted and approved design listed in #560.

My suggestion is to revert the changes to how a bot is opened and implement the shortcuts as described in #560.

cwhitten:
Additionally, we should re-evaluate the accepted keybindings for Windows. I'd like to introduce the following that we can reference https://code.visualstudio.com/shortcuts/keyboard-shortcuts-windows.pdf

The intention of these few bindings in this change was considered really low-hanging and not meant to be comprehensive.

cwhitten:
I've updated the diff to include a candidate change to the `confirmAndSwitchBots` implementation in `activeBotHelper.ts`, essentially removing the confirmation step and just moving on with the "switch" portion of the implementation. There are two call sites for this function in the live code, both in `botCommands.ts`. Can someone explain to me why we can't do something like this this?

packages/app/client/src/ui/shell/mdi/tabBar/tabBar.spec.tsx
```diff
@@ -257,4 +257,20 @@ describe('TabBar', () => {
...
+
+  it('should respond trigger keyboard listeners', () => {
+    const map: any = {};
+    window.addEventListener = jest.fn((event, cb) => {
+      map[event] = cb;
+    });
+    const closeTabSpy = jest.fn();
+
+    mount(
+      <TabBar tabOrder={ [] } documents={ {} } activeDocumentId={ '1234' } closeTab={closeTabSpy}/>
```
tonyanziano: padding between curly braces:

`closeTab={ closeTabSpy }`

packages/app/client/src/ui/helpers/activeBotHelper.ts
```diff
@@ -228,70 +228,57 @@ export const ActiveBotHelper = new class {
...
-
     // TODO: We need to think about merging this with confirmAndCreateBot
     console.log(`Switching to bot ${botPath}`);
 
     try {
-      // prompt the user to confirm the switch
-      const result = await this.confirmSwitchBot();
-
-      if (result) {
-        store.dispatch(EditorActions.closeNonGlobalTabs());
```
tonyanziano: On second glance, I noticed that this `closeNonGlobalTabs()` is never being called in the updated version of `switchBots()`.

This means that if you previously had a livechat tab open with `BotFile1`, and were now switching to `BotFile2` , the livechat tab corresponding to `BotFile1` would not be closed, meaning that we would have a hanging livechat with a bot file that is not currently active. This doesn't really fit within the Emulator's concept of having one "active" bot file.

justinwilaby:
Looks like this PR causes a small defect. Unless this is WOD....

## repro

1. Open Bot #1 and start a conversation 
2. Open Bot #1 again

**Notice** 2 Chat windows are now open

## Expected behavior
The existing chat window should become focused and a second window should not appear.

cwhitten:
@justinwilaby I tend to agree with you. I saw https://github.com/Microsoft/BotFramework-Emulator/blob/85b220825625068b79245f7df2a6ed811ba7be10/packages/app/client/src/ui/helpers/activeBotHelper.ts#L270-L280 and opted not to go with what appeared to be a bit of a convention.

justinwilaby:
>The point of the change was not to do the ultimate and best experience possible, to be honest. It's improving a considerably poor existing experience. I'm happy to revisit down the road with some proper scope and prioritization, but something needs to land in the interim.

I can certainly [relate to that!](https://github.com/Microsoft/BotFramework-Emulator/pull/1192)

Issue #888, opened 2018-09-17
EMULATOR: Remove  "Toggle Developer Tools" items from the "Help" menu


mewa1024:
There are two items and the second one doesn't seem to do anything.

Screenshot from Mac build 4.0.316-alpha2, (but this also appears in Windows 4.0.474-nightly):
![image](https://user-images.githubusercontent.com/39317542/45645084-46b46400-ba74-11e8-8b47-f5d1c2576017.png)

justinwilaby:
The inspector dev tools toggle is used when an inspector is present in the extensions panel.

mewa1024:
@vishwacsena @justinwilaby Can we reopen this and move to backlog? (1) It's not obvious what it does, so it seems like an advanced feature, and (2) having to navigate back to the Help menu isn't the greatest UI for a toggle. If it is a useful feature that is used frequently, a keyboard shortcut would be better.

justinwilaby:
@mewa1024 - debugging the emulator internals is probably not something that is necessary often. When it is necessary, there are other ways to connect a debugger which does't involve explicit menu options. I vote to remove both menu items entirely.

mewa1024:
@vishwacsena v4.2?

vishwacsena:
The fix here is to remove the dev tools menu.

corinagum:
Resolved with PR #1081

PR #1218, opened 2019-01-08
allow for toggling devtools in development env

a-b-r-o-w-n:


justinwilaby:
Since the emulator uses an old version of Chrome with a less useful dev tools, we removed this 2 months ago in favor of launching with the flag `--remote-debugging-port=<port>` and using `chrome://inspect` instead. This also allows us to see and debug all instances of `<WebView>`

a-b-r-o-w-n:
@justinwilaby That is still an option in this change. However, using the remote debugger, I was unable to use the dom inspection feature, which is really useful when dealing with the webchat component.

justinwilaby:
#888 explicitly removed this menu option as a design decision. Putting it back in should probably go through a design review/approval first.

a-b-r-o-w-n:
Ok. I'll close for now, but I'd like to revisit at some point. I think having the keybinding open up chrome's devtools would still be nice.

PR #1219, opened 2019-01-09
upgrade webchat to v4

a-b-r-o-w-n:
Todo:

- [x] Speech - getting a 401 after sending a stt message from the bing api. working with Andrea to update the token the emulator uses to authenticate to bing.
- [x] Themes -- will submit another PR with adaptive cards theme fixes
- [x] Clean up any v3 related code / styles
- [x] a11y -- parity with released version, can discuss additions to UX
- [x] Specs


packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
```diff
@@ -65,24 +68,25 @@ export class Log extends React.Component<LogProps, LogState> {
...
+              ) {
+                return {
+                  selectedActivity: activity,
+                  currentlyInspectedActivity: showInInspector ? activity : null,
+                };
               }
-            }
+
+              return undefined;
+            });
```
a-b-r-o-w-n: The reason for this change is because we now have full control over updating the observable. Therefore we can standardize on the expected value and no longer need to branch when the observable has been updated from webchat or the log.

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+class ActivityWrapper extends Component<ActivityWrapperProps> {
+  setSelectedActivity = (activity) => () => {
+    const { onClick } = this.props;
+
+    onClick(activity);
+  }
+
+  handleKeyDown = (activity) => (e: React.KeyboardEvent) => {
```
justinwilaby: The handler name should follow JS conventions. In this case, `onKeyDown` or `onWebChatKeyDown` is recommended and more consistent with the handler names elsewhere in the app.

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const { onClick } = this.props;
```
justinwilaby: `onClick` is redundant, consider just using `this.props.onClick(activity)`

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
```diff
@@ -77,6 +83,37 @@
   }
 }
 
+.chatActivity {
```
justinwilaby: Naming should follow CSS conventions. In this case, `chat-activity` is recommended.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
```diff
@@ -77,6 +83,37 @@
...
+  cursor: pointer;
+  border-top: 1px solid transparent;
+  border-bottom: 1px solid transparent;
+  transition: all .1s linear;
```
justinwilaby: Using 'all' for transitions is not recommended since it may contribute to unwanted side-effects in later iterations. Please consider targeting a specific style.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
```diff
@@ -77,6 +83,37 @@
...
+    .bubble {
+      p {
+        font-family: var(--default-font-family);
+      }
+    }
+  }
+}
+
+
+.selectedActivity {
```
justinwilaby: Naming: `selected-activity`

packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
```diff
@@ -49,41 +49,38 @@ export interface ChatProps {
   onStartConversation: any;
   currentUserId: string;
   locale: string;
+  selectedActivity: any | null;
+  updateSelectedActivity: (activity: any) => void;
```
justinwilaby: `updateSelectedActivity: (activity: Activity) => void;` could work here.

justinwilaby:
Unit test is failing. `FAIL packages/app/client/src/ui/editor/emulator/parts/chat/chat.spec.ts`

packages/app/client/src/ui/styles/themes/dark.css
```diff
@@ -22,6 +22,9 @@ html {
   --error-outline: #BE1100;
 
   /* webchat overridden colors */
+  /* TODO: update for dark theme */
+  --webchat-selected-activity-bg: #F6F5D2;
```
justinwilaby: This variable is omitted in `high-contrast.css`

packages/app/client/src/ui/styles/themes/dark.css
```diff
@@ -22,6 +22,9 @@ html {
...
+  --webchat-selected-actvity-border: #00BCF2;
```
justinwilaby: This variable is omitted in `high-contrast.css`

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
justinwilaby: Also, handlers should be `private`

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
```diff
@@ -31,28 +31,77 @@
...
+      document.selectedActivity$.next({
+        ...activity,
+        showInInspector: true
+      });
+    }
+  }
 
   render() {
-    const { endpointUrl } = this.props.document || { endpointUrl: '' };
+    const { endpointUrl } = this.props.document;
```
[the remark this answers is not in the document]
justinwilaby: `const { endpointUrl } = this.props.document || { endpointUrl: '' };` provides sufficient guarding agains `null` or `undefined` document. Consider putting this back in.

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
```diff
@@ -31,28 +31,77 @@
...
+    selectedActivity$?: BehaviorSubject<any>;
+  };
   mode?: EmulatorMode;
   onStartConversation?: () => any;
   className?: string;
 }
 
-export default class ChatPanel extends React.Component<ChatPanelProps, {}> {
+interface ChatPanelState {
+  selectedActivity: any | null;
```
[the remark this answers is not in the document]
a-b-r-o-w-n: I didn't know about the activity type. I'll add that. My original plan for including `| null` was to improve the type defs in the future.

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
```diff
@@ -31,28 +31,77 @@
...
+    selectedActivity: null
+  };
+
+  selectedActivitySub: Subscription;
+
+  componentDidUpdate() {
+    const { document } = this.props;
+
+    if (document && document.selectedActivity$ && !this.selectedActivitySub) {
+      this.selectedActivitySub = document.selectedActivity$.subscribe(obj => {
```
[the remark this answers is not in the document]
a-b-r-o-w-n: :+1:

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
a-b-r-o-w-n: I would suggest that camel casing class names is preferred over css naming conventions when using css in js. The imported module uses camel case names regardless. Doing likewise in the css module follows the principle of least surprise.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
a-b-r-o-w-n: will do.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
justinwilaby: The history is that we began with camelCase css names then moved to kabab-case after a discussion about letting implementation details force use to a move away from idioms and best practices. There was a sweep application wide to change all css names from camelCase to kabab-case which is where we're at now. 

I feel it's important to be binary on these issues. We either use camelCase or kabab-case, not a mix. If there is a decision to move back to camelCase, we should make this change across all scss files in the app and place it in a separate PR.

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
```diff
@@ -54,7 +55,9 @@ export class Log extends React.Component<LogProps, LogState> {
   constructor(props: LogProps, context: LogState) {
     super(props, context);
     this.state = {
-      count: 0
+      count: 0,
```
compulim: - Sort

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
```diff
@@ -65,24 +68,25 @@ export class Log extends React.Component<LogProps, LogState> {
...
-              const { clicked } = fromLog;
-              if (clicked) {
-                this.currentlyInspectedActivity = activity;
+            // this activity came from the log (activities from the log are raw)
+            // ex: { id: , from: , to: , ... }
+            const activity = obj;
+            const { showInInspector } = activity;
+            const currentlyInspectedActivity = showInInspector ? activity : null;
+
+            this.setState((prevState) => {
```
compulim: -  No parenthesis for single argument lambda function

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
```diff
@@ -65,24 +68,25 @@ export class Log extends React.Component<LogProps, LogState> {
...
+              if (
+                prevState.selectedActivity !== activity ||
```
compulim: - `||` should drop to next line (prefixed, instead of suffixed)

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
```diff
@@ -65,24 +68,25 @@ export class Log extends React.Component<LogProps, LogState> {
...
+                prevState.currentlyInspectedActivity !== currentlyInspectedActivity
+              ) {
+                return {
+                  selectedActivity: activity,
```
compulim: - Sort
- Remove extra comma
- We could shave some bytes if we name `activity` as `selectedActivity` few lines up there

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
```diff
@@ -65,24 +68,25 @@ export class Log extends React.Component<LogProps, LogState> {
...
+                  currentlyInspectedActivity: showInInspector ? activity : null,
```
compulim: - `currentlyXXX` seems not common in naming
- We just computed the result of `currentlyInspectedActivity` outside of `setState`, do we want to compute here again?
- I didn't follow the logic, but could this be a `true`/`false` instead?
   - Looks like it can be derived from the `selectedActivity` too, if yes, no need another state

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
```diff
@@ -108,8 +112,8 @@ export class Log extends React.Component<LogProps, LogState> {
         {
           this.props.document.log.entries.map(entry =>
             <LogEntry key={ `entry-${key++}` } entry={ entry } document={ this.props.document }
-              selectedActivity={ this.selectedActivity }
-              currentlyInspectedActivity={ this.currentlyInspectedActivity } />
+              selectedActivity={ this.state.selectedActivity }
```
compulim: - Sort (not your fault, but I will try to prettify/upgrade the code every time I touch the line)

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
```diff
@@ -31,28 +31,73 @@
...
 import * as styles from './chatPanel.scss';
 
 import { ChatContainer } from '../parts/chat/chatContainer';
 import { EmulatorMode } from '../emulator';
 
 interface ChatPanelProps {
-  document: any;
+  document?: {
+    endpointUrl: string;
+    selectedActivity$?: BehaviorSubject<any>;
```
compulim: Web Chat may update the activity (into another immutable object, i.e. pointer will change), few reasons:

- We have a plan to support `updateActivity` activity
- After Web Chat send out a message (from = user), Direct Line will echo it back with additional information (ACK), we will merge/replace it with the copy inside Web Chat
- Web Chat may alter something inside the activity from time to time, e.g. mark the activity as spoken

So, I am unsure if it is reliable to hook into the activity object, or just the activity ID.

For activity ID, there is another story too:

- Activity ID is generated by Direct Line platform
- When the user send out the activity, we put a pseudo ID inside `channelData`, so we can track it for ACK
- When Direct Line platform echo/ack the message back, we will merge/replace the activity, so we got the activity ID

And instead of `any`, can we type it here?

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
```diff
@@ -31,28 +31,73 @@
...
   render() {
     const { endpointUrl } = this.props.document || { endpointUrl: '' };
 
     return (
       <div className={ `${styles.chatPanel} ${this.props.className || ''}` }>
         <header>{ endpointUrl }</header>
-        <ChatContainer mode={ this.props.mode } document={ this.props.document }
-                       onStartConversation={ this.props.onStartConversation }/>
+        <ChatContainer
+          mode={ this.props.mode }
```
compulim: - Sort

packages/app/client/src/ui/editor/emulator/emulator.tsx
```diff
@@ -150,8 +150,9 @@ class EmulatorComponent extends React.Component<EmulatorProps, {}> {
     }
     const selectedActivity$ = new BehaviorSubject({});
     const subscription = selectedActivity$.subscribe((obj: any) => {
```
compulim: - `any` or nullable `Activity`

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
```diff
@@ -77,6 +83,37 @@
...
+.chatActivity {
+  cursor: pointer;
+  border-top: 1px solid transparent;
+  border-bottom: 1px solid transparent;
+  transition: border .1s linear, color .1s linear, background .1s linear;
+
+  &:focus {
+    border-top: 1px solid var(--webchat-selected-actvity-border);
+    border-bottom: 1px solid var(--webchat-selected-actvity-border);
+    outline: none;
```
compulim: Note: touching `outline` might affect accessibility.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
```diff
@@ -32,15 +32,16 @@
 //
 import * as React from 'react';
 import { Component } from 'react';
+import { Activity } from '@bfemulator/sdk-shared';
```
compulim: - Should we sort `import` in Emulator?

packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
```diff
@@ -49,41 +50,38 @@ export interface ChatProps {
...
+    //       fetchCallback: getSpeechToken.bind(null, endpoint, false),
+    //       fetchOnExpiryCallback: getSpeechToken.bind(null, endpoint, true)
+    //     }),
+    //     speechSynthesizer: new Speech.BrowserSpeechSynthesizer()
+    //   } : null,
   };
 }
 
+function isCardSelected(selectedActivity: Activity | null, activity: Activity) {
+  return selectedActivity && activity.id && selectedActivity.id === activity.id;
```
compulim: Sometimes, `Activity` may not have an assigned `id` yet.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
```diff
@@ -116,23 +112,39 @@ export class Chat extends Component<ChatProps> {
...
+      };
+
       return (
         <div id="webchat-container" className={ `${styles.chat} wc-app wc-wide` }>
-          <WebChat
+          <ReactWebChat
             locale={ this.props.locale }
             key={ document.directLine.token }
+            styleOptions={ styleOptions }
+            activityMiddleware={ () => next => card => children => (
```
compulim: - Sort
- Try to put functions out of `render()` because it hurt performance

packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
```diff
@@ -116,23 +112,39 @@ export class Chat extends Component<ChatProps> {
...
+              <ActivityWrapper
+                activity={card.activity}
```
compulim: - Space surrounding bracelets

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
```diff
@@ -65,24 +68,25 @@ export class Log extends React.Component<LogProps, LogState> {
...
-              const activity = obj;
-              this.selectedActivity = activity;
-              const { fromLog = {} as ActivitySelectionFromLog }: { fromLog: ActivitySelectionFromLog } = activity;
-              // check if it was clicked or hovered
-              const { clicked } = fromLog;
-              if (clicked) {
-                this.currentlyInspectedActivity = activity;
+            // this activity came from the log (activities from the log are raw)
+            // ex: { id: , from: , to: , ... }
+            const activity = obj;
```
compulim: Do we need this one for rename? (I didn't look at full code)

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
```diff
@@ -65,24 +68,25 @@ export class Log extends React.Component<LogProps, LogState> {
...
+                prevState.currentlyInspectedActivity !== currentlyInspectedActivity
+              ) {
+                return {
+                  selectedActivity: activity,
+                  currentlyInspectedActivity: showInInspector ? activity : null,
+                };
               }
-            }
+
+              return undefined;
```
compulim: - Unneeded (no `return` always means `return undefined`)

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
```diff
@@ -108,8 +112,8 @@ export class Log extends React.Component<LogProps, LogState> {
         {
           this.props.document.log.entries.map(entry =>
             <LogEntry key={ `entry-${key++}` } entry={ entry } document={ this.props.document }
-              selectedActivity={ this.selectedActivity }
-              currentlyInspectedActivity={ this.currentlyInspectedActivity } />
+              selectedActivity={ this.state.selectedActivity }
+              currentlyInspectedActivity={ this.state.currentlyInspectedActivity } />
```
compulim: - Prefer to leave `/>` on the next line for better open/close indentation

packages/app/client/src/ui/editor/emulator/parts/log/logEntry.tsx
```diff
@@ -76,16 +75,14 @@ export class LogEntry extends React.Component<LogEntryProps> {
   inspectAndHighlightInWebchat(obj: any) {
     this.inspect(obj);
     if (obj.id) {
-      const fromLog: ActivitySelectionFromLog = { clicked: true };
-      this.props.document.selectedActivity$.next({ ...obj, fromLog });
+      this.props.document.selectedActivity$.next({ ...obj, showInInspector: true });
```
compulim: If you are patching the activity, use `channelData`, because `Activity` should be strictly-typed.

packages/app/client/src/ui/editor/emulator/parts/log/logEntry.tsx
```diff
@@ -64,8 +64,7 @@ export class LogEntry extends React.Component<LogEntryProps> {
    * @param obj Can be a conversation activity or network request
    */
   inspect(obj: {}) {
-    const fromLog: ActivitySelectionFromLog = { clicked: true };
-    this.props.document.selectedActivity$.next({ fromLog });
+    this.props.document.selectedActivity$.next({ showInInspector: true });
```
compulim: The name `selectedActivity$` suggest it is an activity.

If `selectedActivity$` is keeping information about two things: `Activity` and `showInInspector`. It should be named with inclusion of the "show in inspector" information.

packages/app/client/src/ui/editor/emulator/parts/log/logEntry.tsx
```diff
@@ -76,16 +75,14 @@ export class LogEntry extends React.Component<LogEntryProps> {
...
+      this.props.document.selectedActivity$.next({ ...obj, showInInspector: true });
     }
   }
 
   /** Highlights an activity in webchat (triggered by hover in log) */
   highlightInWebchat(obj: any) {
     if (obj.id) {
-      const fromLog: ActivitySelectionFromLog = { clicked: false };
-      this.props.document.selectedActivity$.next({ ...obj, fromLog });
+      this.props.document.selectedActivity$.next({ ...obj, showInInspector: false });
```
compulim: In line 96, you are using multiple lines. Let's converge on the code style, at least, for `selectedActivity$.next` call with `{ ...obj, showInInspector: XXX }`.

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
```diff
@@ -31,28 +31,73 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 import * as React from 'react';
+import { Subscription, BehaviorSubject } from 'rxjs';
```
compulim: - Sort

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
```diff
@@ -31,28 +31,73 @@
...
 import * as styles from './chatPanel.scss';
 
 import { ChatContainer } from '../parts/chat/chatContainer';
 import { EmulatorMode } from '../emulator';
 
 interface ChatPanelProps {
-  document: any;
+  document?: {
```
compulim: Note: I am not a fan of passing complex and/or stateful props into React component. In this case, `selectedActivity$` might change its internal state from time to time. It could easily introduce perf problem and debuggability issue. But I will leave it to you, may be it is unavoidable, or may be we can put it inside Redux.

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
```diff
@@ -31,28 +31,73 @@
...
+  state = {
+    selectedActivity: null
+  };
+
+  selectedActivitySub: Subscription;
+
+  componentDidUpdate() {
+    const { document } = this.props;
+
+    if (document && document.selectedActivity$ && !this.selectedActivitySub) {
```
compulim: Is it possible to have `selectedActivity$` of null given you protected it thru typing? If yes, is there any reasons why it can be nulled?

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
```diff
@@ -31,28 +31,73 @@
...
+      this.selectedActivitySub = document.selectedActivity$.subscribe(
+        selectedActivity => this.setState({ selectedActivity })
+      );
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.selectedActivitySub) {
```
compulim: Is there any reason the `selectedActivitySub` can be nulled?

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
```diff
@@ -0,0 +1,52 @@
+import * as React from 'react';
+import { Component, ReactNode } from 'react';
+import { Activity } from '@bfemulator/sdk-shared';
+
+import * as styles from './chat.scss';
+
+interface ActivityWrapperProps {
+  children: ReactNode;
```
compulim: - Sort

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
```diff
@@ -0,0 +1,52 @@
...
+  onClick: (activity: Activity) => void;
+}
+
+class ActivityWrapper extends Component<ActivityWrapperProps> {
+  render() {
+    const { activity, children, isSelected } = this.props;
+    let classes = styles.chatActivity;
+
+    if (isSelected) {
+      classes = `${classes} ${styles.selectedActivity}`;
```
compulim: - Space surrounding bracelets

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    // TODO: aria-label?
+    return (
+      <div
+      className={ classes }
+      onClick={this.setSelectedActivity(activity)}
+      tabIndex={0}
```
compulim: For accessibility, this is considered nested button although it is not a `<button>`. We should avoid nested hit area.

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
```diff
@@ -0,0 +1,52 @@
...
+      onKeyDown={this.onKeyDown(activity)}
+      >
```
compulim: Please add `aria-*` for accessibility, for example, telling the user whether this activity is selected/highlighted or not.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
```diff
@@ -1,12 +1,18 @@
 .chat {
   background-color: white;
   height: 100%;
+  width: 100%;
```
compulim: - Sort

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
```diff
@@ -77,6 +83,37 @@
   }
 }
 
+.chatActivity {
+  cursor: pointer;
+  border-top: 1px solid transparent;
+  border-bottom: 1px solid transparent;
+  transition: border .1s linear, color .1s linear, background .1s linear;
```
compulim: Do we have a variable for the speed of animation?

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
```diff
@@ -77,6 +83,37 @@
...
+      p {
+        font-family: var(--default-font-family);
+      }
+    }
+  }
+}
+
+
+.selectedActivity {
+  :global {
```
compulim: - Collapse selector

packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
```diff
@@ -116,23 +112,39 @@ export class Chat extends Component<ChatProps> {
...
         document.directLine,
-        document.selectedActivity$,
-        endpoint,
-        this.props.mode
+        endpoint
       );
 
+      // TODO: update for themes
+      const styleOptions = {
+        bubbleBackground: '#dcdcdc',
```
compulim: - Sort

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
a-b-r-o-w-n: Thanks for the info. I changed it to be kebab cased.

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
```diff
@@ -0,0 +1,52 @@
...
+    let classes = styles.chatActivity;
+
+    if (isSelected) {
+      classes = `${classes} ${styles.selectedActivity}`;
+    }
+
+    // TODO: aria-label?
+    return (
+      <div
+      className={ classes }
```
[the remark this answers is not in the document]
a-b-r-o-w-n: we need prettier! 💅

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
```diff
@@ -0,0 +1,52 @@
...
+      </div>
+    );
+  }
+
+  private setSelectedActivity = (activity: Activity) => () => {
+    this.props.onClick(activity);
+  }
+
+  private onKeyDown = (activity: Activity) => (e: React.KeyboardEvent) => {
+    switch (e.key) {
```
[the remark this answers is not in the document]
a-b-r-o-w-n: :+1:

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
a-b-r-o-w-n: What do you mean by this?

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
a-b-r-o-w-n: just `aria-label` and `aria-role='button'` is what I'm thinking. Do you have any thoughts about what the copy should be?

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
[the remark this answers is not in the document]
a-b-r-o-w-n: @compulim are you suggesting to sort alphabetically?

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
compulim: Combine `.selectedActivity` and `:global` into a single line.

a-b-r-o-w-n: gotcha. will do.

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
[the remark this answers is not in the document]
compulim: Yes, always sort alphabetically for easier Git merging and cosmetically.

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
compulim: Need to check applicability, few things top of my head here, but better to check the list.
- [`aria-multiselectable`](https://www.w3.org/TR/wai-aria-1.1/#aria-multiselectable)
- [`aria-selected`](https://www.w3.org/TR/wai-aria-1.1/#aria-selected)

Read https://www.w3.org/TR/wai-aria-1.1/ for the list.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.scss
a-b-r-o-w-n: I'm not able to find one.

compulim: If animation is new, please add a CSS variable. We might need to disable animation for high-contrast users.

a-b-r-o-w-n: Should I add this in the globals css file?

a-b-r-o-w-n: Looking at other transitions, they are not using a variable. So I'm going to pass on this work for now.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
compulim: `Activity` doesn't have `id` doesn't mean it is not uniquely identifiable, but it need to go thru another mean (`channelData`). This is due to limitation of Direct Line platform.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
compulim: Yea, until then, we will sort things by hand (or by some handy VSCode extensions). I am not oppose to tooling, I just sort things when I write.

Don't know the rules in Emulator. In Web Chat, we sort them into 2 groups. One is "from `node_modules`", the other is local. And separate them by a new line for easier readability. Then use a simple line-sorter extension.

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
a-b-r-o-w-n: I don't think it can be nulled, but it is initialized lazily in `emulator.tsx`.

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
a-b-r-o-w-n: It will not be nulled after being set, but it may never be set.

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
a-b-r-o-w-n: Meaning when a response from the bot includes interactive content?

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
a-b-r-o-w-n: I'll cover a bette a11y in a later commit. Let's revisit the holistic a11y story at that point.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
a-b-r-o-w-n: Deferring this until we add automated tooling. Right now they are sorted by global modules and local modules.

packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
a-b-r-o-w-n: In general, inlining functions in jsx do not impact performance to the degree that developers should always avoid it. There are some cases where it makes sense to make this optimization before profiling, but I don't think this is one. 

That being said, I'll go ahead and move it, but let's try to be cautious about making premature optimizations without first being able to provide profiling data.

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
a-b-r-o-w-n: I don't think we do, but I didn't change because it was already in place. I'll update.

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
a-b-r-o-w-n: Good points. Mostly just adapting the existing logic without _too_ much change. I'll think about this and maybe update.

packages/app/client/src/ui/editor/emulator/parts/log/log.tsx
a-b-r-o-w-n: Typescript wants a function to either always return a value or not. It even fails to compile when just `return` is used.

packages/app/client/src/ui/editor/emulator/parts/log/logEntry.tsx
a-b-r-o-w-n: Again, this is something that existed prior to my change (`const fromLog: ActivitySelectionFromLog = { clicked: true };`) so I left as is. I can go back after getting a review on the core functionality to see if this is necessary or not.

packages/app/client/src/ui/editor/emulator/parts/log/logEntry.tsx
a-b-r-o-w-n: The type for activity is definitely not _strict_ in the sense that **every** key is optional. Also, this follows the previous pattern in the code. We could talk about an abstraction that is used in the emulator that attaches the inspector info, but that should be a conversation to be had offline.

packages/app/client/src/ui/editor/emulator/parts/log/logEntry.tsx
a-b-r-o-w-n: This is purely a line length thing. I'm going to keep as is for now to reduce churn because prettier takes care of this for us.

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
a-b-r-o-w-n: addressed in 9b4275c9

packages/app/client/src/ui/editor/emulator/parts/chat/activityWrapper.tsx
a-b-r-o-w-n: The current a11y experience is consistent with the published version, so I'm calling this done. We can definitely improve the UX so let's discuss offline and get work prioritized.

a-b-r-o-w-n:
With regards to the adaptive cards styling, I am going to work on a follow-up PR that addresses that.

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+      const component = render();
+
+      expect(component.state('selectedActivity')).toEqual(null);
+
+      document.selectedActivity$.next({ new: 'activity' });
+
+      expect(component.state('selectedActivity')).toEqual({ new: 'activity' });
+    });
+
+    it('passes and updater to the chat container', () => {
```
cwhitten: s/and/an

packages/app/client/src/ui/editor/emulator/chatPanel/chatPanel.tsx
```diff
@@ -31,28 +31,74 @@
...
+      this.selectedActivitySub.unsubscribe();
+    }
+  }
+
+  updateSelectedActivity = (activity) => {
```
cwhitten: the naming of this function should better align to its implementation updating the `selectedActivity$` observer and not the internal `selectedActivity` on state. 

also, what if you provided `this.props.document` as a parameter and returned a closure? it cleans up a bit and isn't reliant on `props`.  (working title of function)

```javascript
updateSelectedActivityObservable = (observable) => {
  return (activity) => {
    if (!observable) return;
    observable.next({
       ...activity,
        showInInspector: true 
    })
  }
}
```

```jsx
const { document: { selectedActivity$ }, mode, onStartConversation } = this.props

<ChatContainer
  document={ document }
  mode={ mode }
  onStartConversation={ onStartConversation }
  selectedActivity={ this.state.selectedActivity }
  updateSelectedActivity={ this.updateSelectedActivityObservable(selectedActivity$) }
/>
```

cwhitten:
This needs a rebase, and would conflict heavily with the `prettier` branch. Because we still have tests to run on `prettier` we should land this first.

corinagum:
🍾

PR #1222, opened 2019-01-11
let git handle line endings instead of editor

a-b-r-o-w-n:
This makes developing on multiple platforms easier by allowing git to control which line endings to use on the host system.

EDIT: also adds `CODEOWNERS` file for automatic review requests.


a-b-r-o-w-n:
@tonyanziano Travis is reporting successful builds, but github did not get updated. Any ideas on how to fix?

Issue #1161, opened 2018-12-10
Start a Change Log


tonyanziano:
**Describe the solution you'd like**
I think it would be a good idea for us to start a Change Log or Release Notes so that fixes and new features are more discoverable by our users.

**Describe alternatives you've considered**
As an alternative, we could have an in-app release notes page similar to how VS Code does theirs.

**Additional context**
Could be something like this:

https://keepachangelog.com/en/1.0.0/

tonyanziano:
Addressed in #1230

PR #1230, opened 2019-01-15
Added a changelog.

tonyanziano:
Addresses #1161 

===

Follows the same format as [Webchat's changelog](https://github.com/Microsoft/BotFramework-WebChat/blob/master/CHANGELOG.md).

So for example:

All of our current changes until we release v4.3 will go in the `Unreleased` section. Then, once 4.3 is released, we copy all the changes from `Unreleased` into the `4.3.0` section and wipe out the `Unreleased` section. Then the next set of changes will continue to be added to `Unreleased` until we decide to cut another version (v4.3.1, or v4.4.0, etc.).

Possible change categories under each section include:
 - Added
 - Changed
 - Removed
 - Fixed
 - Deprecated
 - Security

I'm completely open to discussion about this format. [Reference docs here.](https://keepachangelog.com/en/1.0.0/)


corinagum:
Maybe leave open for discussion for a bit, but I'm all for this update :)

CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+### Added
+- Added changelog in PR [#1230](https://github.com/Microsoft/BotFramework-Emulator/pull/1230)
```
a-b-r-o-w-n: I think we should "tag" each change.

ex: `[ui]`, `[core]`, `[build]`, etc.

tonyanziano: So in the case of modifying something like the `README` or this `CHANGELOG` what sort of tag would we use? `[docs]`? or `[infrastructure]`?

a-b-r-o-w-n: I think `docs` makes sense. This is based on https://www.npmjs.com/package/commitizen (which I'd like to introduce at some point), but we can discuss what those tags should be.

PR #1240, opened 2019-01-17
💅 integrate prettier and eslint

### LINKED ISSUE #1235: Move to prettier & eslint
This will increase developer happiness and speed up productivity both in development and code review!

a-b-r-o-w-n:
closes #1235


.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "parser": "typescript",
+  "trailingComma": "es5",
+  "singleQuote": true
+}
```
a-b-r-o-w-n: We can debate here on the prettier formatting options. https://prettier.io/docs/en/options.html

package.json
```diff
@@ -35,6 +40,17 @@
...
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{ts,tsx,js,jsx}": [
+      "prettier --write",
+      "git add"
+    ]
+  },
```
a-b-r-o-w-n: We can potentially expand this to other file types, including markdown and css

cwhitten:
For any observers wondering why eslint instead of tslint, see [this issue on the TypeScript project](https://github.com/Microsoft/TypeScript/issues/29288)

a-b-r-o-w-n:
@tonyanziano and I both did a regression pass which uncovered two unrelated issues (related to #1219 ). Merging this and will file new issues against the regressions.

Issue #1237, opened 2019-01-16
Connect to a bot without a .bot file - revisions


cwhitten:
In the Services pane, replace the `Open Bot` button with copy that says:

The Services caret is default _expanded_ and the content inside is the following:

>NO SERVICES AVAILABLE
>
>To connect the Emulator to services, [open a .bot file](www.example.com) or [create a new bot configuration](www.example.com).
>
>[Learn more about using services.](https://docs.microsoft.com/en-us/azure/bot-service/bot-service-overview-introduction?view=azure-bot-service-4.0#build)

on click of `open a .bot file` - the system file browser opens with .bot files available for selection, and opens the .bot file when the user clicks the enabled `Choose file`

on click of `create a new bot` - the New bot modal opens with the url linked with the current active live chat tab populated in the `Endpoint URL` in the bot configuration form.

justinwilaby:
Reopening after discovering use cases that may break this implementation.

PR #1247, opened 2019-01-22
#1237 - updated implementation to use API instead of internals

justinwilaby:
This PR shows a reduction in code coverage because not all packages were represented in the coverage analysis previously. No coverage was lost with this PR (several more unit tests are in fact included) but it is now accounting for coverage as a whole which reduces coverage % from before.


packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
```diff
@@ -122,16 +173,42 @@ export class OpenBotDialog extends React.Component<
...
-      await this.props.openBot(this.state.botUrl);
+      await this.props.openBot(this.state);
     } catch (e) {
       this.props.sendNotification(e);
     }
   };
+
+  private urlInputRef = (input: HTMLInputElement) => {
+    if (input) {
+      input.addEventListener('input', this.onInputRefChange);
```
a-b-r-o-w-n: Does this have the potential to cause a memory leak?
1. There is nothing preventing this handler from getting added more than once.
2. I don't see anything that removes the listener.

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialogContainer.ts
```diff
@@ -31,37 +31,64 @@
...
+          }
+        );
+        if (result.ok) {
+          return result.text();
+        }
+        throw new Error(
+          `Failed to create a new conversation: ${result.statusText}`
+        );
+      }
+      return ActiveBotHelper.confirmAndOpenBotFromFile(botUrl);
```
a-b-r-o-w-n: This is an odd place to make a network request. Our selectors should be fairly simple. Could this not be moved into a thunked action?

packages/app/client/src/utils/eventHandlers.spec.ts
```diff
@@ -22,7 +22,7 @@
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
 //
-// THE SOFTWARE IS PROVIDED ""AS IS"", WITHOUT WARRANTY OF ANY KIND,
+// THE SOFTWARE IS PROVIDED ''AS IS'', WITHOUT WARRANTY OF ANY KIND,
```
a-b-r-o-w-n: This looks like an editor config issue. Should be reverted.

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
```diff
@@ -36,47 +36,76 @@ import {
...
+      : `Please include route if necessary: "/api/messages"`;
+  }
+
+  private static validateEndpoint(endpoint: string): ValidationResult {
+    if (!endpoint) {
+      return ValidationResult.Empty;
+    }
+
+    if (/(http)(s)?(:\/\/)[\w+]/.test(endpoint)) {
       return endpoint.endsWith('/api/messages')
```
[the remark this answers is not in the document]
justinwilaby: Yes. this is removed from the validation but is still included as a reminder message as per comps

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
justinwilaby: No. Memory leaks only occur when the function reference and the observer do not share the same context. e.g.
```js
// Requires explicit removal else leakage
window.addEventListener('eventType', this.classMemberFunction');

class MyClass {
 private elementRef;
 classMemberFunction() {
  // do stuff
 }

private addListener() {
 // Since elementRef and classMemberFunction share the same 
 // context and are destroyed at the same time, no memory leak is 
 // created and the listener is not required to be removed.
 this.elementRef.addEventListener('eventType', this.classMemberFunction);
 }
}
```

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialogContainer.ts
justinwilaby: We removed thunk as a dependency in favor of sagas. I can move this to a saga but it's a very large amount of work to do so for something that does not interact with state. It might be a good idea but "holy boiler plate batman!"

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
justinwilaby: Also, when react calls this function, we can be assured that the same reference will not be passed twice so no check is necessary. Also, keep in mind that 
```js
object.addEventListener('event', handler);
object.addEventListener('event', handler);
```
Will only add the handler once given the reference to `handler` has not changed.

packages/app/client/src/utils/eventHandlers.spec.ts
justinwilaby: After looking at this in more detail, it seems like `npm run lint:fix` adds single quotes.

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
a-b-r-o-w-n: Gotcha. Instead of adding the event listener directly on the DOM node, we should use `onInput`

packages/app/client/src/utils/eventHandlers.spec.ts
a-b-r-o-w-n: It shouldn't. Prettier does not touch comments.

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
justinwilaby: We cannot. Sadly, any events added in react are not cancelable and have already completed the bubbling phase by the time the handler is hit. Also, react makes no distinction between 'input' and 'change' events for inputs. This has been a source of frustration using a framework that does not maintain parity with the DOM regarding event phases and propagation.

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialogContainer.ts
a-b-r-o-w-n: I'm not sure what you mean by "interact with state". `const { serverUrl, users } = state.clientAwareSettings;` makes this stateful, right?

The problem I have with this is adding business logic to a selector. It's no longer just "mapping state to props". I would still expect to see this represented as a state transition and co-located with other bot specific state actions.

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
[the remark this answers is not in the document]
a-b-r-o-w-n: Ok. This is confusing since this behavior is mixed in with validation logic. Could we separate those?

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
```diff
@@ -122,16 +173,42 @@ export class OpenBotDialog extends React.Component<
...
 
   private onInputChange = (event: ChangeEvent<HTMLInputElement>) => {
-    const { type, files, value } = event.target;
-    const botUrl = type === 'file' ? files.item(0).path : value;
-    this.setState({ botUrl });
+    const {
+      type,
+      files,
+      value,
+      dataset: { prop },
```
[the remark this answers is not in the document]
justinwilaby: I did't use the `name` attr since I cannot guarantee this attribute will be applied to an `<input>` element. I kind of throw it over the fence when using a react component and even though we own the component, a future change could cause this attr to be applied to some other control.

Also, since this is a managed component, all values come from the state and not the `event.elements` array so I felt it was more predictable as a data-attribute.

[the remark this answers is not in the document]
justinwilaby: besides its use in a form element, does using the name attribute have any advantages over data?

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
```diff
@@ -122,16 +173,42 @@ export class OpenBotDialog extends React.Component<
...
+    } else {
+      this.setState({ [prop]: value });
+    }
+  };
+
+  private onInputRefChange = (event: Event) => {
+    const { value } = event.target as HTMLInputElement;
+    try {
+      const { appId, appPassword, endpoint: botUrl } = JSON.parse(value);
+      this.setState({ appId, appPassword, botUrl });
```
a-b-r-o-w-n: I'm having a hard time following this. When will `target.value` be a JSON string?

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
a-b-r-o-w-n: Interesting. I'd like to understand why this combination of event listeners is necessary. To me, nothing too complex is going on that would require us to work outside of the framework.

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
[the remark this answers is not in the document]
a-b-r-o-w-n: It's more predictable as a source of identifying input, ie. principle of least surprise.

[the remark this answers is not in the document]
a-b-r-o-w-n: In the future, it would be wrong to re-purpose the name prop for a form element since that would deviate from common dom attributes.

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.tsx
justinwilaby: It's best if I explain in person. Feel free to drop by.

packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.spec.tsx
```diff
@@ -1,4 +1,6 @@
 //
+import { logEntry, LogLevel, textItem } from '@bfemulator/sdk-shared';
+import { mount } from 'enzyme';
```
a-b-r-o-w-n: This looks out of place.

packages/app/main/src/main.ts
```diff
@@ -1,4 +1,19 @@
...
+  newNotification,
+  Notification,
+  PersistentSettings,
+  Settings,
+  SharedConstants,
+} from '@bfemulator/app-shared';
+import { ProgressInfo } from 'builder-util-runtime';
+import * as Electron from 'electron';
+import { app, systemPreferences } from 'electron';
+import { UpdateInfo } from 'electron-updater';
```
a-b-r-o-w-n: These seem out of place as well. Is your editor doing this or prettier?

packages/emulator/core/src/conversations/middleware/conversationsMiddleware.spec.ts
```diff
@@ -1,4 +1,9 @@
 //
+import {
+  AttachmentData,
+  ConversationParameters,
+  GenericActivity,
+} from '@bfemulator/sdk-shared';
```
a-b-r-o-w-n: out of place

packages/emulator/core/src/utils/activityVisitor.spec.ts
```diff
@@ -1,4 +1,15 @@
...
+  AttachmentContentTypes,
+  CardAction,
+  MediaCard,
+  MessageActivity,
+  OAuthCard,
+  ReceiptCard,
+  SigninCard,
+  ThumbnailCard,
+} from '@bfemulator/sdk-shared';
+
```
a-b-r-o-w-n: out of place.

packages/app/main/src/main.ts
justinwilaby: Looks like my editor is doing this but it's never done this in the past. Its new since the migration to prettier. I'll dig into it.

justinwilaby: Looks like using a multiline comment for the license fixes the issue but this still does not explain why it just started.

justinwilaby: ok, turning off prettier formatting solved this problem. I guess I have to do this manually using `npm run lint:fix` from now on.

a-b-r-o-w-n: hmm.... we could switch to the multi line comment.

justinwilaby:
Merged anyway. Coverage is expected to decrease with this change

PR #1250, opened 2019-01-24
catch error when restarting conversation

### LINKED ISSUE #1241: Restarting conversation fails every other attempt
## Version
4.2.1

## Describe the bug
When trying to restart the conversation with a bot, the restart will clear the log, but not reset webchat, and an error will be thrown in the client console. However, hitting restart again, will resolve the issue and properly restart the conversation and log. This pattern repeats itself.

## To Reproduce
1. Start a livechat session with a bot
2. Click on "Restart conversation"

image

3. Observe that WebChat did not restart the conversation and wipes the log

image

## Expected behavior
The log should reset as well as the conversation.

## Screenshots

Error message:
image

a-b-r-o-w-n:
fixes #1241

This should probably be addressed in direct line, but it's not
immediately clear why the error is being thrown in the first place.

### DIFF
--- a/packages/app/client/src/ui/editor/emulator/emulator.tsx
+++ b/packages/app/client/src/ui/editor/emulator/emulator.tsx
@@ -181,7 +181,14 @@ class EmulatorComponent extends React.Component<EmulatorProps, {}> {
     };
 
     if (props.document.directLine) {
-      props.document.directLine.end();
+      // issue with direct line. #end currently throws an error
+      try {
+        props.document.directLine.end();
+      } catch (e) {
+        if (e.message !== 'conversation ended') {
+          throw e;
+        }
+      }
     }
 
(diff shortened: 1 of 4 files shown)


cwhitten:
Why don't we just land https://github.com/Microsoft/BotFramework-DirectLineJS/pull/133/files ?

packages/app/client/src/ui/editor/emulator/emulator.tsx
```diff
@@ -421,6 +419,19 @@ class EmulatorComponent extends React.Component<EmulatorProps, {}> {
       this.onStartOverClick();
     }
   };
+
+  private endConversation(directLine: any): void {
```
justinwilaby: This can be static since no references to `this` exist here

Issue #1083, opened 2018-11-06
Add Telemetry Infrastructure to Emulator


tonyanziano:
Add telemetry infrastructure to the emulator so that we can see what features of the emulator are being used and how often.

- [ ] language that explains what we're collecting and why
- [ ] opt-out instructions

justinwilaby:
Will this be an opt-in feature?

tonyanziano:
According to @vishwacsena 's spec it will be opt-in by default, and explicit opt-out. We are following the same strategy as az cli.

vishwacsena:
Pushing out to 4.3. @sgellock; @yochay; @Kaiqb FYI.

We need to close on dev design and @tonyanziano will continue to flush that out with the AZ team through 4.2 for implementation in 4.3

PR #1251, opened 2019-01-24
Added App Insights instrumentation.

tonyanziano:
Addresses #1083 

===

This PR adds Instrumentation to the Emulator that will track usage to an App Insights instance. The events being tracked are outlined in the internal document.

### DIFF
--- a/packages/app/client/src/commands/emulatorCommands.ts
+++ b/packages/app/client/src/commands/emulatorCommands.ts
@@ -35,6 +35,7 @@ import { newNotification, SharedConstants } from '@bfemulator/app-shared';
 import {
   Activity,
   CommandRegistryImpl,
+  isLocalHostUrl,
   uniqueId,
 } from '@bfemulator/sdk-shared';
 import { IEndpointService } from 'botframework-config/lib/schema';
@@ -49,7 +50,10 @@ import { CommandServiceImpl } from '../platform/commands/commandServiceImpl';
 
 /** Registers emulator (actual conversation emulation logic) commands */
 export function registerCommands(commandRegistry: CommandRegistryImpl) {
-  const { Emulator } = SharedConstants.Commands;
+  const {
+    Emulator,
+    Telemetry: { TrackEvent },
+  } = SharedConstants.Commands;
 
   // ---------------------------------------------------------------------------
   // Open a new emulator tabbed document
@@ -76,6 +80,12 @@ export function registerCommands(commandRegistry: CommandRegistryImpl) {
             endpointUrl: endpoint.endpoint,
             userId: currentUserId,
           })
+        );
+      }
+
+      if (!isLocalHostUrl(endpoint.endpoint)) {
+        CommandServiceImpl.remoteCall(TrackEvent, 'livechat_openRemote').catch(
+          _e => void 0
         );
       }
 
@@ -140,6 +150,9 @@ export function registerCommands(commandRegistry: CommandRegistryImpl) {
         dialogOptions
       );
       await CommandServiceImpl.call(Emulator.OpenTranscript, filename);
+      CommandServiceImpl.remoteCall(TrackEvent, 'transcriptFile_open', {
+        method: 'file_menu',
+      }).catch(_e => void 0);
     } catch (e) {
       const errMsg = `Error while opening transcript file: ${e}`;
       const notification = newNotification(errMsg);
--- a/packages/app/client/src/data/sagas/azureAuthSaga.ts
+++ b/packages/app/client/src/data/sagas/azureAuthSaga.ts
@@ -66,6 +66,7 @@ export function* getArmToken(
     RetrieveArmToken,
     PersistAzureLoginChanged,
   } = SharedConstants.Commands.Azure;
+  const { TrackEvent } = SharedConstants.Commands.Telemetry;
   azureAuth = yield call(
     CommandServiceImpl.remoteCall.bind(CommandServiceImpl),
     RetrieveArmToken
@@ -80,8 +81,14 @@ export function* getArmToken(
       PersistAzureLoginChanged,
       persistLogin
     );
+    CommandServiceImpl.remoteCall(TrackEvent, 'signIn_success').catch(
+      _e => void 0
+    );
   } else {
     yield DialogService.showDialog(action.payload.loginFailedDialog);
+    CommandServiceImpl.remoteCall(TrackEvent, 'signIn_failure').catch(
+      _e => void 0
+    );
   }
   yield put(azureArmTokenDataChanged(azureAuth.access_token));
   return azureAuth;
--- a/packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.tsx
+++ b/packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.tsx
@@ -73,6 +73,7 @@ interface AppSettingsEditorState {
 const defaultAppSettings: FrameworkSettings = {
   autoUpdate: true,
   bypassNgrokLocalhost: true,
+  collectUsageData: true,
   locale: '',
   localhost: '',
   ngrokPath: '',
@@ -183,9 +184,10 @@ export class AppSettingsEditor extends React.Component<
             <Checkbox
               className={styles.checkboxOverrides}
               checked={uncommitted.bypassNgrokLocalhost}
-              onChange={this.onChangeNgrokBypass}
+              onChange={this.onChangeCheckBox}
               id="ngrok-bypass"
               label="Bypass ngrok for local addresses"
+              name="bypassNgrokLocalhost"
             />
             <Row align={RowAlignment.Center} className={styles.marginBottomRow}>
               <TextField
@@ -215,31 +217,50 @@ export class AppSettingsEditor extends React.Component<
             <Checkbox
               className={styles.checkboxOverrides}
               checked={uncommitted.use10Tokens}
-              onChange={this.onChangeAuthTokenVersion}
+              onChange={this.onChangeCheckBox}
               id="auth-token-version"
               label="Use version 1.0 authentication tokens"
+              name="use10Tokens"
             />
             <SmallHeader>Sign-in</SmallHeader>
             <Checkbox
               className={styles.checkboxOverrides}
               checked={uncommitted.useCodeValidation}
-              onChange={this.onChangeUseValidationToken}
+              onChange={this.onChangeCheckBox}
               id="use-validation-code"
               label="Use a sign-in verification code for OAuthCards"
+              name="useCodeValidation"
             />
             <SmallHeader>Application Updates</SmallHeader>
             <Checkbox
               className={styles.checkboxOverrides}
               checked={uncommitted.autoUpdate}
-              onChange={this.onChangeAutoInstallUpdates}
+              onChange={this.onChangeCheckBox}
               label="Automatically download and install updates"
+              name="autoUpdate"
             />
             <Checkbox
               className={styles.checkboxOverrides}
               checked={uncommitted.usePrereleases}
-              onChange={this.onChangeUsePrereleases}
+              onChange={this.onChangeCheckBox}
               label="Use pre-release versions"
+              name="usePrereleases"
             />
+            <SmallHeader>Data Collection</SmallHeader>
+            <Checkbox
+              className={styles.checkboxOverrides}
+              checked={uncommitted.collectUsageData}
+              onChange={this.onChangeCheckBox}
+              label="Help improve the Emulator by allowing us to collect usage data."
+              name="collectUsageData"
+            />
+            <a
+              target="_blank"
+              href="https://aka.ms/bot-framework-emulator-data-collection"
+              rel="noopener noreferrer"
+            >
+              Learn more.
+            </a>
           </Column>
         </Row>
         <Row className={styles.buttonRow} justify={RowJustification.Right}>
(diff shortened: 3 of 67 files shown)


packages/app/client/src/commands/emulatorCommands.ts
```diff
@@ -221,3 +229,14 @@ export function registerCommands(commandRegistry: CommandRegistryImpl) {
     }
   );
 }
+
+function isLocalHostUrl(url: string): boolean {
```
justinwilaby: Looks like this already exists at `packages/app/main/src/utils/isLocalhostUrl.ts`. This would be a good candidate to move to a shared package

packages/app/client/src/data/sagas/azureAuthSaga.ts
```diff
@@ -80,8 +81,10 @@ export function* getArmToken(
       PersistAzureLoginChanged,
       persistLogin
     );
+    CommandServiceImpl.remoteCall(TrackEvent, 'signIn_success');
```
justinwilaby: These remote calls should probably be yielded so exceptions can be handled.

packages/app/client/src/data/sagas/azureAuthSaga.ts
```diff
@@ -80,8 +81,10 @@ export function* getArmToken(
...
   } else {
     yield DialogService.showDialog(action.payload.loginFailedDialog);
+    CommandServiceImpl.remoteCall(TrackEvent, 'signIn_failure');
```
justinwilaby: same here: consider yielding the remote call

packages/app/client/src/ui/editor/emulator/emulator.tsx
```diff
@@ -376,11 +376,15 @@ class EmulatorComponent extends React.Component<EmulatorProps, {}> {
     option: string = RestartConversationOptions.NewUserId
   ): Promise<void> => {
     const { NewUserId, SameUserId } = RestartConversationOptions;
+    const { TrackEvent } = SharedConstants.Commands.Telemetry;
     this.props.clearLog(this.props.document.documentId);
     this.props.setInspectorObjects(this.props.document.documentId, []);
 
     switch (option) {
       case NewUserId: {
+        CommandServiceImpl.remoteCall(TrackEvent, 'conversation_restart', {
```
justinwilaby: This promise is not handled. Also, consider placing this in the `emulatorContainer.ts` so you can handle any exceptions and dispatch a notification when caught (if applicable):
e.g. sudo code...
```js
const mapDispatchToProps = (dispatch) => (
  {
     trackEvent: async (eventName: string, data: any) => {
      try {
          await CommandService.remoteCall(//...);
      } catch (e) {
        dispatch(notification);
  }
);
```

packages/app/client/src/ui/editor/emulator/emulator.tsx
```diff
@@ -392,6 +396,9 @@ class EmulatorComponent extends React.Component<EmulatorProps, {}> {
       }
 
       case SameUserId:
+        CommandServiceImpl.remoteCall(TrackEvent, 'conversation_restart', {
```
justinwilaby: Same here. Consider handling the promise in the container and dispatching a notification (if appropriate).

packages/app/client/src/ui/editor/emulator/parts/inspector/inspectorContainer.ts
```diff
@@ -50,7 +53,19 @@ const mapStateToProps = (state: RootState, ownProps: any) => {
   };
 };
 
+const mapDispatchToProps = _dispatch => {
+  return {
+    trackEvent: (name: string, properties?: { [key: string]: any }) => {
+      CommandServiceImpl.remoteCall(
```
justinwilaby: The promise here is unhandled. Consider handling it in some meaningful way if necessary

packages/app/client/src/ui/shell/mdi/tabBar/tabBarContainer.ts
```diff
@@ -67,10 +69,22 @@ const mapDispatchToProps = (dispatch): TabBarProps => ({
     documentId: string,
     srcEditorKey: string,
     destEditorKey: string
-  ) => dispatch(splitTab(contentType, documentId, srcEditorKey, destEditorKey)),
+  ) => {
+    CommandServiceImpl.remoteCall(
```
justinwilaby: unhandled promise

packages/app/client/src/ui/shell/mdi/tabBar/tabBarContainer.ts
```diff
@@ -67,10 +69,22 @@ const mapDispatchToProps = (dispatch): TabBarProps => ({
...
+      SharedConstants.Commands.Telemetry.TrackEvent,
+      'tabBar_splitTab'
+    );
+    dispatch(splitTab(contentType, documentId, srcEditorKey, destEditorKey));
+  },
   appendTab: (srcEditorKey: string, destEditorKey: string, tabId: string) =>
     dispatch(appendTab(srcEditorKey, destEditorKey, tabId)),
-  enablePresentationMode: () => dispatch(enablePresentationMode()),
+  enablePresentationMode: () => {
+    CommandServiceImpl.remoteCall(
```
justinwilaby: here as well

packages/app/client/src/ui/shell/navBar/navBarContainer.ts
```diff
@@ -60,6 +63,13 @@ const mapDispatchToProps = (dispatch): NavBarProps => ({
       })
     );
   },
+  trackEvent: (name: string, properties?: { [key: string]: any }) => {
+    CommandServiceImpl.remoteCall(
```
justinwilaby: unhandled promise

packages/app/main/src/main.ts
```diff
@@ -506,6 +512,15 @@ const createMainWindow = async () => {
     saveSettings<PersistentSettings>('server.json', getSettings());
     Electron.app.quit();
   });
+
+  // log app startup time in seconds
+  const endStartupTime = Date.now();
+  const startupTime = (endStartupTime - beginStartupTime) / 1000;
```
justinwilaby: should this be in the `ready-to-show` handler to account for the rendering process startup time too?

packages/app/main/src/telemetry/telemetryService.ts
```diff
@@ -0,0 +1,83 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+import * as AppInsights from 'applicationinsights';
+import { SettingsImpl } from '@bfemulator/app-shared';
+
+import { getSettings } from '../settingsData/store';
+
+const INSTRUMENTATION_KEY = '631faf57-1d84-40b4-9a71-fce28a3934a8';
```
a-b-r-o-w-n: Is this secret?

packages/app/main/src/telemetry/telemetryService.ts
```diff
@@ -0,0 +1,83 @@
...
+    return framework.collectUsageData;
+  }
+
+  private static startup(): void {
+    if (!this._hasStarted) {
+      AppInsights.setup(INSTRUMENTATION_KEY)
+        // turn off extra instrmentation
+        .setAutoCollectConsole(false)
+        .setAutoCollectDependencies(false)
+        .setAutoCollectExceptions(false)
```
a-b-r-o-w-n: This one could be interesting. Why do we not want to collect exceptions?

packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.tsx
```diff
@@ -271,6 +286,12 @@ export class AppSettingsEditor extends React.Component<
     });
   };
 
+  private onChangeCollectUsageData = (): void => {
+    this.setUncommittedState({
+      collectUsageData: !this.state.uncommitted.collectUsageData,
+    });
```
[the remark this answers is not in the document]
tonyanziano: Your first comment echoes what I was thinking. I thought it would be best if the app would avoid crashing if something failed while trying to track telemetry. However, I still believe the error should be swallowed inside of the `TelemetryService` so that it doesn't bubble up to the Command Service and become uncaught.

packages/app/main/src/telemetry/telemetryService.ts
tonyanziano: It is the key that allows anybody to post events to our App Insights instance, yes. I've spoken to the Azure CLI team about how they implemented instrumentation and [they expose the key to the public in the same manner.](https://github.com/Azure/azure-cli/blob/d7a4b887fc2a0d907191be846228b9b528003ee7/src/azure-cli-core/azure/cli/core/telemetry.py#L23)

There is an email thread I can forward to you if you're interested, but basically the rationale behind it is this:

- This does not increase exposure to DDoS attacks, because even if we abstracted the key, someone could write a script or use an auto clicker to perform some instrumented action in the Emulator and hammer our App Insights instance if they wanted to.
- Internal data is not exposed in any way through this key. This only gives the ability to push data into the App Insights instance.
- We can just scrub and throw away bad / garbage events

packages/app/main/src/telemetry/telemetryService.ts
tonyanziano: I just read the docs and it looks like this tracks all uncaught exceptions to App Insights. I don't really see a reason why we shouldn't collect these, but it is out of scope for the instrumentation spec. We can always flip the flag down the road.

packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.tsx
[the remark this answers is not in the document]
tonyanziano: Re: second comment

What would be a better pattern? 
Pass the event into the handler and read the `checked` value of the element?

Currently all the checkboxes in `AppSettingsEditor` use this same pattern, but I'm open to changing them to one that is more reliable and wouldn't be at risk of falling out of sync.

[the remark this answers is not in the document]
justinwilaby: This can be:

```js
this.setUncommittedState({
      get collectUsageData: !this.state.uncommitted.collectUsageData,
    });
```
Which will read from the state at the moment the getter is invoked.

packages/app/client/src/data/sagas/azureAuthSaga.ts
tonyanziano: However, the command service could still throw, so would adding a catch like this to each of the command calls to `TrackEvent` be sufficient?

`commandService.remoteCall(TrackEvent, ...).catch(e => void 0);`

packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.tsx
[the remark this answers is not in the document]
justinwilaby: This can be:

```js
const state = Object.create({}, {
  collectUsageData: {
    get: () => !this.state.uncommitted.collectUsageData,
    enumerable: true // important since rest spread is used.
  }
}):
this.setUncommittedState(state);
```
Which will read from the state at the moment the getter is invoked.

packages/app/client/src/data/sagas/azureAuthSaga.ts
a-b-r-o-w-n: That seems reasonable to me. I think your reasoning is solid about not needing to wait for the telemetry as well.

packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.tsx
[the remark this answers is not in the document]
a-b-r-o-w-n: That suggestion might be a little too complex. I think in general, the better approach would be to refactor the `setUncommittedState` method to allow an updater function instead that gets invoked with previous state. But as I said, I don't think it will be an issue in this case unless someone spams the checkbox.

packages/app/main/src/telemetry/telemetryService.ts
a-b-r-o-w-n: That's fine with me. Just checking.

Is there a common place for config values such as this? If not, we may consider creating one. Even a config package. No need to update if not, but something to keep in mind. (I can think of a few other things that could live in that module as well)

packages/app/main/src/telemetry/telemetryService.ts
a-b-r-o-w-n: :+1:

packages/app/main/src/telemetry/telemetryService.ts
tonyanziano: No official config file as far as I know. We do have a `.env` file in the root for using Travis, but that's about as close as it gets.

packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.tsx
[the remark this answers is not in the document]
tonyanziano: I've gone with a refactor including Justin's suggestion for now. However, I think we could investigate your approach down the road because we need to refactor `AppSettingsEditor.tsx` anyway and write a test for it, which ideally would be done at the same time.

packages/app/main/src/main.ts
tonyanziano: Good call.

PR #1279, opened 2019-01-31
Removed .zip build target from mac build. Added file pattern to ignor…

tonyanziano:
…e config files when creating build artifact.


tonyanziano:
Yeah this one was a little long. I'll try to shorten it next time.

PR #1280, opened 2019-02-01
Fix theme menu state on mac

### LINKED ISSUE #1239: Currently selected Theme does not appear in Menu like it should
## Version
4.2.1

## Describe the bug
Selecting different Themes via the File menu are not mutually exclusive

## To Reproduce
1. Click File -> Themes -> Dark
2. Click File -> Themes -> Light
3. Click File -> Themes
4. See in the menu that both Light and Dark are checked, but Light is applied.

## Expected behavior
Only one Theme should be indicated

## Screenshots
screen shot 2019-01-17 at 10 12 42 am

a-b-r-o-w-n:
closes #1239 

On a mac, the `radio` type UI is the exact same as `checkbox` type, but functions correctly. This has not effect on Windows or Linux.

### DIFF
--- a/packages/app/main/src/appMenuBuilder.spec.ts
+++ b/packages/app/main/src/appMenuBuilder.spec.ts
@@ -234,7 +234,7 @@ describe('AppMenuBuilder', () => {
   });
 
   it('should update the recent bots list', () => {
-    mockRemoteCall = jest.fn((..._args) => Promise.resolve(null));
+    mockRemoteCall = jest.fn(() => Promise.resolve(null));
     const mockBotPath = join('path', 'to', 'bot');
     const mockRecentBots = [
       { displayName: 'bot1', path: mockBotPath },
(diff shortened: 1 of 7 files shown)


packages/app/main/src/appMenuBuilder.spec.ts
```diff
@@ -308,7 +308,7 @@ describe('AppMenuBuilder', () => {
       },
     };
     mockRemoteCall = jest.fn(commandName => {
-      if ((commandName = SharedConstants.Commands.Misc.GetStoreState)) {
+      if (commandName === SharedConstants.Commands.Misc.GetStoreState) {
```
a-b-r-o-w-n: Someone comment about this if this was previously correct.

tonyanziano: I definitely wrote those tests and that error must have slipped through. The updated version looks good to me.

PR #1297, opened 2019-02-04
Fix restart conversation

a-b-r-o-w-n:


packages/app/client/src/ui/editor/emulator/emulator.tsx
```diff
@@ -335,6 +335,16 @@ export class EmulatorComponent extends React.Component<EmulatorProps, {}> {
     };
   };
 
+  private getConversationId() {
```
justinwilaby: Should this be an accessor?

packages/app/client/src/ui/editor/emulator/emulator.tsx
```diff
@@ -335,6 +335,16 @@ export class EmulatorComponent extends React.Component<EmulatorProps, {}> {
...
+    const { document } = this.props;
+
+    if (document && document.conversationId) {
```
justinwilaby: This can optionally be simplified: 

```js
return document ? document.conversationId : 'default-conversation';
```

packages/app/client/src/ui/editor/emulator/emulator.tsx
a-b-r-o-w-n: I'm not sure it matters too much. If I were to cache the result, then yes, but this doesn't really represent a property of the chat class. This could have easily lived in the render function.

PR #1302, opened 2019-02-06
Added production AI instrumentation key.

tonyanziano:


packages/app/main/src/telemetry/telemetryService.ts
```diff
@@ -36,7 +36,7 @@ import { SettingsImpl } from '@bfemulator/app-shared';
 
 import { getSettings } from '../settingsData/store';
 
-const INSTRUMENTATION_KEY = '631faf57-1d84-40b4-9a71-fce28a3934a8';
+const INSTRUMENTATION_KEY = 'cc91f74a-1abd-4be0-8a96-fddedbc08dd7';
```
a-b-r-o-w-n: What was the other key we were using? Should we still use that key for dev / test environment?

tonyanziano: The other key pointed to an instance in my own personal subscription that I set up. I'm going to get rid of that one. If we ever do need to do testing it would be really easy to stand up another instance and swap the key out.

PR #1315, opened 2019-02-13
Data url links will now prompt to save to disk.

### LINKED ISSUE #1130: Download link opens Windows Store and fails to download
_From @paulio on August 8, 2018 15:19_

Create an embedded text document launches the Windows Store rather than download the file
``` 
var contentType = "text/plain";
var attachment = new Attachment
{
    Name = "Transcript.txt",
    ContentUrl = $"data:{contentType};base64,{Convert.ToBase64String(bytes)}",
    ContentType = contentType
};
```

_Copied from original issue: Microsoft/BotFramework-Emulator#752_

tonyanziano:
Fixes #1130 

===

image

### DIFF
--- a/packages/app/client/src/hyperlinkHandler.ts
+++ b/packages/app/client/src/hyperlinkHandler.ts
@@ -40,7 +40,7 @@ import { CommandServiceImpl } from './platform/commands/commandServiceImpl';
 const Electron = (window as any).require('electron');
 const { shell } = Electron;
 
-export function navigate(url: string) {
+export function navigate(url: string = '') {
   const { TrackEvent } = SharedConstants.Commands.Telemetry;
   try {
     const parsed = URL.parse(url) || { protocol: '' };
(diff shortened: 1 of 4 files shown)


packages/app/client/src/hyperlinkHandler.ts
```diff
@@ -50,7 +50,16 @@ export function navigate(url: string) {
...
     } else {
       CommandServiceImpl.remoteCall(TrackEvent, 'app_openLink', { url }).catch(_e => void 0);
-      shell.openExternal(url, { activate: true });
+      // manually create and click a download link for data url's
+      if (url.startsWith('data:')) {
+        const a = document.createElement('a');
+        a.href = url;
+        a.download = '';
+        a.click();
+        a.remove();
```
justinwilaby: This is inconsequential since the element is never being added to the dom. It can probably be removed.

tonyanziano: remove the remove?!?

tonyanziano: I'll change it when I inevitably have to update the branch once the azure gov fix goes into master.

PR #1325, opened 2019-02-21
Fixed issue where restart conversation was not clearing history.

### LINKED ISSUE #1323: Latest version of Emulator does not clear the conversation history when "Restart conversation" is clicked
## Version
4.2.142494

## Describe the bug
Emulator does not clear the conversation history when "Restart conversation" is clicked

## To Reproduce
1. Enter text into the conversation panel
2. Click on 'Restart conversation'
3. Observe that the conversation history is still present

## Screenshot

image

## Expected behavior
The conversation history should disappear

[bug]

tonyanziano:
Fixes #1323

### DIFF
--- a/packages/app/client/src/ui/editor/emulator/emulator.tsx
+++ b/packages/app/client/src/ui/editor/emulator/emulator.tsx
@@ -147,7 +147,11 @@ export class EmulatorComponent extends React.Component<EmulatorProps, {}> {
     }
   }
 
-  startNewConversation = async (props: EmulatorProps = this.props): Promise<any> => {
+  startNewConversation = async (
+    props: EmulatorProps = this.props,
+    requireNewConvoId: boolean = false,
+    requireNewUserId: boolean = false
+  ): Promise<any> => {
     if (props.document.subscription) {
       props.document.subscription.unsubscribe();
     }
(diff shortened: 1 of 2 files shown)


packages/app/client/src/ui/editor/emulator/emulator.tsx
```diff
@@ -360,18 +374,17 @@ export class EmulatorComponent extends React.Component<EmulatorProps, {}> {
         this.props.trackEvent('conversation_restart', {
           userId: 'new',
         });
-        const newUserId = uniqueIdv4();
-        // set new user as current on emulator facilities side
-        await CommandServiceImpl.remoteCall(SharedConstants.Commands.Emulator.SetCurrentUser, newUserId);
-        this.props.updateChat(this.props.documentId, { userId: newUserId });
+        // start conversation with new convo id & user id
+        this.startNewConversation(undefined, true, true);
```
compulim: Why don't we create unique ID before we call `startNewConversation`? E.g. `this.startNewConversation({ ...this.props, document: { ...this.props.document, conversationId: uniqueIdv4(), userId: uniqueIdv4() } });` And can we always call `setCurrentUser` every time we call `startNewConversation`?

Your existing logic may be good or have some reasons behind, I just didn't see through it yet. If there are any, please consider me approved. 😅

tonyanziano: I don't see why this approach wouldn't work. I actually like it a lot more than making a function call like `(undefined, true, true)` -- seems odd

packages/app/client/src/ui/editor/emulator/emulator.tsx
```diff
@@ -360,18 +374,17 @@ export class EmulatorComponent extends React.Component<EmulatorProps, {}> {
...
         break;
       }
 
       case SameUserId:
         this.props.trackEvent('conversation_restart', {
           userId: 'same',
         });
-        this.startNewConversation();
+        // start conversation with new convo id
+        this.startNewConversation(undefined, true);
```
compulim: Or should it be like this?

```suggestion
        this.startNewConversation(undefined, true, false);
```

packages/app/client/src/ui/editor/emulator/emulator.tsx
tonyanziano: I tried this approach out and it actually causes new livechat windows to be opened on every restart, as well as prevents webchat from connecting to directline when restarting with a new user id. I'm not entirely sure why though. I believe it might have to do with setting the current user every time.

PR #1335, opened 2019-02-28
Fixed broken luis inspector.

tonyanziano:
Cherry pick of #1334 into master.


tonyanziano:
Rebased and re-pushed with Chris' change

PR #1400, opened 2019-03-28
Sidecar debugging - Milestone 0.9

justinwilaby:
"Sidecar" Debugger progress


justinwilaby:
image

packages/app/client/src/data/sagas/chatSagas.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+  const value = source[key];
+  delete target[key];
+  target[prependWith + key] = value;
+}
+
+export function* chatSagas(): IterableIterator<ForkEffect> {
+  yield takeEvery(ChatActions.showContextMenuForActivity, showContextMenuForActivity);
+  yield takeEvery(ChatActions.closeConversation, closeConversation);
+}
```
a-b-r-o-w-n: I'd like to see some test coverage on this, especially around the object diffing.

packages/emulator/core/src/conversations/middleware/getConversations.ts
```diff
@@ -0,0 +1,83 @@
...
+  return (req: Restify.Request, res: Restify.Response, next: Restify.Next): void => {
+    const conversationParameters: ConversationsGetConversationsOptionalParams = req.params;
+    const { continuationToken } = conversationParameters;
+    const { conversations: api } = botEmulator.facilities;
+    const conversations = api.getConversationIds().map(id => api.conversationById(id));
+    const response = {} as ConversationsGetConversationsResponse;
+
+    if (continuationToken) {
+      response.conversations = conversations
+        .filter(skipWhileExpression<Conversation>(a => a.conversationId !== continuationToken))
```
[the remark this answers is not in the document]
justinwilaby: This strategy was taken from the botbuilder-js repo [here](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/src/memoryTranscriptStore.ts#L174) with a similar implementation [here](https://github.com/Microsoft/botbuilder-js/blob/2115012394877abcfc8234ebad43d09b9ac2d9d0/libraries/botbuilder/src/fileTranscriptStore.ts#L62). I'd like to maintain parity with the botframework since they have likely implemented continuation tokens from the backend team.

packages/app/client/src/data/sagas/chatSagas.ts
justinwilaby: Coverage at 100% is my goal for today.

packages/extensions/botState/src/index.ts
```diff
@@ -0,0 +1,236 @@
...
+      const child = { name: key } as HierarchicalData;
+      if (data[key] !== null && typeof data[key] === 'object') {
+        child.children = [];
+        BotStateVisualizer.getChildren(data[key], child);
+      } else {
+        child.value = data[key];
+      }
+      parent.children.push(child);
+    });
+  }
```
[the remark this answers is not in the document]
justinwilaby: I'll update the name to better describe it's purpose.

packages/extensions/botState/src/index.ts
```diff
@@ -0,0 +1,236 @@
...
+      .attr('stroke-linejoin', 'round')
+      .attr('stroke-width', 3)
+      .selectAll('g')
+      .data(root.descendants().reverse())
+      .join('g')
+      .attr('transform', d => `translate(${d.y},${d.x})`);
+
+    node
+      .append('circle')
+      .attr('fill', d => (d.children ? '#555' : '#999'))
```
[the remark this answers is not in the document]
justinwilaby: I'll implement theming once designs are complete.

packages/emulator/core/src/conversations/middleware/getConversations.ts
```diff
@@ -0,0 +1,83 @@
...
+    return !skipping;
+  };
+};
+
+export default function fetchConversation(botEmulator: BotEmulator) {
+  return (req: Restify.Request, res: Restify.Response, next: Restify.Next): void => {
+    const conversationParameters: ConversationsGetConversationsOptionalParams = req.params;
+    const { continuationToken } = conversationParameters;
+    const { conversations: api } = botEmulator.facilities;
+    const conversations = api.getConversationIds().map(id => api.conversationById(id));
```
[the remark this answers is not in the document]
justinwilaby: I wanted to reduce the number of unit tests I'd have to write by using the existing APIs. :) Adding more functions means unit tests must follow :\

packages/extensions/botState/src/index.ts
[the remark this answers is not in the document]
a-b-r-o-w-n: Unless there is a requirement to mutate, we should avoid it.

packages/app/client/src/data/sagas/chatSagas.ts
```diff
@@ -0,0 +1,187 @@
...
+  Object.keys(currentStatePaths).forEach(path => {
+    if (path in previousStatePaths && previousStatePaths[path] === currentStatePaths[path]) {
+      delete currentStatePaths[path];
+      delete previousStatePaths[path];
+    }
+  });
+
+  // Clone the bot state and update the keys to show changes
+  const diff: Activity = JSON.parse(JSON.stringify(currentBotState));
+  diff.valueType = 'https://www.botframework.com/schemas/diff';
```
cwhitten: what is this for?

packages/app/client/src/data/sagas/chatSagas.ts
```diff
@@ -0,0 +1,187 @@
...
+    if (obj[key] !== null && typeof obj[key] === 'object') {
+      createPathMap(obj[key], paths, path);
+    } else {
+      paths[path] = obj[key];
+    }
+  }
+  return paths;
+}
+
+function buildDiff(prependWith: string, path: string, target: any, source: any): void {
```
cwhitten: I'm wondering if https://github.com/flitbit/diff#readme would be the right foundation for the diffing. Did you look at this?

packages/emulator/core/src/conversations/middleware/getConversations.ts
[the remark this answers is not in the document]
a-b-r-o-w-n: parity != identical implementation. The purpose of this piece of code is to select all conversations after the continuation token, however, it does so with a layer of indirection that is not necessary and is less performant than my suggestion. There is nothing domain-specific about this predicate function so I would rather see us improve it.

packages/app/main/src/botFrameworkService.ts
```diff
@@ -66,7 +66,7 @@ export class BotFrameworkService {
 
     this.server = new RestServer();
 
-    const { url, port } = await this.server.listen();
+    const { url, port } = await this.server.listen(9000);
```
cwhitten: what happens if this port is not available?

packages/emulator/core/src/conversations/middleware/getConversations.ts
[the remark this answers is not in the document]
a-b-r-o-w-n: Also, these conversations are not guaranteed to be sorted (at least from what I can see).

packages/extensions/botState/bf-extension.json
```diff
@@ -0,0 +1,25 @@
...
+  "name": "Bot State Visualizer",
+  "client": {
+    "inspectors": [
+      {
+        "name": "Bot State Visualizer",
+        "src": "public/index.html",
+        "criteria": [
+          {
+            "path": "valueType",
+            "value": "https://www.botframework.com/schemas/botState"
```
cwhitten: if we need to use these strings, let's reference them from this file.

packages/app/client/src/data/action/debugModeAction.ts
```diff
@@ -0,0 +1,52 @@
...
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+import { DebugMode } from '@bfemulator/app-shared';
+
+export const SWITCH_DEBUG_MODE = 'switchDebugMode';
+export declare type DebugType = 'switchDebugMode';
```
tonyanziano: Why is the `declare` keyword used here? My understanding is that the `declare` keyword in Typescript lets the compiler know that the class / type / variable has been defined somewhere in the program and is available, but that seems unnecessary here since you are explicitly declaring it for the first time.

packages/app/client/src/ui/editor/emulator/parts/log/logEntry.tsx
```diff
@@ -148,38 +161,38 @@ export class LogEntry extends React.Component<LogEntryProps> {
     );
   }
 
-  renderItem(item: ILogItem, key: string) {
+  renderItem(item: LogItem, key: string) {
```
tonyanziano: Once this gets rebased onto latest, it would be great if you could perform the same refactor on the LuisDeepLink log item I added. 👍

packages/app/client/src/data/sagas/chatSagas.ts
justinwilaby: This is used to identify the contents of the `value` field in the trace activity. It's arbitrary and only internally used.

packages/app/client/src/data/sagas/chatSagas.ts
justinwilaby: No. I looked at several others but my diff strategy was so trivial, I couln't justify bringing in a 3rd party dependency.

packages/extensions/botState/bf-extension.json
justinwilaby: Since this file us part of the `extensions` project and extensions are not libraries, how would this file get pulled into another package?

packages/app/client/src/data/action/debugModeAction.ts
justinwilaby: I'll Remove it.

packages/emulator/core/src/conversations/middleware/getConversations.ts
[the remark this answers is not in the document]
justinwilaby: This is a [.net language feature](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.skipwhile?view=netframework-4.7.2). IMO, with your suggestion, using `reduce` to perform a `filter` detaches intent from the lexical grammar. Subjectively, one could argue that this adds cognitive overhead since the reader would have to understand that this does not `reduce` but instead `filters` the array which breaks the "Principle of least astonishment". If you don't mind, the existing pattern is an established one that is used across the frameworks so I'd like to keep it as-is.

packages/app/main/src/botFrameworkService.ts
justinwilaby: It would be similar to the behavior of your locally running bot if port 3978 isn't available. This is  a loose end to tie up. Since sidecar debugging requires the Emulator to be discoverable on a fixed port, but normal mode does not, I need to restart the internal server accordingly when the user switches to it.

packages/emulator/core/src/conversations/middleware/getConversations.ts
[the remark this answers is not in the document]
a-b-r-o-w-n: @justinwilaby I'd appreciate if we could actually come to a resolution before marking these conversations as such. 

I do not think that "This is a .net language feature" is satisfactory in determining how we should implement features in typescript. Typescript is more functional and provides better utilities to operate on lists.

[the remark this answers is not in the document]
justinwilaby: I really appreciate the feedback and want to thank you for taking the time to comment. I'm really not overly concerned with it. If you feel strongly about it, I'll gladly implement your way. 

I do think we need to leave feedback based on subjectivity up to the author to decide however. While you found the this implementation to be confusing, I did not. Nor did the authors of this implementation in the BotFramework. Suggestions are great but the notion of mandating them after they are made does not seem to fit the character of what PRs are supposed to accomplish. 

I acknowledged your suggestion and kindly disagreed :) Also, the existing solution makes use of TypeScript while your suggestion appears to be 100% JS. Could elaborate on how your solution is better rather than just another way to accomplish the same thing? I'm open to convincing ;)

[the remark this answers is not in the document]
cwhitten: It's often a good idea to use language built-ins over custom code. In this case using `reduce` relies more on the language constructs than a custom closure implementation. The closure comes off as overly clever.

Some general thoughts:

1. This code that you copied from botbuilder-js was contributed initially from Southworks over a year ago and likely a referencing implementation from C# concept. It is not idiomatic JS.
2. If this convention was meant to be reused across all BotFramework projects, a shared library should probably exist and instead of just copying code over like this. That doesn't exist so we shouldn't feel like we need to reach to botbuilder-js every time we want to accomplish something.
3. Writing code that your team understands is sometimes more important than writing code that you understand.

packages/app/client/src/data/sagas/chatSagas.ts
cwhitten: see my magic string comment earlier

packages/app/client/src/data/sagas/chatSagas.ts
cwhitten: While your diff strategy may be trivial right now, the diff strategy(strategies) we mature into may not be, and to use something. For example `deep-diff` lets your separate values added from values edited. I also like the deep-diff api more than yours. 

`diff(lhs, rhs[, options, acc])` — calculates the differences between two objects, optionally using the specified accumulator.

`observableDiff(lhs, rhs, observer[, options])` — calculates the differences between two objects and reports each to an observer function.

`applyDiff(target, source, filter)` — applies any structural differences from a source object to a target object, optionally filtering each difference.

`applyChange(target, source, change)` — applies a single change record to a target object. NOTE: source is unused and may be removed.

`revertChange(target, source, change)` -- reverts a single change record to a target object. NOTE: source is unused and may be removed.

`redux-dev-tools` uses this as their underlying diffing api.

packages/extensions/botState/bf-extension.json
cwhitten: What I'm saying is that if you need to use magic strings, we should probably define & them in one place.

packages/extensions/botState/src/BotStateVisualizer.ts
```diff
@@ -0,0 +1,230 @@
...
+      const oldThemeComponents = document.querySelectorAll('[data-theme-component="true"]');
+      const head = document.querySelector('head');
+      const fragment = document.createDocumentFragment();
+      const promises = [];
+      // Create the new links for each theme component
+      themeInfo.themeComponents.forEach(themeComponent => {
+        const link = document.createElement('link');
+        promises.push(
+          new Promise(resolve => {
+            link.addEventListener('load', resolve);
```
cwhitten: How will these listeners get cleaned up?

packages/app/client/src/ui/editor/emulator/parts/log/logEntry.tsx
justinwilaby: Done!

packages/extensions/botState/bf-extension.json
justinwilaby: Got it. I'll see if I can get an enum for these values in the `botbuilder-schema` package.
On a side note, interestingly enough, the term ["Magic string"](https://en.wikipedia.org/wiki/Magic_string) does not describe a mere hard coded string. I think this was adopted because strings can be magical but only in the context of direct user input.

packages/extensions/botState/src/BotStateVisualizer.ts
justinwilaby: They do not need cleaning up manually since the GC does this for us. `link` references `resolve` and `resolve` is anonymous so it is gc'd correctly when `link` goes out of scope.

packages/app/client/src/data/sagas/chatSagas.ts
justinwilaby: I'll move this to an enum

packages/app/client/src/data/sagas/chatSagas.ts
justinwilaby: Ok. I'll implement this.

packages/emulator/core/src/conversations/middleware/getConversations.ts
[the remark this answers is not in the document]
justinwilaby: Ok. Great! I'm convinced. I'll implement Andy's strategy. Thank you for clarifying.

packages/app/client/src/data/sagas/chatSagas.ts
```diff
@@ -0,0 +1,206 @@
...
+  );
+  botStateClone.valueType = 'https://www.botframework.com/schemas/diff';
+  // values that were added
+  rhs.forEach(path => {
+    buildDiff('+', path, botStateClone.value, botStateClone.value);
+  });
+  // values that were removed
+  lhs.forEach(path => {
+    buildDiff('-', path, botStateClone.value, previousBotState.value);
+  });
```
a-b-r-o-w-n: I think this is still really coupled with the current view of the diff. Is there any way to apply this at the view layer since it is a view concern?

packages/emulator/core/src/facility/conversationSet.ts
```diff
@@ -62,17 +62,25 @@ export default class ConversationSet extends EventEmitter {
...
-    return this.conversations.map(conversation => conversation.conversationId);
+    return Object.keys(this.conversations);
+  }
+
+  public deleteConversation(conversationId: string): boolean {
+    return delete this.conversations[conversationId];
+  }
+
+  public getConversations(): Conversation[] {
+    return Object.keys(this.conversations).map(id => this.conversations[id]);
```
a-b-r-o-w-n:
```suggestion
    return Object.values(this.conversations);
```

packages/extensions/botState/bf-extension.json
a-b-r-o-w-n: It's magic because it is referenced in more than one place by a string literal. This is a pattern that can lead to hard to track down bugs (maybe a typo somewhere). It's better to rely on the compiler to catch those than humans during code review.

justinwilaby: I was referring to the definition of a magic string according to the article, not the use of the term to incorrectly describe a hard coded string literal. The term is commonly used this way however but it is a misnomer to call hard coded strings "magic strings". :)

packages/app/client/src/data/sagas/chatSagas.ts
justinwilaby: Data transformation is an application concern which belongs in business logic, it is not a view concern. A saga is an appropriate place for this since the role of the view is to render data, not transform it.

justinwilaby: I think moving it to a utility or service would be appropriate. Thoughts?

packages/app/main/src/botFrameworkService.ts
justinwilaby: Added error notification if the port is not available.

packages/app/client/src/data/sagas/chatSagas.ts
a-b-r-o-w-n: The part that is coupled to the view is how you are building up diffs. Adding a `-` or `+` to the key name is coupled with the current presentation of the diff. If we were to change how to visualize the diff, your solution may require changes both here and the extension. However, if the extension received the raw data, it could decide how to present the diff without needing to change the saga. I don't see this as data transformation because it is really preparing the data for the presentation layer.

I think given the time constraint, you probably don't need to change this implementation.

packages/app/client/src/data/sagas/chatSagas.ts
```diff
@@ -73,16 +73,18 @@ const getPreviousBotState = (state: RootState, selectedTrace: Activity): Activit
     }
   );
   const targetLogEntry = filteredLogItems.pop();
-  return targetLogEntry.payload.obj as Activity;
+  return targetLogEntry && (targetLogEntry.payload.obj as Activity);
```
a-b-r-o-w-n: How does this not break the compiler? This is changing the return type to be `undefined | Activity`.

packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.tsx
```diff
@@ -92,6 +97,16 @@ declare type ElectronHTMLWebViewElement = HTMLWebViewElement & {
 };
 
 export class Inspector extends React.Component<InspectorProps, InspectorState> {
+  private static renderAccessoryIcon(icon: string) {
```
a-b-r-o-w-n: It's not clear to me why this needs to be a static method. Can you explain?

packages/extensions/botState/bf-extension.json
```diff
@@ -16,8 +16,33 @@
...
+            "states": {
+              "default": {
+                "label": "STATE JSON"
+              },
+              "selected": {
+                "label": "STATE JSON",
+                "aria-selected": true
+              }
+            }
+          }
```
a-b-r-o-w-n: Any reason that diff couldn't appear here?

packages/extensions/botState/src/BotStateVisualizer.ts
```diff
@@ -54,46 +71,49 @@ export class BotStateVisualizer {
...
+    const div = document.querySelector(this.jsonSelector);
+    div.innerHTML = '';
+
+    if (!this._dataProvider) {
+      return;
+    }
+
+    // Rendering JSON
+    if (this.viewState === ViewState.Json) {
+      div.innerHTML = json2HTML(this._dataProvider);
```
a-b-r-o-w-n: What does this look like in diff mode?

packages/app/client/src/data/sagas/chatSagas.ts
justinwilaby: TypeScript allows returning `null` or `undefined` implicitly for all function signatures. For example:
```js 
const getActivity = ():Activity{
  return undefined;
}
```js
is the same as 
```js 
const getActivity = ():Activity|null|undefined}{
  return undefined;
}
```

packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.tsx
justinwilaby: As a best practice, if your function does not refer to `this` in it's body, it should be static or defined outside the class. The advantage is that the JIT does not need to include the function in the prototype chain which can [improve performance](https://jsperf.com/static-vs-prototype-methods-test) on some browsers. Also, readability is improved since static functions are guaranteed not to have context, much the same way `const` is guaranteed not to change.

packages/extensions/botState/bf-extension.json
justinwilaby: I'm not sure I follow. `bf-extension.json` is essentially a manifest for the extension meant to be consumed by the Emulator.

packages/extensions/botState/src/BotStateVisualizer.ts
justinwilaby: The raw json used to create the diff is seen. This means `+` and `-` are included in key names and all arrays are converted to objects. The current designs do not specify it but it is possible to disable json views when in diff mode.

a-b-r-o-w-n: What does @mewa1024 think?

packages/extensions/botState/bf-extension.json
a-b-r-o-w-n: As I understand this, these add buttons to toggle from graph to tree view? Is there any technical reason why diff could not also be included as an option? Wouldn't that be better than having to right-click on an activity to switch to diff mode?

justinwilaby: To implement this we'd need either a checkbox to toggle between diff and normal data or we'd need `Diff Graph` and `Diff Json` buttons. The current extension protocol supports only buttons ATM. @cwhitten liked my context menu because it scales better. Also, the context menu has the "copy" operations along side the diff.

At this point, I think design changes will introduce too much risk. We're planning to make a second pass after this release to address the little things that aren't straightforward to implement in code.

packages/extensions/botState/src/BotStateVisualizer.ts
justinwilaby: We met about this yesterday. We landed on applying the same color treatment to the JSON diff as we do to the dendrogram.

packages/app/client/src/data/sagas/chatSagas.ts
justinwilaby: Can I resolve this @a-b-r-o-w-n ? 
Essentially adding `null|undefined` is superfluous and unnecessary for TS function signature.

a-b-r-o-w-n: Ah so we don't have `strictNullChecks` enabled? We should consider enabling that compiler option. It will probably not be feasible to do outright, but at least in the meantime we could enforce strict typing in code review.

justinwilaby: Can we do this after our release? I'd like to stay the course and not get sidetracked with nit-picks

packages/app/client/src/commands/emulatorCommands.ts
```diff
@@ -70,15 +72,23 @@ export function registerCommands(commandRegistry: CommandRegistryImpl) {
...
-            conversationId,
-          })
-        );
+        const action = ChatActions.newChat(documentId, 'livechat', {
+          botId: 'bot',
+          endpointId: endpoint.id,
+          endpointUrl: endpoint.endpoint,
+          userId: currentUserId,
+          conversationId,
+          // directLine: createDirectLine({
```
justinwilaby: placeholder to refactor createDirectLine out of `emulator.tsx` and into this command. Code is commented out but works in preliminary tests.

cwhitten:
We're going to merge this and will start working on list of deficiencies introduced with this change before releasing the feature.

Issue #1312, opened 2019-02-12
Modals are very hard to read on high contrast theme


tonyanziano:
## Version
Latest source

## Describe the bug
The over-app modals are all very difficult to read on high contrast theme.

## To Reproduce
1. Set theme to High Contrast
2. Edit an endpoint, service, or create a new bot and look at the modal.

![image](https://user-images.githubusercontent.com/3452012/52661979-1acccb80-2eb8-11e9-8b2c-89979fd2630c.png)

## Expected behavior
The text should be easily readable.

[bug]

tonyanziano:
Addressed in #1402

PR #1402, opened 2019-03-28
[Style] Update modals to increase readability on High-Contrast Theme

Aliandi:
Fixes issue #1312 

### Changes made
- Update `.dialog`  _color_ and _bg-color_ of the high-contrast theme to match the theme's palette.

imagen

imagen

imagen

### DIFF
--- a/packages/app/client/src/ui/styles/themes/high-contrast.css
+++ b/packages/app/client/src/ui/styles/themes/high-contrast.css
@@ -84,7 +84,7 @@ html {
   --input-label-color-disabled: var(--neutral-7);
   --input-border: 1px solid #C8C8C8;
   --input-border-focus: var(--p-button-border-focus);
-  --input-border-error: 1px solid #D02E00;
+  --input-border-error: 1px solid #F38518;
   --input-placeholder-color: var(--neutral-7);
   --input-border-disabled: 1px solid var(--neutral-7);
 
@@ -220,6 +220,8 @@ html {
 .dialog {
   --dialog-bg: var(--neutral-16);
   color: var(--neutral-4) !important;
+  --error-text: #F38518 !important;
+
 }
 
 .dialog .ms-Button-label {
--- a/packages/app/client/src/ui/styles/themes/high-contrast.css
+++ b/packages/app/client/src/ui/styles/themes/high-contrast.css
@@ -218,8 +218,10 @@ html {
 }
 
 .dialog {
-  --dialog-bg: var(--neutral-1);
+  --dialog-bg: var(--neutral-16);
+  color: var(--neutral-4) !important;
 }
+
 .dialog .ms-Button-label {
   color: var(--neutral-15);
 }
--- a/packages/app/client/src/ui/styles/themes/high-contrast.css
+++ b/packages/app/client/src/ui/styles/themes/high-contrast.css
@@ -125,8 +125,10 @@ html {
   --s-button-bg-active: transparent;
   --s-button-bg-disabled: transparent;
 
-  /* Dialog a.k.a. Modal and panels */
+  /* Dialogs & Modals */
   --dialog-bg: var(--neutral-16);
+  --dialog-color: var(--neutral-4);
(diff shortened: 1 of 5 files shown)


packages/app/client/src/ui/styles/themes/high-contrast.css
```diff
@@ -218,8 +218,12 @@ html {
 }
 
 .dialog {
-  --dialog-bg: var(--neutral-1);
+  --dialog-bg: var(--neutral-16);
+  color: var(--neutral-4) !important;
+  --error-text: #F38518 !important;
```
tonyanziano: It would be great if we could avoid using `!important` here. One approach you could try is making the `--error-text` variable [inside of the dialog component's style definition](https://github.com/Microsoft/BotFramework-Emulator/blob/master/packages/sdk/ui-react/src/widget/dialog/dialog.scss#L12) depend on a top-level css variable defined in each of the theme stylesheets.

Ex:

**.../dialog.scss**
```
.dialog {

  ...

  --error-text: var(--dialog-error-color);

   ...
}
```

**themes/high-contrast.css**
```
--dialog-error-color: #F38518;
```

**themes/light.css**
```
--dialog-error-color: <default color here>;
```

etc.

This way, each theme could control the error text without using `!important`.

packages/app/client/src/ui/styles/themes/high-contrast.css
```diff
@@ -218,8 +218,12 @@ html {
...
+
 }
+
 .dialog .ms-Button-label {
   color: var(--neutral-15);
-}
+}
```
tonyanziano: Newline at the end of the file

packages/app/client/src/ui/styles/themes/high-contrast.css
```diff
@@ -218,8 +218,12 @@ html {
 }
 
 .dialog {
-  --dialog-bg: var(--neutral-1);
+  --dialog-bg: var(--neutral-16);
+  color: var(--neutral-4) !important;
```
tonyanziano: Try a similar approach to what I wrote below for the `--error-text` variable. It would be great to have a `--dialog-color` variable or something similar that each theme could control without having to use `!important.

tonyanziano:
Also, it would be great if you could add an entry to [the change log.](https://github.com/Microsoft/BotFramework-Emulator/blob/master/CHANGELOG.md)

Under **Unreleased** and above **v4.3.3** you can add a new section for **Fixed** and put your entry under that.

packages/app/client/src/ui/styles/themes/high-contrast.css
Aliandi: We've been trying this approach with no luck yet, but we'll keep looking into it!

tonyanziano: I've added a commit that should work. Please try it locally and see if that works! 😃

cleemullins:
> Also, it would be great if you could add an entry to [the change log.](https://github.com/Microsoft/BotFramework-Emulator/blob/master/CHANGELOG.md)
> 
> Under **Unreleased** and above **v4.3.3** you can add a new section for **Fixed** and put your entry under that.

Over in the BotBuilder SDK, we have a script that pulls the PR comments for all PRs in a given date range. We use that to create the release notes & changelog at the end of each cycle. You guys are more than welcome to share that!

justinwilaby:
This PR is at odds with the original requirements for modals in the Emulator. Before this is merged, it should go through a design review and get sign off:

image

cleemullins:
@justinwilaby What's the process for a design review?

cwhitten:
@cleemullins it's as simple as assigning a Bot Framework Designer! I'll go ahead and do this. @mewa1024 can you revisit the direction to override the theme variables in all modals? I'm wondering if this break accessibility requirements in High Contrast mode. Justin tells me the VSCode does something similar. This is odd to me.

mewa1024:
Yes, the black background for High Contrast mode is fine.

PR #1407, opened 2019-04-02
Fixed secret prompt dialog's opaque background.

tonyanziano:
The secret prompt dialog had an opaque background which blocked out all of the underlying app.

===

Before:

modal_before

After:

modal_after


justinwilaby:
I believe this was by design. If the botfile is encrypted, the Emulator cannot function until it is decrypted. The `opacity: 1` seems fairly deliberate.

tonyanziano:
Hmm, I'll ask design and see what they say. It seems really weird and jarring to have the entire UI obscured by a dialog when none of the other modals do that.

You aren't able to interact with the underlying UI anyways when a modal is active, or at least you used to not be able to because of the focus trap logic, but I think that regressed when the Fabric to Non-Fabric switch happened.

tonyanziano:
Just checked with @mewa1024 and confirmed that it is supposed to be transparent and follow the same design pattern as the other dialogs.

PR #1411, opened 2019-04-06
Await on bot/user startConversation invocation

cwhitten:


packages/app/main/src/services/luisApiService.ts
```diff
@@ -76,7 +76,7 @@ export class LuisApi {
     // a single array of LuisModel[]
     const luisModels = results
       .filter(result => !('error' in result))
-      .reduce((agg: LuisModel[], models) => (agg.push(...(models as LuisModel[])), agg), []) as LuisModel[];
+      .reduce((agg: LuisModel[], models) => (agg.push(...models as LuisModel[]), agg), []) as LuisModel[];
```
cwhitten: prettier was mad about this, but didn't auto-format it for me.  @justinwilaby you were the last to work here. any qualms?

PR #1439, opened 2019-04-21
Save transcript fix

cwhitten:
As it stands, the transcript path for a conversation opened via url is incorrect when there is an active bot opened via .bot file. With https://github.com/Microsoft/BotFramework-Emulator/pull/1438 it enables this change to support saving transcripts in both .bot file conversations and conversations opened via url.


packages/app/main/src/botHelpers.ts
```diff
@@ -181,3 +183,16 @@ export async function removeBotFromList(botPath: string): Promise<void> {
   const { Commands } = SharedConstants;
   await mainWindow.commandService.remoteCall(Commands.Bot.SyncBotList, bots).catch();
 }
+
+export function getTranscriptsPath(activeBot, conversation): string {
```
tonyanziano: types?

cwhitten:
Closing in lieu of https://github.com/Microsoft/BotFramework-Emulator/pull/1440

PR #1456, opened 2019-04-24
Allow the Emulator user to specify a User ID

### LINKED ISSUE #1344: Allow the Emulator user to specify a User ID to be used for conversations
see https://github.com/Microsoft/BotFramework-Emulator/issues/197 & https://github.com/Microsoft/BotFramework-Emulator/issues/1254

denscollo:
Fixes #1344
### Description
This PR allows the Emulator users to specify a User ID which will be used within conversations.

#### Functionality changes
To implement this feature we made the following changes:

- Add an input to the Emulator Settings so that the user can configure an ID
- Change the startNewConversation method to choose between a randomly generated UserId, the UserId set in the props or the UserId set by the user.
- When the Use custom id checkbox is set to false, the input is disabled and cleaned. When true, it is a required field and it gets enabled.
- The User ID field has a validation so that the user is not able to enter anything other than a GUID.
- Restarting a conversation with the same userId will choose between the one set by the user or the one set by default (if the first one is empty).
- Restarting a conversation with a new userId will generate a random ID the same way it was doing it until now.
- Remove _Restart Conversation_ Split Button
- Add two new buttons _Restart with Same User Id_ and _Restart with New User Id_ to replace _Restart Conversation_ Split Button.
- Add validation to the `Save` button to avoid setting the require user ID flag as true but not setting an ID

#### UI Changes

**Before**
imagen
imagen

**After**
imagen
imagen

### Testing
**Manual testing**
BETranscript-Test

**Automated testing**
Added automated tests for the new settings options.

imagen


packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.tsx
```diff
@@ -90,6 +90,7 @@ export class AppSettingsEditor extends React.Component<AppSettingsEditorProps, A
 
   public componentDidMount(): void {
     this.props.getFrameworkSettings();
+    this.setState({ ['useCustomId']: this.props.framework.useCustomId });
```
justinwilaby: This should probably be placed in the `getDerivedStateFromProps` function

packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.tsx
```diff
@@ -227,6 +251,7 @@ export class AppSettingsEditor extends React.Component<AppSettingsEditorProps, A
     const change = { [name]: checked };
     this.setState(change);
     this.updateDirtyFlag(change);
+    if (name === 'useCustomId' && checked === false) this.setState({ ['userGUID']: '' });
```
justinwilaby: What's the reasoning behind using `this.setState({ ['userGUID']: '' });` versus just `this.setState({ userGUID: '' });` ?

packages/app/client/src/ui/editor/emulator/emulator.tsx
```diff
@@ -143,19 +147,29 @@ export class Emulator extends React.Component<EmulatorProps, {}> {
 
   startNewConversation = async (
     props: EmulatorProps = this.props,
-    requireNewConvoId: boolean = false,
-    requireNewUserId: boolean = false
+    requireNewConvoId?: boolean,
+    requireNewUserId?: boolean
   ): Promise<any> => {
+    if (requireNewUserId == null) {
```
justinwilaby: The previous implementation is more idomatic since `null` is falsy, null checking and setting these to false is superfluous. I would recommend reverting this change.

Aliandi:
Ok @justinwilaby and @tonyanziano ! We are working on the feedback!

tonyanziano:
The text field label looks good when disabled, but the disabled state persists when the text field is enabled. The color of the label should change color depending on the state of the text field.

image

Right now:

disabled -> light gray
enabled -> same light gray

Desired behavior:

disabled -> light gray
enabled -> darker gray

denscollo:
Hi @tonyanziano  We are still working on the Label Color feedback that you gave us and the one about the `getDerivedStateFromProps` function. As for the User Id Label, we are having some issues with the disabled property of the label but we are working on it.

tonyanziano:
Sounds good! @denscollo 👍

packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.tsx
```diff
@@ -90,6 +90,7 @@ export class AppSettingsEditor extends React.Component<AppSettingsEditorProps, A
 
   public componentDidMount(): void {
     this.props.getFrameworkSettings();
+    //   this.setState({ ['useCustomId']: this.props.framework.useCustomId });
```
justinwilaby: This can be removed.

packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.tsx
```diff
@@ -227,6 +249,7 @@ export class AppSettingsEditor extends React.Component<AppSettingsEditorProps, A
     const change = { [name]: checked };
     this.setState(change);
     this.updateDirtyFlag(change);
+    if (name === 'useCustomId' && checked === false) this.setState({ userGUID: '' });
```
justinwilaby: Braces are missing. You can run `npm run lint:fix` to format based on our configuration. It should be:
```js
if (name === 'useCustomId' && checked === false) {
  this.setState({ userGUID: '' });
}
```

Aliandi: Ok! We are fixing this! We've noticed that running `npm run lint:fix` didn't change the format of this line.

tonyanziano:
Nice job, thanks!

Issue #1443, opened 2019-04-22
Remaining sidecar debugging issues


justinwilaby:
Remaining Sidecar debugging issues:

**P0's**
- [x] Label color contrast issues
- [x] Bot State selected color contrast issues
- [x] Graph/Json tab focus indication
- [x] Switching to sidecar mode after failing to open a bot file produces a blank screen
- [x] Slider on edge of webchat can get stuck unless you pull it waaaay over to the right side of the app. (tony) #1460
- [x] Toggling back to Emulator mode from Sidecar has zero UI indication of the mode switch
- [x] "Diff with previous" appears to only utilize the most recent bot state
- [x] Bot State element revisions (Justin) #1458
        Colors: (theme-specific)
          - Non-selected bot states should be grey
          - Selected Bot State should be white
          - On hover of any Bot State should be white
          - In diff mode, the "previous" diffed bot state font should be grey
          - In diff mode, the previous bot state indicator should be red, not blue
- [x] Default to the JSON view instead of the graph view, re-order the tabs in the extension pane (Tony) #1457 
- [x] Welcome Page changes (Chris) #1461
       When in sidecar mode, change the following:
            1. Remove the My Bots box
            2. Remove the "Sign in with Azure controls"
            3. Remove the "create a new bot configuration" text + link
            5. Change Heading copy to say: 

    Bot Framework Emulator
    Bot Inspector <working title>
    Version 4 (BETA)

- [x] Documentation (Justin) #1463 
        - Add a link that on the Welcome page in Sidecar mode that opens the Sidecar Debugging Doc
        - Add a Sidecar Debugging Doc as a tabbed document (Copy incoming) (Justin)
        - Add a Sidecar link in the Help menu that when clicked opens the Sidecar debugging doc in a new tab

When SDK bits are ready:
- [x] Infrastructure changes (Justin) #1465 
       - Remove starting a restify server on a fixed port (9000) when entering Sidecar mode
       - Cleanup additional logic pertaining to this server:port

- [x] Bot Connections (Tony)
        - Connect to a bot in Sidecar mode via url
        - Issue `/INSPECT open` command to bot when connection initiate

- [x] Token storage/rendering (Justin)
        - Receive `/INSPECT connect <guid>` command on response of /DEBUG open command and store it somewhere 
        - Render `/INSPECT connect <token>` command as an Activity in the WebChat control as the first interactable element in the conversation tab. When clicked, the JSON pane exposes the command

[...]

PR #1470, opened 2019-04-29
Implemented '/INSPECT open' command when connecting to a debug bot vi…

tonyanziano:
…a url.

===

One of #1443

### DIFF
--- a/packages/app/client/src/data/sagas/botSagas.ts
+++ b/packages/app/client/src/data/sagas/botSagas.ts
@@ -31,7 +31,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-import { newNotification, UserSettings } from '@bfemulator/app-shared';
+import { DebugMode, newNotification, SharedConstants, UserSettings } from '@bfemulator/app-shared';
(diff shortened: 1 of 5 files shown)


packages/app/client/src/data/sagas/botSagas.ts
```diff
@@ -76,6 +77,30 @@ export function* openBotViaUrl(action: BotAction<Partial<StartConversationParams
...
+          text: '/INSPECT open',
+        };
+        const postActivityResponse = yield call(
+          [CommandServiceImpl, CommandServiceImpl.remoteCall],
+          SharedConstants.Commands.Emulator.PostActivityToConversation,
+          conversationId,
+          activity
+        );
+        if (postActivityResponse.statusCode >= 400) {
+          throw new Error(`An error occurred while POSTing "/INSPECT open" command to conversation ${conversationId}`);
```
justinwilaby: Because saga exceptions aren't the greatest when it comes to capturing, I think we should dispatch a notification with the error message instead of throwing.

tonyanziano: The whole thing is inside of a try-catch that does exactly that!

PR #1477, opened 2019-05-01
Removed offending package lock downgrading npde-fetch to 1.7.3

justinwilaby:
This PR removes the package lock files for qna maker that were causing the build to downgrade `node-fetch` to 1.7.3 which is old and does not follow the DOM standards for `fetch`. 

The [fetchRemote](https://github.com/Microsoft/BotFramework-Emulator/blob/master/packages/app/main/src/commands/electronCommands.ts#L158) command was failing on the nightlies with 'arrayBuffer is not a function` when clearly it is, just not on node-fetch v1.7.3


tonyanziano:
Looks good to me. Going to test out a built version off the branch and then approve once I confirm the fix.

PR #1484, opened 2019-05-01
Fixed LUIS No Models modal issues #1471

### LINKED ISSUE #1471: LUIS No Models modal issues
Have no LUIS models on your account and see the following:

Screen Shot 2019-04-29 at 1 09 24 PM

It should not say {label}, and instead have the value interpolated into the string.

Clicking on "Connect to a LUIS model manually" - nothing happens. Clicking on the link should do something

Clicking on "Learn more about LUIS models" - nothing happens. Clicking on the link should something.

Clicking on "Go to LUIS" - the modal closes. Clicking on the link should actually go to LUIS.

### LINKED ISSUE #1488: Go to LUIS link does nothing
## Version
4.3.57327

## Describe the bug
I tried to add LUIS, and after succesfully logging in I get this message.  Clicking on the GOTO LUIS does nothing.

image

## Expected behavior
* I expect to be able to add luis
* I expect the link to work

[bug]

ivanov-dan:
Fixes #1471, fixes #1488


packages/app/client/src/data/sagas/servicesExplorerSagas.ts
```diff
@@ -209,6 +209,25 @@ function* openConnectedServiceDeepLink(action: ConnectedServiceAction<ConnectedS
   }
 }
 
+function launchExternalLink(action: ConnectedServiceAction<ConnectedServicePayload>): Window {
```
justinwilaby: Since we want to track clicks using Telemetry , all external links need to get routed through the [OpenExternal command](https://github.com/Microsoft/BotFramework-Emulator/blob/master/packages/app/main/src/commands/electronCommands.ts#L184).

The syntax would be:
```js
function* launchExternalLink(action: ConnectedServiceAction<ConnectedServicePayload>):IterableIterator<Any> {
  const serviceType = action.payload.serviceType;  
  switch(serviceType) {
  // switch cases ...
    ... yield call([CommandServiceImpl, CommandServiceImpl.remoteCall], SharedConstants.Commands.Electron.OpenExternal, url);
  }
}
```

tonyanziano: @ivanov-dan once you make this change, I think we are good to take this PR. I tested out all the modals and the problems seem to be fixed. Good job :)

ivanov-dan: I've added some tests.

PR #1503, opened 2019-05-02
Normalized inspector-preload path and removed unneeded code

justinwilaby:
related to #1443 

This is one possible solution - tested locally and as a packaged app.


justinwilaby:
Overlap with another dev. Removing this PR in favor of his work

PR #1505, opened 2019-05-02
Fixed preload script path hack.

tonyanziano:
One of #1443 

===

Now use Electron `app.getAppPath()` API and is fed through the main side to the client on app start.

### DIFF
--- a/packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.tsx
+++ b/packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.tsx
@@ -32,6 +32,8 @@
 //
 
 // Cheating here and pulling in a module from node. Can be easily replaced if we ever move the emulator to the web.
+import { join, normalize } from 'path';
+
 import {
   ExtensionInspector,
   InspectorAccessory,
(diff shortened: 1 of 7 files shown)


packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.tsx
```diff
@@ -297,13 +299,8 @@ export class Inspector extends React.Component<InspectorProps, InspectorState> {
...
   private createWebView(state: InspectorState): ElectronHTMLWebViewElement {
-    // const { cwdAsBase } = this.props;
-    // const preload = `file://${cwdAsBase}/../../../node_modules/@bfemulator/client/public/inspector-preload.js`;
-    // const preload = `file://node_modules/@bfemulator/client/public/inspector-preload.js`;
-    // THIS IS A HACK!! WILL FIX LATER
-    const preload = state.inspector.src
-      .replace(/extension-.*/, 'client/public/inspector-preload.js')
-      .replace('asar.unpacked', 'asar');
+    const appPath = normalize(this.props.appPath);
+    const preload = normalize(join('file://', appPath, '/node_modules/@bfemulator/client/public/inspector-preload.js'));
```
a-b-r-o-w-n: I'm curious if `require.resolve` will work here instead?

a-b-r-o-w-n: I tried it and it doesn't look like it would work. :(

tonyanziano: Would there be a benefit to that approach?

PR #1510, opened 2019-05-02
Implemented autocomplete functionality in open bot dialog

tonyanziano:
One of #1443 

===

Don't worry, I'll squash all of the commits.

image

image

image

@mewa1024 I just used the styling from the Split Button drop down specs

### DIFF
--- a/packages/app/main/src/settingsData/reducers/savedBotUrlsReducer.ts
+++ b/packages/app/main/src/settingsData/reducers/savedBotUrlsReducer.ts
@@ -31,15 +31,21 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-export * from './autoComplete/autoComplete';
-export * from './largeHeader/largeHeader';
-export * from './mediumHeader/mediumHeader';
-export * from './smallHeader/smallHeader';
-export * from './dialog';
-export * from './button/primaryButton';
-export * from './button/defaultButton';
-export * from './textField/textField';
-export * from './checkbox/checkbox';
-export * from './insetShadow/insetShadow';
-export * from './spinner/spinner';
-export * from './splitButton';
+import { ADD_SAVED_BOT_URL, SavedBotUrlsAction, SavedBotUrlsActionPayload } from '../actions/savedBotUrlsActions';
+
+export function savedBotUrlsReducer(
+  state: string[] = [],
+  action: SavedBotUrlsAction<SavedBotUrlsActionPayload>
+): string[] {
+  switch (action.type) {
+    case ADD_SAVED_BOT_URL:
+      if (!state.some(botUrl => botUrl === action.payload)) {
+        state.push(action.payload);
+      }
+      break;
+
+    default:
+      break;
+  }
+  return state;
+}
--- a/packages/sdk/ui-react/src/widget/autoComplete/autoComplete.tsx
+++ b/packages/sdk/ui-react/src/widget/autoComplete/autoComplete.tsx
@@ -1,0 +1,318 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license.
+//
+// Microsoft Bot Framework: http://botframework.com
+//
+// Bot Framework Emulator Github:
+// https://github.com/Microsoft/BotFramwork-Emulator
+//
+// Copyright (c) Microsoft Corporation
+// All rights reserved.
+//
+// MIT License:
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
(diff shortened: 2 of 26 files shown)


packages/sdk/ui-react/src/widget/autoComplete/autoComplete.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+  // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  // IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+  // FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+  // COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+  // IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+  // CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+  // from https://github.com/bevacqua/fuzzysearch
+  private fuzzysearch(needle: string, haystack: string) {
```
tonyanziano: Don't know if the above copyright and link to source will be compliant, but we can discuss it.

packages/app/main/src/settingsData/reducers/savedBotUrlsReducer.ts
```diff
@@ -0,0 +1,51 @@
...
+//
+
+import { ADD_SAVED_BOT_URL, SavedBotUrlsAction, SavedBotUrlsActionPayload } from '../actions/savedBotUrlsActions';
+
+export function savedBotUrlsReducer(
+  state: string[] = [],
+  action: SavedBotUrlsAction<SavedBotUrlsActionPayload>
+): string[] {
+  switch (action.type) {
+    case ADD_SAVED_BOT_URL:
```
justinwilaby: Single case switch statement could be just an if statement. It doesn't seem likely this will grow beyond a single action.

packages/sdk/ui-react/src/widget/autoComplete/autoComplete.tsx
```diff
@@ -0,0 +1,318 @@
...
+  private get listboxId(): string {
+    return `auto-complete-listbox-${this.state.id}`;
+  }
+
+  private get textboxId(): string {
+    // textbox id is only necessary if we have a label
+    return this.props.label ? `auto-complete-textbox-${this.state.id}` : undefined;
+  }
+
+  private get filteredItems(): string[] {
```
justinwilaby: I would expect the matches to have some sort of emphasis on the searched value. This can be achieved through using a `<strong>` tag for the matched string:

```js
...
return filteredItems.map(item => item.replace(new RegExp(this.value, 'gi'), `<strong>${this.value}</strong>`);
```
Then use `dangerouslySetInnerHtml` on the `<li>` tag

tonyanziano: That would be a nice visual improvement but it might not work so well with this search algorithm because it just checks for the succession of characters in string and returns true if it exists.

For example:

**ir** would match on **i**nte**r**national because there is an i in the string, followed by an r. So the meaningful way to emphasize that would probably be to bold everything between the first "i" and the last "r" so you would be left with **inter**national.

The method you suggested would provide some strange results in some words:

`er` matches on `Herbert` so the end result would be H**er**b**er**t

`o` matches on `localhost:3948/api/orders` so the end result would be l**o**calh**o**st:3948/api/**o**rders with 3 bolded `o`'s

In my opinion, bolding every character between the first and last matched characters would be the way to go, but that seems like a QoL improvement that could be added in a later PR and not critical to this first implementation.

packages/app/main/src/settingsData/reducers/savedBotUrlsReducer.ts
tonyanziano: I considered the possibility of us adding a way to wipe the list of URLs  -- or a particular URL -- in the future if it gets too large.

packages/sdk/ui-react/src/widget/autoComplete/autoComplete.tsx
tonyanziano: Opened an issue to track this work: #1518

justinwilaby: I see. I did't look closely at the fuzzy search function to see this is what it's doing. Maybe an alternative would be to implement the emphasis for sub strings that are exact sequential matches and then sort them so they are first in the list. 

Thoughts?

Also, although very long lists are unlikely, have we considered the performance implications of the loop nesting in the search algo for long lists? How would a user remove items that may have gotten in there from a typo or that are no longer needed?

tonyanziano: That strategy for highlighting sounds good to me!

As for the performance with long lists, there are two trivial solutions I can think of:
 - put a limit on the number of saved items we feed into the component (easy to manage in the context of saved bot endpoints, but would have to be managed for each autocomplete placed throughout the app)
 - swap out the current search algorithm for a more performant one (very easy fix)

The list of searchable items is provided by the parent of the widget, so any item removal / clear functionality would have to be implemented by the parent. It would be very straight forward in this case and we would just need to add a new action handler to the `savedBotUrls` reducer.

tonyanziano: Opened an issue to track work for list item removal / clearing UI: #1525

PR #1541, opened 2019-05-08
Add PublishToCoveralls.ps1

BruceHaley:
For publishing the coveralls report to the coveralls site. This to be run in a release pipeline like Upload-Coveralls-Results-from-DotNet-CI https://fuselabs.visualstudio.com/SDK_v4/_releaseDefinition?definitionId=20&_a=environments-editor-preview

### DIFF
--- a/scripts/PublishToCoveralls.ps1
+++ b/scripts/PublishToCoveralls.ps1
@@ -1,0 +1,55 @@
+#
+# Installs and runs Coveralls.exe to upload coverage files to https://coveralls.io/.
+# Arguments example: -coverallsToken $(Coveralls.Token) -pathToCoverageFiles $(Build.SourcesDirectory)\CodeCoverage
+#
+Param(
+    [string]$coverallsToken,
+    [string]$pathToCoverageFiles
+)
+
+Write-Host Install tools
+$basePath = (get-item $pathToCoverageFiles ).parent.FullName
+$coverageAnalyzer = "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Team Tools\Dynamic Code Coverage Tools\CodeCoverage.exe"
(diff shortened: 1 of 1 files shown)


justinwilaby:
@BruceHaley - 

I'm not sure why this is required. The first build process was one I created and was able to upload coverage data to coveralls using these [env variables](https://github.com/microsoft/BotFramework-Emulator/blob/master/env.js) and the [coveralls package](https://www.npmjs.com/package/coveralls)

The test script then became `jest --coverage --coverageReporters=text-lcov | node -r env.js node_modules/coveralls/bin/coveralls`

Does this not work anymore?

BruceHaley:
@justinwilaby , the object here is to publish Coveralls results from a release pipeline. Publishing from a CI build where the source repo is visible to the public and the CI build can be triggered by an external PR would allow a test to be submitted that exposes the Coveralls token, set as an environment variable. See this discussion for more: https://github.com/lemurheavy/coveralls-public/issues/1272#issuecomment-482732330

justinwilaby:
That makes sense...carry on... :)

tonyanziano:
Just curious, is there any reason why this can't just be a PowerShell task inside of VSO as part of the release? Or is the release flow different from a normal pipeline?

Just wondering if we need to actually add a script file to the source just to be used by the build machine.

BruceHaley:
@tonyanziano , I think you are asking "why not put the powershell script inline instead of having the Powershell task get it from a file"? When the script is simple and fewer than about 20 lines, I generally put it inline. When it is longer, I generally execute it from a .ps1 file. Many of our builds run scripts stored in a source repo. The folder is often named "build". (A copy of this script is already stored in the SDK_v4 repo at https://github.com/microsoft/botbuilder-dotnet/tree/master/build .)

PR #1543, opened 2019-05-08
Jwilaby/#1531 focus issues

### LINKED ISSUE #1531: Open bot modal loses focus on url autocomplete selection
## Version
4.4.1

## Describe the bug
Selecting from from the list sends focus behind the modal. 

## To Reproduce
1. Go to Open bot modal
2. Select a url from the autocomplete list
3. Hit "Enter", nothing happens.
4. Tabbing around shows focus behind the modal

## Expected behavior
Upon selection, focus should be on the url input element

[bug]

justinwilaby:
Fixes #1531 

Fix provided by focusing the next input field in the form when a selection is made from the autocomplete widget.

### DIFF
--- a/packages/sdk/ui-react/src/widget/autoComplete/autoComplete.tsx
+++ b/packages/sdk/ui-react/src/widget/autoComplete/autoComplete.tsx
@@ -44,6 +44,7 @@ export interface AutoCompleteProps {
   items?: string[];
   label?: string;
   onChange?: (updatedValue: string) => void;
+  onSelectionComplete?: (value: string) => void;
   placeholder?: string;
   value?: string; // use only for a "controlled input" experience
 }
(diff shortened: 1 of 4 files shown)


packages/sdk/ui-react/src/widget/autoComplete/autoComplete.tsx
```diff
@@ -182,10 +183,15 @@ export class AutoComplete extends Component<AutoCompleteProps, AutoCompleteState
...
+  private onListItemMouseDown = (result: string) => () => {
     if (this.props.onChange) {
       this.props.onChange(result);
     }
+
+    if (this.props.onSelectionComplete) {
+      this.props.onSelectionComplete(result);
+    }
+
     this.setState({ currentInput: result });
```
tonyanziano: my suggestion would be to initialize a ref to the `<input>` element in the component, and then at the end of this function, call your `RAF(() => inputRef.focus())` method

tonyanziano: the "ENTER" key handler auto-focuses the input because focus never actually leaves the input during keyboard navigation, because ["focus" is controlled via aria-selected & aria-activedescendant](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_focus_activedescendant)

PR #1558, opened 2019-05-10
 [ISSUE#1534] Fix Removing an overridden User ID doesn't remove the overridden User ID in conversations

### LINKED ISSUE #1534: Removing an overridden User ID doesn't remove the overridden User ID in conversations
## Version
4.4.1

## Describe the bug
If I remove a User ID override in the settings page and save the settings, the User ID does not get cleared out.

## Expected behavior
It should clear the User ID appropriately

[bug]

Aliandi:
Fixes #1534 

### Description
This PR fixes the issue when removing an overridden User ID doesn't remove the overridden User ID in conversations

### Changes made
In order to handle this as clean as possible, we added a basic configuration file for all the conversations in which there's an stable `userId`, that'll be used when there's not `userId ` set from the Settings Page. 
By doing this, there's an inmutable set of data and mantains consistency through sessions.

#### Functionality changes
To implement this feature we made the following changes:

- Add new file `conversationDefaults` that has the `userId` and `documentId` stable values.
- Update `appSettingsEditor.tsx` file to set the userGUID as empty instead of null when the UserID from the Settings' Page is disabled.
- Update `emulator.tsx` file to import `conversationDefaults` values and use it instead of props.document.userID that is not being overriden on `stableId` variable declaration.


Aliandi:
Hey guys! I am pushing the fix of the button right now, but we are still trying to reproduce the ID issue, we will push that asap.

Aliandi:
@tonyanziano we are not being able to repro the issue, is what we are showing in this gif ok? Are we missing anything?
BFETestUserID

tonyanziano:
After, doing a wipe & a clean build of everything, I can no longer repro the issue using the steps I provided. 

However, I was able to produce some incorrect behavior by starting a new live chat session very quickly after submitting my custom user id.

I'm not able to produce the same behavior on the most recent nightly.

denscollo:
Hi @tonyanziano , after running a few tests, we were able to reproduce this behavior that you are mentioning. We believe that it's not caused by the userid itself but by the restart conversation functionality. If it's ok, we'll create a new separate issue to solve it.

tonyanziano:
@denscollo That sounds reasonable to me!

PR #1574, opened 2019-05-14
Added 'Clear State' menu item.

ivanov-dan:
Do **NOT** approve yet:
This will need proper message for the user when they are prompted to confirm 'Clear State'.


cwhitten:
@mewa1024 can you give copy guidance on what to render in a confirmation control when a user selects this item from a menu option?

Perhaps something like: 

`This will log you out of Azure (if logged-in) and remove any session based data. Continue?`

packages/app/main/src/commands/emulatorCommands.spec.ts
```diff
@@ -561,4 +561,9 @@ describe('The emulatorCommands', () => {
     expect(result.id).toBe('someId');
     expect(postActivitySpy).toHaveBeenCalled();
   });
+
+  it('should clear state', async () => {
+    const result = await mockCommandRegistry.getCommand(SharedConstants.Commands.Emulator.ClearState).handler();
+    expect(result).toBe(true);
```
tonyanziano: The test here could be a little more comprehensive instead of only testing for the result to be true. Usually, tests in the Emulator try to check to make sure that the command service was invoked with the proper commands. This can be done by mocking `mainWindow.commandService`, calling the command handler (for ClearState in your case), then checking to make sure the proper commands were called after the handler returns.

An example of accomplishing this can [be found here](https://github.com/microsoft/BotFramework-Emulator/blob/master/packages/app/client/src/commands/emulatorCommands.spec.ts#L168) or [here.](https://github.com/microsoft/BotFramework-Emulator/blob/master/packages/app/main/src/protocolHandler.spec.ts#L416)

Pseudo-code example for your case:

```
it(should clear state, () => {
  
  const remoteCallSpy = jest.spyOn(mainWindow.commandService, 'remoteCall').mockResolvedValue(true);
  const handler = registry.getCommand(ClearState);
  const result = await handler();
 
  // check result
  expect(result).toBe(true);

  // check to see if correct commands were invoked
  expect(remoteCallSpy).toHaveBeenCalledWith(ShowMessageBox, true, { ...message box options });
  ...
  ...


```

This test should also validate the conditional that changes the prompt dialog message.

ivanov-dan: Take another look.

justinwilaby: @ivanov-dan - I'll walk you through this since it involves a slightly different path after the tech debt branch is merged in. You and I will need a good hour or so.

cwhitten:
@ivanov-dan are you opening a new pull-request with these changes?

ivanov-dan:
I created a new one which was approved and merged.

PR #1577, opened 2019-05-15
Emulator Launch, Commands and Extension Tech Debt Buydown

justinwilaby:
This WIP addresses provides some much needed buydown of tech debt regarding how the Emulator is launched, the extensions infrastructure and the command system.

This PR provides:
1. Decorators for registering a command and a DI container for the command service. e.g.
```js
class MyClass {
 @CommandServiceInstance()
 private commandService: CommandServiceImpl;

 @Command('commandName')
 protected myCommand(args) {
    // Do stuff
  }
}
```
2. Removal of the ts-loader from webpack which halves the build times
3. Targeting ES6+ in the build and using terser for minification
4. Bumped `botframework-config` out of `preview` and into `v4.4`


packages/app/client/src/data/sagas/azureAuthSaga.ts
```diff
@@ -46,31 +46,40 @@ import { RootState } from '../store';
 
 const getArmTokenFromState = (state: RootState) => state.azureAuth;
 
-export function* getArmToken(action: AzureAuthAction<AzureAuthWorkflow>): IterableIterator<any> {
-  let azureAuth: AzureAuthState = yield select(getArmTokenFromState);
-  if (azureAuth.access_token) {
+export class AzureAuthSaga {
+  @CommandServiceInstance()
+  private static commandService: CommandServiceImpl;
```
tonyanziano: So in the case where we don't have a class (Sagas for example), the pattern for importing the command service is to create a class and inject it into a member variable via the decorator?

packages/app/client/src/index.tsx
```diff
@@ -34,44 +34,23 @@
...
-registerAllCommands(CommandRegistry);
 
 // Start rendering the UI
 ReactDOM.render(
   React.createElement(Provider, { store }, React.createElement(Main as any)),
   document.getElementById('root')
 );
 
-// Tell the main process we're loaded
-CommandServiceImpl.remoteCall(SharedConstants.Commands.ClientInit.Loaded)
```
tonyanziano: where does this code live now?

packages/app/client/src/ui/shell/explorer/botNotOpenExplorer/botNotOpenExplorerContainer.ts
```diff
@@ -31,25 +31,30 @@
...
 });
 const mapDispatchToProps = (dispatch: (action: Action) => void): BotNotOpenExplorerProps => ({
-  openBotFile: () => ActiveBotHelper.confirmAndOpenBotFromFile(),
-  sendNotification: (error: Error) =>
-    dispatch(beginAdd(newNotification(`An Error occurred on the Bot Not Open Explorer: ${error}`))),
+  openBotFile: async () => {
+    try {
+      await ActiveBotHelper.confirmAndOpenBotFromFile();
+    } catch (e) {
+      dispatch(beginAdd(newNotification('An Error occurred on the Bot Not Open Explorer: oh noes!')));
```
tonyanziano: probably shouldn't have `oh noes!` as the error message 😉

packages/app/client/src/ui/shell/mainContainer.ts
```diff
@@ -52,6 +56,16 @@ const mapDispatchToProps = (dispatch): MainProps => ({
       dispatch(PresentationActions.disable());
     }
   },
+  applicationMountComplete: async () => {
+    await new Promise(resolve => {
```
tonyanziano: disregard previous comment. found it! :)

packages/app/client/src/data/sagas/azureAuthSaga.ts
justinwilaby: Yes. Since decorators require a descriptor and descriptors are retrieved from objects, a class is required but can be all static. In theory, this could work on plain objects but classes seem more idiomatic.

tonyanziano: gotcha

packages/sdk/shared/src/command/service.ts
```diff
@@ -33,104 +33,96 @@
...
+  send(channel: string, ...args: any[]): void;
 }
 
-export class CommandServiceImpl extends DisposableImpl implements CommandService {
-  private readonly _channel: Channel;
-  private readonly _registry: CommandRegistry;
-  private readonly _channelName: string;
-  private readonly _ipc: IPC;
-  private _notFoundHandler: (commandName: string, ...args: any[]) => any;
+export class CommandServiceImpl {
```
tonyanziano: It would give me a lot more confidence if there were tests for these changes

justinwilaby: Coverage is at 100% for those files since they are used in over 400 tests.

cwhitten:
@justinwilaby have we gone through the 4.5 release checklist on this change?

packages/app/main/src/appUpdater.ts
```diff
@@ -191,6 +165,140 @@ class EmulatorUpdater extends EventEmitter {
...
+            ShowUpdateAvailableDialog,
+            updateInfo.version
+          );
+          if (result) {
+            // show but don't block on result of progress indicator dialog
+            await this.commandService.remoteCall(UpdateProgressIndicator, {
+              label: 'Downloading...',
+              progress: 0,
+            });
+            await this.commandService.remoteCall(ShowProgressIndicator);
```
tonyanziano: This `await` needs to be removed or else the progress bar will be shown, but it will hang since this is waiting for the dialog to be dismissed until executing the next line which downloads the update. This defeats the purpose of showing the progress bar.

This is why it [wasn't awaited in the previous implementation.](https://github.com/microsoft/BotFramework-Emulator/blob/master/packages/app/main/src/main.ts#L101)

tonyanziano:
:shipit: :shipit: :shipit: :shipit: :shipit: :shipit:

PR #1582, opened 2019-05-15
Fixed the spinner issue when publishing LUIS after training

### LINKED ISSUE #1572: Publish to LUIS UI hangs with a loading spinner
from @gabog 

## Version
4.4.1

## Describe the bug

Training works fine in the extension when I add a new intent but publish stays spinning after a click it. The model does publish but the UI never turns off the spinning animation.

## To Reproduce
1. Get the LUIS extension rendered on a LUIS trace
2. Train and publish to the model
4. See scrolling

## Expected behavior
Upon success, the spinning animation should go away.

## Screenshots
thumbnail_image001

[bug]

ivanov-dan:
fixes #1572

### DIFF
--- a/packages/extensions/luis/client/src/App.tsx
+++ b/packages/extensions/luis/client/src/App.tsx
@@ -59,7 +59,7 @@ import { LuisTraceInfo } from './Models/LuisTraceInfo';
 const $host: InspectorHost = (window as any).host;
 const LuisApiBasePath = 'https://westus.api.cognitive.microsoft.com/luis/api/v2.0';
 const TrainAccessoryId = 'train';
-const PublichAccessoryId = 'publish';
+const PublishAccessoryId = 'publish';
 const AccessoryDefaultState = 'default';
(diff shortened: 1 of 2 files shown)


packages/extensions/luis/client/src/App.tsx
```diff
@@ -321,7 +321,7 @@ export class App extends Component<any, AppState> {
       $host.logger.error(err.message);
       $host.trackEvent('luis_publishFailure', { error: err.message });
     } finally {
-      $host.setAccessoryState(TrainAccessoryId, AccessoryDefaultState);
+      $host.setAccessoryState(PublishAccessoryId, AccessoryDefaultState);
```
justinwilaby: So this was the crux of the fix?

tonyanziano: Looks like it was resetting the state of the train button instead of the publish button?

ivanov-dan: Yes, that was it.

PR #1593, opened 2019-05-28
Jwilaby/update extension deps

justinwilaby:
This PR updates the extensions to use `botframework-schema` for data models


justinwilaby:
Package-lock files updated!

PR #1595, opened 2019-05-28
Side-by-side conversation modes

### LINKED ISSUE #1564: [Bot Inspector] Side-by-side conversation modes
### Side-by-side conversation modes 

Currently, one must globally toggle into Bot Inspector mode to utilize the feature. Because we are introducing state inspection, state diffing and potentially other valuable Bot related data we should allow users to have the conversation with the Bot in the WebChat/Direct Line channel embedded with the Emulator and establish an additional connection to the same Bot in Inspector Mode, moving between the conversation tabs as a core part of the workflow. Today this would require running an additional Emulator instance, which is heavy and unnecessary. 

*High-level code changes*

- Some components of the Emulator render based on a global DebugMode flag. Instead of being driven by this global value in a menu item, it would be driven by the active conversation tab (Emulator mode or Inspector mode)

- Via the Open Bot modal or some other means, when we make a connection to a bot we need to know if the user intends to connect and have the conversation or connect and inspect a conversation. This could be done by adding a checkbox in the Open Bot modal or connecting to a Bot in inspector mode via a menu item. Think Debug -> Start Debugging in VS Code. 

- writing DebugMode to `server.json` is no longer necessary.

justinwilaby:
Resolves #1564 

This PR provides:
1. Removal of an explicit global debug mode
2. A "Start debugging" menu option and related checkbox in the open bot dialog
3. The ability to connect to a bot and debug the same bot without multiple instances of the Emulator
4. **NOTE** contextual service pane switching is omitted since this would create too large of a PR

Screen Shot 2019-05-29 at 11 53 57 AM

Screen Shot 2019-05-29 at 11 55 30 AM

Screen Shot 2019-05-29 at 11 56 33 AM

Screen Shot 2019-05-29 at 11 56 41 AM


packages/app/client/src/data/reducer/editor.ts
```diff
@@ -32,40 +32,60 @@
...
+}
+
+export interface DocumentUI {
+  horizontalSplitter: SplitterSize[];
+  verticalSplitter: SplitterSize[];
+}
+
+export interface SplitterSize {
+  absolute: boolean;
+  percentage: number;
```
tonyanziano: bonus points for all this typing!!! 👍 😃

packages/app/client/src/ui/editor/welcomePage/welcomePage.tsx
```diff
@@ -87,19 +84,6 @@ export class WelcomePage extends React.Component<WelcomePageProps, {}> {
     return (
       <React.Fragment>
         <LargeHeader>Bot Framework Emulator</LargeHeader>
-        {this.props.debugMode === DebugMode.Sidecar && (
-          <div className={styles.botInspectorHeaderContainer}>
-            <MediumHeader>
-              Bot Inspector <span className={styles.betaLabel}>(BETA)</span>
```
tonyanziano: remove `bot-inspector-header-container` and `beta-label` styles from `welcomePage.scss`

tonyanziano: Also just noticed that the "Save Transcript" button is missing in debug mode:

image

justinwilaby: Well yes. Mainly because the round trip is broken. If the user exports a transcript in debug mode and then re-imports it, nothing is shown in webchat because there is no way to set the mode

tonyanziano: Ahhh, I see. How did we do it before? How did we select the right mode when opening both types of transcripts?

justinwilaby: It was a manual process. You could use `file->sidecar debug mode` then import the transcript. Now there's just no way to get into that mode and still import a transcript from a file so I removed the export option until we come up with another way.

PR #1596, opened 2019-05-28
V dai/clear state

ivanov-dan:


packages/app/main/src/commands/emulatorCommands.ts
```diff
@@ -277,4 +277,33 @@ export class EmulatorCommands {
...
+      message: signedInMessage,
+      type: 'question',
+    });
+
+    if (bClearState === 1) {
+      await this.commandService.call(SharedConstants.Commands.Electron.ShowMessageBox, false, {
+        message: 'You have successfully cleared state.',
+        title: 'Success!',
+      });
+      const { session } = require('electron');
```
justinwilaby: Although this is generally ok, we have established a pattern of placing requires at the top of the file using `import`. In this case it would be `import * as electron from 'electron';` or `import { session } from 'electron';`

packages/app/main/src/commands/emulatorCommands.ts
```diff
@@ -277,4 +277,33 @@ export class EmulatorCommands {
...
+      await new Promise(resolve => session.defaultSession.clearStorageData({}, resolve));
```
justinwilaby: This should probably go in at line 299.

packages/app/main/src/commands/emulatorCommands.ts
ivanov-dan: Fixed.

packages/app/main/src/commands/emulatorCommands.ts
justinwilaby: Is there a technical reason the user is notified before the storage data is cleared?

ivanov-dan: This is a scenario that @cwhitten  requested.

PR #1618, opened 2019-06-05
Secrets are now managed by native OS credential store.

### LINKED ISSUE #1464: [Telemetry Compliance] Security: Ensure any secrets are stored in compliant crypto stores (or not cached)
## Ensure compliant handling of secrets
Ensure any secrets handled by the emulator are stored in compliant crypto stores or not cached at all for security compliance.

### Relevance
* encryption/decryption of .bot files
* encryption/decryption of app-id/password

tonyanziano:
Resolves #1464 

---

### Please try to break this :)

Bot secrets are now stored by your OS in your native credential store:

  Windows: Credential Manager
  Mac: Keychain
  Linux: Secret Service API / libsecret

image

image

**NOTE #1:** Due to the fact that we now use `keytar` which is a [native node addon](https://nodejs.org/api/addons.html), it has to be built properly for Electron when launching and running the app, and must be built for Node when running the tests (Jest environment). 

I couldn't figure out a way to have it easily built for both environments, so if you decide to flip back and forth between running tests and launching the app, then you will need to use the scripts in the `root/package.json` file:

  - `rebuild:keytar:electron`: use when you want to run the app
  - `rebuild:keytar:node`: use when you want to run the tests

Let me know if you think this process could be improved or better documented.

---

**NOTE #2:** The library used to rebuild `keytar` for Electron, `electron-rebuild`, does not play nicely with Lerna monorepos, and so `keytar` had to be added as a dependency to the root `package.json` file. This should not have any adverse effects other than there being an extra line in the file. I opened an issue against `electron-rebuild` and it can be found [here](https://github.com/electron/electron-rebuild/issues/302).


justinwilaby:
This looks good. I think we should consider code that removes the store bot secrets from the `server.json` with this rollout.

@cwhitten - thoughts?

tonyanziano:
@justinwilaby bot secrets are currently stored in `bots.json` via the `BotInfo` type, unless you are talking about something else. This PR has code that removes that property from those entries.

cwhitten:
I think we need to do that before we can submit for the security & privacy reviews. @tonyanziano what do you think?

tonyanziano:
I thought that removing the property would be sufficient, and that further writes to `bots.json` would then omit the property, but that doesn't seem to be the case after testing it just now.

I can add some code that purges any secrets from `bots.json` before loading / writing.

packages/app/main/src/utils/getBotsFromDisk.ts
```diff
@@ -45,7 +45,10 @@ export const getBotsFromDisk = (): BotInfo[] => {
     const botsJsonContents = readFileSync(botsJsonPath);
     const botsJson = botsJsonContents ? JSON.parse(botsJsonContents) : null;
     if (botsJson && botsJson.bots && Array.isArray(botsJson.bots)) {
-      return botsJson.bots;
+      return botsJson.bots.map(bot => {
+        delete bot.secret;
+        return bot;
```
tonyanziano: this code purges the `secret` property from any `BotInfo` objects when we first load bots from disk

PR #1619, opened 2019-06-06
Collapsible JSON Viewer 

### LINKED ISSUE #1566: [Bot Inspector] Collapsible JSON
### Collapsible JSON viewer 

Going into the Beta of the feature, we anticipated that emitted bot states could be at times hard to read given deeply nested JSON objects. Prototypes created to soften this issue were around new data visualizations (trees, graphs) but internal feedback was suggestive that more thought was needed if it was the right thing to do. Users are all suggesting the same thing: Collapsible JSON. Rather than creating an entirely new visualization we can augment our existing JSON view to be collapsible, and potentially provide this view across all JSON payloads. 

*High-level code changes*

- Refactor a significant portion of the existing JSON viewer to be collapsible by integrating with or heavily borrowing from a library like [react-json-view](https://github.com/mac-s-g/react-json-view).

- Understand existing WebView caching strategies to avoid performance overhead, include benchmarks against existing load times
- Consolidate debug extension & all up json view [P1]
- Consider using React Hooks & a css-in-js library like [emotion](https://emotion.sh/docs/introduction)
- Theming

justinwilaby:
## Contents
The PR promotes the `json` extension to a React project and installs the [react-json-tree](https://github.com/reduxjs/redux-devtools/tree/master/packages/react-json-tree) as the default viewer and resolves #1566 based on the acceptance criteria listed within it. 
**NOTE** This PR does not attempt to combine the functionality of the Bot State viewer since it would present too large of a work unit. This work will be done in a separate PR.

## Themes
Themes are supported using the built in [base16 theming](https://github.com/chriskempson/base16-textmate) and have small differences in color treatment as a result of the way this theming system groups the colorization of data types.

Light
Screen Shot 2019-06-13 at 9 14 40 AM
Dark
Screen Shot 2019-06-13 at 9 14 26 AM
High Contrast
Screen Shot 2019-06-13 at 9 14 13 AM

## A11y
a11y, including keyboard navigation, is implemented manually via mutation observers since there does not appear to be out of the box support with the component dependency.

For Reference: [this DOM structure](http://oaa-accessibility.org/examplep/treeview1/) was duplicated for this control.


packages/extensions/json/src/jsonViewer.tsx
```diff
@@ -0,0 +1,255 @@
...
+  'high-contrast': highContrast,
+};
+
+export interface JsonViewerState {
+  data: any;
+  themeName: string;
+}
+
+export class JsonViewer extends Component<{}, JsonViewerState> {
+  private readonly mutationObserver: MutationObserver;
```
tonyanziano: Is there a reason the implementation for `MutationObserver` is left out? Is this still in the works?

justinwilaby: In this case, the implementation is found in the constructor.

tonyanziano: I see that you are `new`ing a `MutationObserver` in the constructor, but where is the actual definition for the `MutationObserver` class? Is that still WIP / conceptional?

a-b-r-o-w-n: MutationObserver is a web api: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

tonyanziano: Oh, cool! I had no idea. Thanks.

justinwilaby:
Although I am indifferent to expanding all nodes by default. Most tools showing a hierarchy do not expand all nodes by default. In the spirit of improving UX with the introduction of this control, we may not have enough data to make this decision.

I can say that the authors of the widget chose not to. If I were to speculate, this could be because most users prefer to see only the immediate child nodes and a sensible default configuration took this into consideration.

tonyanziano:
This is purely anecdotal, but all the JSON viewers I've used tend to initially show all nodes collapsed by default.

---
First 3 results from google using the followng JSON:

```
{
  "outer1": {
    "inner1": {
      
    }
  },
  "outer2": {
    "inner2": {
      "inner-inner1": {
        
      }
    }
  }
}
```

image

image

image

tonyanziano:
Should the focus border color in High Contrast be orange to match the rest of the app?

image

image

cwhitten:
having the children initially collapsed sounds reasonable. if we get feedback we can always revisit.

justinwilaby:
Theme names were coming in as sentence case but the css was expecting lowercase. I made the correction.

Thanks @tonyanziano for the good catch!

PR #1624, opened 2019-06-11
Added first-time telemetry modal

### LINKED ISSUE #1236: [Telemetry Compliance] First-time splash page / modal for Telemetry. Docs page as well.
We need a first-time splash page or modal telling the user about how our telemetry works and what we are collecting, as well as how to opt out of the data collection.

We also need a docs page that we can link to; the settings page will have the same link. Currently the link points to [https://aka.ms/bot-framework-emulator-data-collection]().

Here is what the app settings UI looks like:

image

tonyanziano:
Resolves #1236 

===

This PR shows a dialog on first-time startup that will tell the user about our data collection policy and allows them to opt in or out.

**Updated dialog:**

image


packages/app/client/src/commands/uiCommands.ts
```diff
@@ -245,4 +246,11 @@ export class UiCommands {
   protected openBotViaURL(conversationParams: StartConversationParams): void {
     store.dispatch(openBotViaUrlAction(conversationParams));
   }
+
+  // ---------------------------------------------------------------------------
+  // Shows the data collection dialog
+  @Command(UI.ShowDataCollectionDialog)
+  protected async showDataCollectionDialog() {
+    return await DialogService.showDialog(DataCollectionDialogContainer);
```
justinwilaby: Consider removing `async` and `await` here since it's not really needed. Keeping it would be similar to the following:

```js
protected showDataCollectionDialog(): Promise<any> {
  return new Promise(resolve =>  {
    DialogService.showDialog(DataCollectionDialogContainer)
    .then(response => resolve(response); // <-- using the resolved value from one promise as the resolved value to another. Instead, skip this intermediary promise and return the original.
  }
}
```

packages/app/client/src/ui/dialogs/dataCollectionDialog/dataCollectionDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Component, ReactNode } from 'react';
+
+import * as styles from '../dialogStyles.scss';
+
+export interface DataCollectionDialogProps {
+  hideDialog: (dataCollectionEnabled: boolean) => void;
+}
+
+export class DataCollectionDialog extends Component<DataCollectionDialogProps, {}> {
+  constructor(props: DataCollectionDialogProps) {
```
justinwilaby: This constructor is superfluous and can be removed.

packages/app/client/src/ui/dialogs/dataCollectionDialog/dataCollectionDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+    super(props);
+  }
+
+  public render(): ReactNode {
+    return (
+      <Dialog className={styles.dialogMedium} cancel={this.onConfirmOrCancel} title="Help us improve?">
+        <p>
```
justinwilaby: Would this be more readable and allow for more control as a template string?
```js
{`The Emulator includes a telemetry feature that collects usage information. It is important that the Emulator team understands how the tool is being used so that we can improve it. `}
```

packages/app/client/src/ui/dialogs/dataCollectionDialog/dataCollectionDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+          The Emulator includes a telemetry feature that collects usage information. It is important that the Emulator
+          team understands how the tool is being used so that we can improve it.{' '}
+          <a target="__blank" href="https://aka.ms/microsoft">
+            Learn more
+          </a>
+        </p>
+        <p>You can turn data collection on or off at any time in your Emulator Settings.</p>
+        <DialogFooter>
+          <DefaultButton text="Not now" onClick={this.onConfirmOrCancel.bind(this, false)} />
+          <PrimaryButton text="Yes, collect data" onClick={this.onConfirmOrCancel.bind(this, true)} />
```
justinwilaby: For performance reasons, it's best to avoid using bound functions as handlers in `render`. As an alternative, use the [name](https://developer.mozilla.org/en-US/docs/Web/API/Element/name) property to differentiate them. e.g.
```js
<PrimaryButton text="Yes, collect data" onClick={this.onConfirmOrCancel} name="yes" />
//...
private onConfirmOrCancel = (event: MouseEvent<HTMLButtonElement>) => {
  const collectData = event.target.name === 'yes';
  // Do stuff
}
```

tonyanziano: I like your approach much better, this bound pattern seemed weird to me.

PR #1625, opened 2019-06-11
Add background image for when there are no tabs open

### LINKED ISSUE #1508: Add a Background Image for When There Are No Tabs Open
**Describe the solution you'd like**
I think it would be great if we had a background graphic similar to what VSCode has when there are no tabs open:

image

Currently, the Emulator's background is just a blank screen which isn't very nice to look at:

image

**Describe alternatives you've considered**
For starters, I was thinking we could plop an SVG of the logo on that background _until_ and _if_ we decide to go with something else.

**EDIT**:

Quick mockup I did (I think it would look better with text underneath):

image

@DesignPolice / @mewa1024 any thoughts?

denscollo:
Fixes #1508

# Specific changes
- Add background image in the dialog explorer
- Fix dialog-bg color in dark theme

# Testing
image

### DIFF
--- a/packages/app/client/src/ui/shell/main.scss
+++ b/packages/app/client/src/ui/shell/main.scss
@@ -20,15 +20,17 @@
   }
 
   & .mdi-wrapper {
+    background-color: var(--mdi-bg);
     height: 100%;
     width: 100%;
   }
 
   & .mdi-wrapper::before {
     content: "";
+    background-color: var(--mdi-img-bg);
     background-image: url('../media/ic_bot_framework.svg');
     width: 100%;
-    height: calc(100% - 70px); /* centered below toolbar */
+    height: 100%;
     background-size: 50% 50%;
     background-position: 50% 50%;
     background-repeat: no-repeat;
--- a/packages/app/client/src/ui/shell/main.scss
+++ b/packages/app/client/src/ui/shell/main.scss
@@ -24,6 +24,18 @@
     width: 100%;
   }
 
+  & .mdi-wrapper::before {
+    content: "";
+    background-image: url('../media/ic_bot_framework.svg');
+    width: 100%;
+    height: calc(100% - 70px); /* centered below toolbar */
+    background-size: 50% 50%;
+    background-position: 50% 50%;
+    background-repeat: no-repeat;
+    position: absolute;
+    opacity: .2;
+  }
+
   & .secondary-mdi {
     border-left: 1px solid var(--neutral-9);
   }
--- a/packages/app/client/src/ui/styles/themes/dark.css
+++ b/packages/app/client/src/ui/styles/themes/dark.css
@@ -132,7 +132,7 @@ html {
   --s-button-bg-disabled: var(--neutral-2);
 
   /* Dialogs & Modals */
-  --dialog-bg: var(--neutral-13);
+  --dialog-bg: var(--neutral-2);
   --dialog-color: var(--neutral-14);
   --dialog-error-color: var(--error-text);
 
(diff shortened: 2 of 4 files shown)


tonyanziano:
@DesignPolice & @mewa1024 can you take a look at this and give a sign-off when you get a chance?

I noticed the image is completely invisible in High Contrast mode, and this is because we didn't have any comps for High Contrast. Any thoughts on what we want to do in that situation?

image

packages/app/client/src/ui/styles/themes/dark.css
```diff
@@ -132,7 +132,7 @@ html {
   --s-button-bg-disabled: var(--neutral-2);
 
   /* Dialogs & Modals */
-  --dialog-bg: var(--neutral-2);
+  --dialog-bg: var(--neutral-13);
```
tonyanziano: Changing this CSS variable will have some significant side effects throughout the rest of the app because it controls the background color for dialogs. For exmaple, here's what the "New bot" dialog looks like in your branch:

image

I think that a better approach would be to create a new set of css variables that controls the background color only for the mdi & mdi background image so that they are decoupled from the dialog styling. 

ex:
```
--mdi-bg: <color>;
--mdi-img-bg: <color>;
```

packages/app/client/src/ui/shell/main.scss
```diff
@@ -24,6 +24,18 @@
     width: 100%;
   }
 
+  & .mdi-wrapper::before {
+    content: "";
+    background-image: url('../media/ic_bot_framework.svg');
+    width: 100%;
+    height: calc(100% - 70px); /* centered below toolbar */
```
tonyanziano: What's the justification for this height? The `<MDI />` component encompasses the toolbar component and height, but when there are no tabs visible, which will always be the case when this image is showing, then the `<MDI />` is essentially just one, big container, so I think the image can be centered on the background at full height.

When I inspect the background with the above height, it shows the image hoisted above the center 70px

image

Changing the height back to `height: 100%` gives us the image centered over the entire `<MDI />`:

image

which is how the image appears in the mocks.

DesignPolice:
Since this is mostly decorative. For anyone with vision issues I would want it to go away.  So high contrast... just either set the opacity to 0% or 100% so that it is not visible. We may need to verify what it would do with a screen reader, I would hate for it to read:

code 
code
giant logo
code 
code

thanks Marc

packages/app/client/src/ui/styles/themes/dark.css
denscollo: We didn't noticed this when testing. Sorry for that! We'll create a new variable to manage the component background color.

packages/app/client/src/ui/shell/main.scss
denscollo: Thanks for the feedback! We'll be applying this shortly to match the mocks

packages/app/client/src/ui/styles/themes/dark.css
tonyanziano: No worries! Sounds good.

packages/app/client/src/ui/styles/themes/light.css
```diff
@@ -241,4 +241,8 @@ html {
   --radio-bg-color: var(--neutral-1);
   --radio-border-color: var(--neutral-16);
   --radio-selected-border-color: #3B99FC;
+
+  /* MDI Component */
+  --mdi-bg: var(--neutral-2);
```
tonyanziano: Might want to add a `--mdi-bg` variable to `high-contrast.css` for consistency and to decouple it from the `--dialog-bg` variable

denscollo: Ok, I'll do that right away!

Issue #1606, opened 2019-05-31
4.4.2 does not work with Gov-hosted bots


stevengum:
Currently connecting to an Azure Gov bot does not work on the 4.4.2 version of the emulator.

- Connect via endpoint
  - The connect via endpoint dialog does not include the "Azure for US Government Learn more." checkbox that is available when adding a new endpoint to the bot file.
  - Webchat correctly renders, but all messages are Unauthorized
  - Emulator emits the standard "incorrect MSA AppID and MSA AppPassword message"
- Connect via bot file
  - The "Azure for US Government Learn more." was checked.
  - Webchat does not finish rendering.
  - No data on if any activities are sent to the bot

Above data was from testing with @hristp 
The bot hosted in Azure Gov was working with the Test in WebChat in the portal.

### Completion criteria:
- [x] **Connecting to bot in Azure Gov via endpoint should work**
   - ~[ ] Need extendable UI for different sovereign regions~ _Not needed until we add more National Clouds_
- [x] **Connecting to bot in Azure Gov via .bot file should work**
   - ~[ ] Need extendable UI for different sovereign regions~ _Not needed until we add more National Clouds_
- [x] **Inspector mode also works with bot in Azure Gov**
- [ ] **OAuth with a Gov bot should work**
- [x] **Add protocol support (`"livechat"` and `"inspector"`) (new arg for public, `.us`)**

stevengum:
I'm going to try a repro today or early next week, I think this might technically be a P0 but I will leave that to you @cwhitten

stevengum:
I have a repro, but am currently working on #1567

cwhitten:
Confirmed regression against v4.2. I've re-labeled as a P0 for the July release train.

hristp:
Consider using the string "AzureUSGovernment" for Fairfax, to be consistent with naming throughout Azure CLI.

stevengum:
On further investigation, passing a `channelService` value in to `getBotEndpoint` when it pushes a new Endpoint onto the EndpointSet will permit communication with a Gov-hosted bot.

It looks like we may need to add a new header for channel service (e.g. `x-emulator-channelservice`) to pass it in. This leads to the question of what do we want this value to look like, either a simple value like `"public"` or `"usgov"` or the actual value used `""` or `"https://botframework.azure.us"`

`packages\emulator\core\src\conversations\middleware\getBotEndpoint.ts`

```ts
import * as Restify from 'restify';

[...]

PR #1633, opened 2019-06-13
Add Gov support for connecting with bots via endpoint

stevengum:
Partial Fix for #1606 

### Changes:
- Connecting to bot in Azure Gov via endpoint
- Connecting to bot in Azure Gov via .bot file
- Inspector mode also works with bot in Azure Gov
- Add protocol support (`"cloud="`)
- Add and update tests


packages/app/client/webpack.config.js
```diff
@@ -172,7 +172,6 @@ const buildConfig = mode => {
   };
   if (mode === 'development') {
     config.module.rules[0].use = use;
-    config.plugins.unshift(new HotModuleReplacementPlugin(), new HardSourceWebpackPlugin());
```
tonyanziano: It's safe to pull these imports out of the top of the file as well now.

PR #1658, opened 2019-06-25
#1565 Bot State and Bot State Diff Pagination

### LINKED ISSUE #1565: [Bot Inspector] Paged State-Diffing
### Paged State-Diffing 

Currently to move into state diff mode one has to open a context menu (right-click) and select an option from that menu to render the diff. This was a temporary solution until we had the cycles to build state-diff paging into the Bot State Inspector panel, where you can both move into state diff mode on a given Bot State, and page forward/backward to future/previous bot state diffs via buttons. 

*High-level code changes* 

- Introduce new buttons in or near the bot state extension to move into and out of state diff mode and page through state diff modes when currently in state diff mode. 
- Extension manifest change to support Icons
- Adjustment of the Bot State context menu in the chat control to no longer provide the menu item. 
- Provide the extension with all current Activities so it can maintain its own state.

justinwilaby:
Fixes #1565


packages/sdk/main/built/index.d.ts
```diff
@@ -0,0 +1,2 @@
+export * from './ipc';
```
tonyanziano: git ignore `/main/built` ?

PR #1663, opened 2019-07-01
[ISSUE#1622] Can't open new tab with same endpoint URL by clicking the endpoint

### LINKED ISSUE #1622: Cannot connect to new bot without credentials after previously connecting to one with
## Version
4.4.2

## Describe the bug
If I connect to a bot that has credentials then try and open a new chat with another bot that does not have credentials even though I leave the options blank it still cannot connect and I get 401 error. The only way around this is to close the application and start again.

## To Reproduce
1. Start up a bot that has an app ID and app password
2. Connect to bot using emulator
3. Start up a bot that does not have an app ID and app password
4. Connect to bot using emulator. Not how there are now 401 error
5. Restart emulator
6. Conect again to bot using emulator. Works as expected now.

## Expected behavior
I expect to be able to create chats between multiple bots easily.

## Screenshots

[bug]

denscollo:
Fixes #1622

# Description
The existing validations ([here](https://github.com/microsoft/BotFramework-Emulator/blob/88bfc5531e8a2c45db4a6f790db78abbb1f5b68b/packages/emulator/core/src/conversations/middleware/getBotEndpoint.ts#L70-L75)) were not taking into account the case in which a user wants to connect a bot to the emulator using the same endpoint without credentials. We remove the validations to able to rewrite the existing credentials with no contents in this case. 

# Specific changes
- Remove validations to be able to connect a bot without credentials

# Testing
We following the steps to reproduce the issue and the bots were connected successfully.
image

### DIFF
--- a/packages/emulator/core/src/conversations/middleware/getBotEndpoint.ts
+++ b/packages/emulator/core/src/conversations/middleware/getBotEndpoint.ts
@@ -65,14 +65,10 @@ export default function getBotEndpoint(botEmulator: BotEmulator) {
           new BotEndpoint(params.bot.id, params.bot.id, botUrl, msaAppId, msaPassword, false, channelService)
         );
       } else {
-        // update the endpoint in memory with the new
-        // appId and password if they exist in the params
-        if (msaAppId) {
-          endpoint.msaAppId = msaAppId;
-        }
-        if (msaPassword) {
-          endpoint.msaPassword = msaPassword;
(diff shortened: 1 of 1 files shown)


tonyanziano:
I re-ran the CI build for the branch [and it passed](https://fuselabs.visualstudio.com/BotFramework-Emulator/_build/results?buildId=68541).

PR #1670, opened 2019-07-09
Adjusted open url dialog styling.

tonyanziano:
Fixed some minor styling issues with the open url dialog:
- long links were not wrapping properly
- buttons were overflowing out of dialog

**Before:**

dialog_styling_before

**After (long link):**

dialog_styling_after

**After (short link):**

dialog_styling_after_small


tonyanziano:
@corinagum I was leaning towards no since there is no reported issue to go with it. What would Web Chat do in this case?

corinagum:
Yeah I don't think it's a big enough deal. In Web Chat we would file the issue then fix it but I say skip that step here xD

Issue #1672, opened 2019-07-11
Ensure that webSpeechPonyfillFactory is never set to 'null' in the store


tonyanziano:
Web Chat [does a check for `webSpeechPonyFill`](https://github.com/microsoft/BotFramework-WebChat/blob/master/packages/component/src/Dictation.js#L75) in the `<Dictation>` component. It usually defaults to `{}` if the pony fill is `undefined`, however the Emulator passes it in with a value of `null` which skips the default and causes a bad reference error.

We should adjust the logic in the Emulator so that this value is never set to `null` in the store, but `undefined` instead.

PR #1675, opened 2019-07-11
Bumped Web Chat and Emulator to 4.5.0

tonyanziano:
Also fixed an issue that was causing Web Chat to throw when the component was mounted. I have an issue tracking that long-term fix at #1672 .


corinagum:
Looks like the linter failed to run :/

PR #1721, opened 2019-08-13
Consolidated application state store.

### LINKED ISSUE #1697: Consolidated state management
# Spec for Emulator State Consolidation

## Overview

The application state within the Emulator is hard to manage and prone to errors because of the fact that the main ("server") and renderer ("client") processes share some state, however there is no single source of truth. This can be fixed by unifying the state between the renderer and main processes of the application.

### The Problem

The Emulator uses the library [Redux](https://redux.js.org/) to manage application state that is then rendered in the view layer using React. Redux uses the concept of a centralized representation of state called a [store](https://redux.js.org/basics/store). The only way to change state within the store is by dispatching or invoking what is called an [action](https://redux.js.org/basics/actions). Actions are usually triggered by interacting with the UI of the application, such as clicking on a button, entering text into an input, dragging elements, etc.

Usually, an application will have a separate state store for the UI (renderer process) and the backend (main process). The Emulator started off by following this same pattern, and had a separate store for the UI and the backend. In the incipient stages of the Emulator, this was fine and served it's purpose because the renderer and main processes cared about different state:

 - renderer: list of currently-used bots, what navigation bar section is currently selected, current theme, etc.
 - main: what server settings were read from disk, bots loaded from disk, etc.

However, as the application grew in complexity, the renderer and main processes have started to depend on the same application state. That means that some application state needs to be shared between both processes. In order to make sure that renderer and main processes represent the same application state, some extra effort is required to guarantee that their states are in sync. Think about what would happen if some state changed on the renderer process, such as an application settings that was changed -- where ngrok.exe is located -- and then that setting was not updated on the main process. Now if the user was trying to connect to a remote bot using ngrok, the main process would still be using the outdated setting, and wouldn't be able to properly launch ngrok.

This leads to several situations where state that is shared between both renderer and main process is being modified, and thus has to be sync'd between both before proceeding in the case that some other shared state is modified. So instead of just issuing an action, that would then change the state of one store, we have to issue an action that changes the state in one store, wait for that to complete, issue another action to change the state of the second store, and finally continue running once that completes. This basically doubles the amount of points of possible failure in our application when changing shared state, and it also makes debugging much more difficult.

### Proposed Solution / Approach

Unify the renderer and main process state stores into one state store that is shared over the entire application. This is a viable approach in our case due to the architecture of Electron. Electron couples the main and renderer processes very closely and makes communication between the two virtually instantaneous which means that we do not have to worry about the latency of network calls when broadcasting updates to the state store.

I think that the easiest and least disruptive solution -- inspired heavily by the [electron-redux package](https://github.com/hardchor/electron-redux#the-solution) -- would be to combine the current state shape of both the renderer and main process into one store on the main process side. Then create a duplicate of that store on the renderer side that serves only as a mirror of the main store. Finally, connect the two using a custom [redux middleware](https://redux.js.org/advanced/middleware) that takes every dispatched [redux action](https://redux.js.org/basics/actions) and forwards it to the opposite process's store using [Electron's IPC API](https://electronjs.org/docs/api/ipc-main#sending-messages).

### Pros

 - Automatically sync state between main and renderer processes
 - We get to remove a lot of code responsible solely for explicitly syncing state
 - Code will be easier to follow

### Cons

- A lot of code churn primarily on the UI side
- Will need to duplicate action & reducer code on both main and renderer side

tonyanziano:
Resolves #1697 

===

This is a huge PR, so here are a couple of notes:

- The real magic happens in `forwardToMain.ts`, `forwardToRenderer.ts`, & `store.ts` on both main and client
- The reducers & actions had to be combined and duplicated on both sides. I'm thinking that in a future PR, these could all be moved to the `/app/shared/` package so that we only need one copy of each reducer and set of actions
- Removed a lot of unnecessary code calling between main / renderer to sync data
- Removed some unused / unnecessary reducer & action code
- Added missing tests for all state actions ~~(planning to add tests for all state reducers)~~ **DONE**


cwhitten:
@tonyanziano from the description it looks like you still plan to implement some additional tests?

> (planning to add tests for all state reducers)

tonyanziano:
@cwhitten Correct, but I was thinking of including them in a separate PR to keep this one as small as possible.

I can include them in this PR though if you'd prefer.

tonyanziano:
Reducer tests have been added.

Issue #1532, opened 2019-05-07
4.5 Release Checklist


cwhitten:
### .bot files
- [ ] create a new .bot file with keys encrypted
- [ ] edit existing .bot file providing secret to decrypt keys for .bot file not created by the Emulator
- [ ] load a .bot file to start conversation

### starting conversations
- [ ] start conversation via .bot file
- [ ] start conversation via url
- [ ] restart .bot file conversation (same user id, new user id, overridden user id)
- [ ] restart url conversation (same user id, new user id, overridden user id)

### chat control
- [ ] rendered activity types are clickable and load in the inspector

### log panel
- [ ] see appropriate status codes (200, 201, 401, 400, etc)
- [ ] all clickable elements load appropriate inspectors (Message, protocol events, LUIS traces, etc)

### inspectors
- [ ] LUIS train/publish via a LUIS trace

### bot inspector
- [ ] rendered activity types are clickable and load in the inspector (message, bot state)
- [ ] log items render appropriately and are clickable

### transcripts
- [ ] emulator mode save
~~- [ ] inspector mode save~~ removed until we can distinguish between regular & inspector transcripts
- [ ] load into conversation in transcript mode

### service connections
- [ ] log in to Azure
- [ ] navigate discovered resources/subscriptions
- [ ] add LUIS/QnA/Dispatch services to .bot file
- [ ] add new service manually

### samples
- [ ] core-bot
- [ ] welcome users
- [ ] bot authentication
- [ ] nlp-with-dispatch
- [ ] qnamaker

### protocols/schemes
- [ ] using `bfemulator://` can start the emulator with the appropriate context

### platforms
- [ ] macOS
- [ ] Windows
- [ ] Linux 

### auto-update
- [ ] macOS
- [ ] Windows
- [ ] Linux

corinagum:
Checking for an update on 4.4.1-59264 (must be in administrator mode) enables Restart to auto update, which updates to 4.4.1. Is this behavior we want?

tonyanziano:
@corinagum If you don't have the "Use pre-release versions" checkbox checked in Settings, then yes, it is expected to update to the latest stable release. The new patch, 4.4.2, will not require admin privileges though.

tonyanziano:
Signing off for:

- [X] Windows
- [X] Linux

tonyanziano:
4.5 is live!

https://github.com/microsoft/BotFramework-Emulator/releases/tag/v4.5.0

[...]

PR #1763, opened 2019-08-21
Added several e2e tests from the release checklist.

### LINKED ISSUE #1621: [Testing] Implement E2E / Integration Testing
~~It would be great to have E2E tests that we could run on the app. Ideally in packaged form, but unpackaged would be better than what we currently have.~~

~~Here are some options:~~

~~[Spectron](https://electronjs.org/spectron) - a framework for testing un / packaged Electron apps
[Cypress](https://www.cypress.io/) - framework for testing anything that runs in a browser (would take some extra effort to get it working with the Emulator)~~

**UPDATED:**

Initial e2e testing infrastructure has been set up using Spectron in PR #1696 

This issue will now be used to track implementing integration tests for all the corresponding release checklist items in #1532.

tonyanziano:
Fixes #1621 

===

Added tests to cover several checklist items (#1532):

  - importing / exporting transcripts
  - starting a conversation via .bot file / URL
  - setting custom ID via settings, and restarting conversation with same / new user ID
  - clicking on log panel items and activities in Web Chat and ensuring that the inspector populates with the JSON inspector

image


packages/app/main/e2e/createBot.spec.ts
```diff
@@ -79,7 +79,7 @@ describe('Creating a bot', () => {
     await app.client.click('button[name="create-bot-save"]');
 
     // wait for Web Chat to show up
-    await app.client.waitForExist('div[role="log"] + div[role="form"]', 2000);
+    await app.client.waitForExist('div[role="log"] + div[role="form"]', 3000);
```
cwhitten: any reason why this had to be bumped up?

packages/app/main/e2e/README.md
```diff
@@ -0,0 +1,23 @@
+# Bot Framework Emulator End-to-End (e2e) Tests
+
+## Pre-requisites
+
+To run all the tests successfully , you must first install all the dependencies for the `e2e-test-bot`, and then run it before running the tests:
+
+1. Navigate to the `<root>/packages/app/main/e2e/test-bot/` directory.
```
cwhitten: did you consider installation alternatives? it would be a good dev experience to hook this install in to the existing install off of lerna. we wouldn't want to do this out of a development environment, though. looking into the `bootstrap` documentation it looks like we can us an `--ignore` field that may be able blacklist this package driven by configuration. maybe we introduce a new script: `npm run bootstrap:dev`? we could introduce a new package named `tools`?

packages/app/main/e2e/createBot.spec.ts
tonyanziano: One thing I learned from setting up all these tests is that starting a chat session with a bot via .bot file is a significantly slower process than starting a chat session with a bot via URL.

I noticed this in the test suite that talks to a bot via .bot file, and bumped up the Web Chat wait times there, and I bumped this one up to be consistent. It was also failing intermittently with the shorter time.

I think that when I get some cycles I'm going to investigate the difference between the URL / .bot file code paths and see if we can unify the two, because they are pretty different and I don't really understand why they start conversations differently.

Issue #1766, opened 2019-08-22
[MAS1.4.3] [Visual Requirements -Settings] Luminosity Ratio is less than 4.5:1 for “User ID” element 


Kumar2608:
**Actual Result:**
Luminosity Ratio is less than 4.5:1 for “User ID” element

**Expected Result:**
Luminosity Ratio should be greater than or equal to 4.5:1 for “User ID” element

 **Repro Steps:**
1.Open the Bot Framework V4 Emulator.
2.Navigate to Welcome tab screen and select it, screen gets opened  
3.Navigate to the open bot and select it  
4.Open bot dialog gets opened and connect to the open bot 
5.Live chat screen gets appears 
6.Navigate to the Edit ngrok settings and select it 
7.Emulator screen gets opened 
8.Navigate to the “User ID” element 
9.Open 'Colour Contrast Analyser' application.
10.Verify whether Luminosity Ratio is less than 4.5:1 for “User ID” element

**User Impact:**
Users with low vision or colour blindness needs that difference between contrast of text is always high

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Browser: Edge
Application: Bot Framework V4 Emulator
Tool: Colour contrast analyser

**MAS Reference Link:**
[Sharepoint Link](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={a73546c7-765f-489c-b18f-afb659fe99e6})
![MAS1 4 3_Setting](https://user-images.githubusercontent.com/53518994/63507027-6958c380-c4f4-11e9-8ae0-185d41fa2f2e.png)

Kumar2608:
Verified in latest bot emulator Version 4.5.2-78641 and issue got fixed Hence closing this bug
Now Luminosity ratio is 3:1 for User ID element

PR #1791, opened 2019-08-29
[#1766] Increase contrast of input labels on light theme

corinagum:
#1766

Disabled input contrast ratio on light theme was not high enough. This fix changes the text color of disabled inputs and labels in light mode. (Dark mode and high contrast already pass)

See below:
image

Got a verbal okay on this from @DesignPolice, but lmk if anyone wants to discuss this. At the moment the list of approved greys for Emulator does not have a large selection of text colors that will work with a11y on white.


DesignPolice:
Thanks, Corina - I think this will be just fine without having to change lots of other moving parts. Appreciate all your help.

Issue #1699, opened 2019-07-24
Introduce a context menu -> copy JSON in the JSON inspector


cwhitten:
With the collapsible JSON extension, users lost the ability to easily copy the payload. We'll introduce functionality for them to get the JSON data on their clipboard.

1. Right clicking on the activity in the chat control
2. A button in the inspector itself allowing the user to copy the raw data
3. a context-menu somewhere inside the extension with the option to copy

see #1694

cwhitten:
@mewa1024 can you comment on one of the three options we have available to us above?

mewa1024:
I like option 2. The button should say "Copy" and have no icon (to avoid a proliferation of icons when in sidecar mode).

Longterm, can w consider another Pivot that says "RAW JSON", which allows the user switch between views, similar to this view in Firefox?
![image](https://user-images.githubusercontent.com/39317542/61979090-bdc95b80-afa7-11e9-961d-8cc05ab7c231.png)

mockup:
![image](https://user-images.githubusercontent.com/39317542/61979361-7b544e80-afa8-11e9-
![image](https://user-images.githubusercontent.com/39317542/61979537-fc134a80-afa8-11e9-9a8e-0f4ee7ae6f39.png)

tonyanziano:
Note to fixer (@denscollo ):

It looks like this is already implemented for all activity types -- Activity, Trace, Command -- when the livechat session is in debug (inspector) mode:

![image](https://user-images.githubusercontent.com/3452012/62251521-dd47f600-b3a4-11e9-997a-42a6a7b1685f.png)

However, it is disabled for livechat sessions in default mode. @cwhitten do you happen to know if that was an intentional design decision for inspector mode only?

If not, I don't see any harm in enabling the same context menu for activities in default mode;

The function responsible for initiating the context menu in `chat.tsx` can be found [here](https://github.com/microsoft/BotFramework-Emulator/blob/master/packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx#L283) and the chat saga that eventually contacts the main process to show the menu can be found [here](https://github.com/microsoft/BotFramework-Emulator/blob/master/packages/app/client/src/data/sagas/chatSagas.ts#L81).

cwhitten:
Let's go ahead and wire up the Copy button as @mewa1024 suggests, and show the `Copy json` menu item on all activity types regardless of connection mode.

tonyanziano:
Addressed in #1801

Issue #1694, opened 2019-07-18
Would be great to also have a raw JSON view


JeffCordes:
## Version
4.5.2

## Describe the bug
You can no longer view raw json from events and activities posted to and from the bot.  I understand the new tree view may be convenient and familliar to some, and would support it being a view option, but still keep the ability to view raw json.

## To Reproduce
1. Open bot emulator connect to a bot and send messages.
2. Try to copy the body of an event to try paste into another curl program like postman
3. notice there is no longer json, but a nested tree view

## Expected behavior
you can view raw json and copy/paste it

[bug]

tonyanziano:
Hi @JeffCordes ,

Thanks for the input. Would something like a "Copy Raw JSON" button at the top of the JSON view be useful?

Or would you rather it had a tab / dropdown that allowed you to switch between raw and tree view?

JeffCordes:
Thanks @tonyanziano 

I would prefer to have toggle to switch between views, as sometimes you would need to to copy only certain objects out of the json, like the entire from and recipient objects.

To give this some context, we are working on a bot that consumes many custom events (IEventActivity that has a custom data object in its value parameter),  The need for the json is really because there is no way to send a custom event in the emulator.  So the way to emulate the events is to copy an existing message json, manipulate it you be our payload, and post in in postman.

Perhaps in a future feature request the ability to send user defined events that automatically get all the correct user/conversation/etc info for the chat session could be useful.  like we supply a "name" and "value" (in json) and emulator can post that to the bot with all the conversation info.

we have rolled back versions and turned off updates for now.  unfortunately the new view is a blocker for us for now.

tonyanziano:
@JeffCordes I can definitely see why that would be useful in your case.

We actually have had several asks for a feature that allows the user to send custom event payloads such as the one you described -- see #416 .

Thanks again for the feedback, I'll mention this to design.

darrenj:
I just ran into this as well, a raw JSON view would be good but in the meantime a copy raw JSON button/option would go a long way. I use this a lot to retrieve a well formed Activity for other types of testing so the loss of the ability to copy/paste is an issue for this. 

For now you can workaround by saving transcript and opening in a text editor but not ideal :-)

darrenj:
[...]

PR #1801, opened 2019-08-30
[ISSUE#1699] Introduce a context menu -> copy JSON in the JSON inspector

denscollo:
Solves #1699 #1694 

### Description
Allows the users to copy the activities in JSON format using the _"Copy json"_ button placed in the inspector panel or using the context menu items for the activities in the chat session. 

### Changes made
- Add _"Copy json"_ button
- Add test for context menu displayed for the activities in default mode
- Enable context menu for activities in default mode 
- Add test when the "Copy json" button is clicked

### Testing
The next image shows how the "Copy json" button works.
image

Also, this one shows the context menu enabled for the activities in default mode.
image

### DIFF
--- a/packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.tsx
+++ b/packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.tsx
@@ -33,7 +33,7 @@
 
 // Cheating here and pulling in a module from node. Can be easily replaced if we ever move the emulator to the web.
 // @ts-ignore
-import * as Electron from 'electron';
+import { clipboard } from 'electron';
 import {
   EmulatorChannel,
   ExtensionChannel,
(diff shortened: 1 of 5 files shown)


packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.tsx
```diff
@@ -365,6 +366,11 @@ export class Inspector extends React.Component<InspectorProps, InspectorState> {
 
   private accessoryClick = (event: MouseEvent<HTMLButtonElement>): void => {
     const id = event.currentTarget.name;
+
+    if (id == 'copyJson') {
+      return Electron.clipboard.writeText(JSON.stringify(this.state.inspectObj, null, 2));
```
tonyanziano: Minor nitpick:

Would prefer if the import statement was:

`import { clipboard } from 'electron'`

and this line was:

`return clipboard.writeText(...)`

Issue #1743, opened 2019-08-19
[MAS1.4.3] [Visual Requirements-Create a new bot configuration] Luminosity Ratio is less than 4.5:1 for “Hide, Copy, Error message” controls


Kumar2608:
**Actual Result:**
Luminosity Ratio is less than 4.5:1 for “Hide, Copy, Error message” controls and Learn more links 

**Expected Result:**
Luminosity Ratio should be greater than or equal to 4.5:1 for “Hide, Copy, Error message” controls and Learn more links

**Note**: Same issue repro for below path 
Resources->Transcripts and chat files->choose a different location.
Luminosity ration is less than 4.5:1 for Browse control
Please refer the attachment:MAS1 4 3_Choose different location

**Repro Steps:**
1.Open the Bot Framework V4 Emulator.
2.Welcome tab screen gets opened
3.Navigate to the Create a new bot configuration and select it
4.Dialog gest opened
5.Navigate to the “Hide, Copy, Error message” controls and Learn more links  
6.Open 'Colour Contrast Analyser' application.
7. Verify whether Luminosity Ratio is less than 4.5:1 for “Hide, Copy, Error message” controls and Learn more links  

**User Impact:**
Users with low vision or colour blindness needs that difference between contrast of link is always high.

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Browser: Edge
Application: Bot Framework V4 Emulator
Tool: Colour contrast analyser

**MAS Reference Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={a73546c7-765f-489c-b18f-afb659fe99e6}](url)
![MAS1 4 3_Create new bot configuration](https://user-images.githubusercontent.com/53518994/63261870-21426280-c2a2-11e9-9383-c5cb1eb72cb2.png)
![MAS1 4 3_Choose different location](https://user-images.githubusercontent.com/53518994/63588840-1864d000-c5c5-11e9-886c-08fc1517c0b8.png)

corinagum:
Looping @DesignPolice in - I have fixed the anchor color, but I wanted to check with you on the yellow text. (I also sent this message on Teams, if you need more context.

Currently, the text color is #B89500 on `#f4f4f4` background, but the contrast ratio isn't high enough.

[...]

PR #1827, opened 2019-09-04
#1743 - Fix endpoint notify text color

corinagum:
#1743 

Previously the endpoint notification text for opening a bot was showing as a low-contrast yellow. I fixed the text to a dark grey per @DesignPolice. Since the openBotDialog and botCreationDialog components were displaying different colors for these texts, I made both of them dark grey. 

image
image

High Contrast:
image
image

(And for sanity, Dark:)
image

### DIFF
--- a/packages/sdk/ui-react/src/widget/autoComplete/autoComplete.scss
+++ b/packages/sdk/ui-react/src/widget/autoComplete/autoComplete.scss
@@ -61,7 +61,7 @@
     position: absolute;
     top: calc(100% + 2px);
     left: 0;
-    color: var(--endpoint-warning);
+    color: var(--auto-complete-warning-color);
   }
 }
 
--- a/packages/sdk/ui-react/src/widget/autoComplete/autoComplete.scss
+++ b/packages/sdk/ui-react/src/widget/autoComplete/autoComplete.scss
@@ -61,7 +61,7 @@
     position: absolute;
     top: calc(100% + 2px);
     left: 0;
-    color: var(--neutral-14);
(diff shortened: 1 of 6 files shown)


tonyanziano:
Check this on High Contrast, I'm pretty sure it's going to be invisible because the HC dialog background is black.

DesignPolice:
> Check this on High Contrast, I'm pretty sure it's going to be invisible because the HC dialog background is black.

:-( well in HC if it was black background it would need to switch to White... seems like it should go to a white background vs. black?

tonyanziano:
@DesignPolice in HC we used to have light-themed dialogs with an #F4F4F4 background, but we switched away from that.

We could discuss going back to that model if it makes more sense from a design standpoint.

packages/sdk/ui-react/src/widget/autoComplete/autoComplete.scss
```diff
@@ -61,7 +61,7 @@
     position: absolute;
     top: calc(100% + 2px);
     left: 0;
-    color: var(--error-text);
+    color: var(--endpoint-warning);
```
tonyanziano: It seems strange to have the autocomplete widget depend on such a specific color variable (`--endpoint-warning`) when in the future it could be applied to different pieces of UI.

I would go with a separate, more general variable name, and then that variable can pull its value from the `--endpoint-warning` variable:


Ex.
```
--endpoint-warning: var(--neutral-14);
...
--autocomplete-warning-color: var(--endpoint-warning);
```

PR #1839, opened 2019-09-05
Turn nodeIntegration off in OAuth windows

danwkennedy:
OAuth providers can use client side scripts such as JQuery during the authorize process. Electron has trouble with client side JQuery due to node extensions it adds to the window object leading to the user seeing a blank page. Since the authorize window should be treated like a normal browser, turn the node integrations off as recommended here:
https://electronjs.org/docs/faq#i-can-not-use-jqueryrequirejsmeteorangularjs-in-electron

### DIFF
--- a/packages/app/main/src/windowManager.ts
+++ b/packages/app/main/src/windowManager.ts
@@ -133,7 +133,7 @@ export class WindowManager {
       height: 600,
       title: 'Sign In',
       webPreferences: {
-        nodeIntegration: false
+        nodeIntegration: false,
       }
     });
     this.add(win);
--- a/packages/app/main/src/windowManager.ts
+++ b/packages/app/main/src/windowManager.ts
@@ -134,7 +134,7 @@ export class WindowManager {
       title: 'Sign In',
       webPreferences: {
         nodeIntegration: false,
-      }
+      },
     });
     this.add(win);
     const webContents = win.webContents;
--- a/packages/app/main/src/windowManager.ts
+++ b/packages/app/main/src/windowManager.ts
@@ -132,6 +132,9 @@ export class WindowManager {
       width: 800,
       height: 600,
       title: 'Sign In',
+      webPreferences: {
+        nodeIntegration: false
+      }
     });
     this.add(win);
     const webContents = win.webContents;


tonyanziano:
@danwkennedy were you experiencing a blank authorization window when trying to sign in to Azure? 

Or is this a preventative change?

danwkennedy:
@tonyanziano This is to support OAuth logins for Azure DevOps. On the latest emulator version, I’d get a blank screen whenever I tried to log in. What was weird was I could see the js scripts get loaded in the emulator’s DevTools and if I copy pasted the link into Chrome it worked just fine. A little digging led me to that Electron page.

I also think this is the expected behavior since that window should sandbox the authorization call in a browser window. There’s an experimental “sandbox” setting in Electron that I believe does this but it’s not fully supported so I didn’t opt into using it.

danwkennedy:
@tonyanziano is there any interest in getting this PR through. We're getting to the point where we're distributing custom built versions of the emulator to work around this issue.

tonyanziano:
Hi @danwkennedy ,

I apologize for taking so long on this one. We are currently undergoing an accessibility audit and have been very busy with resolving those issues.

I think this PR is reasonable, and I'm find with taking it. Let me confirm with my teammates and we can get this in.

packages/app/main/src/windowManager.ts
```diff
@@ -132,6 +132,9 @@ export class WindowManager {
       width: 800,
       height: 600,
       title: 'Sign In',
+      webPreferences: {
+        nodeIntegration: false
```
tonyanziano: need a comma at the end of the line or linting will fail:

`nodeIntegration: false,`

packages/app/main/src/windowManager.ts
```diff
@@ -132,6 +132,9 @@ export class WindowManager {
...
+      }
```
tonyanziano: need a comma at the end of the line or linting will fail:

`},`

danwkennedy: Added comma.

packages/app/main/src/windowManager.ts
danwkennedy: Added the comma.

danwkennedy:
@tonyanziano Thanks! Fixed the two missing commas.

Issue #1819, opened 2019-09-03
[MAS4.1.2] [Screen Reader-Debug] Screen reader is not reading the Name for “Previous and Next” arrow buttons  


Kumar2608:
**Actual Result:**
Screen reader is not reading the Name for “Previous and Next” arrow buttons present in the Json tab 
Screen reader is reading as button only

**Expected Result:**
Screen reader should narrate the Name for “Previous and Next” arrow buttons  

Example: Screen reader should narrate as Previous bot state button \Next bot state button

**Repro Steps:**
Turn on The NVDA 
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened
3. Bot screen upon connecting via azure/teams
4.Debug screen gets appeared on emulator 
5. Navigate to the “Previous and Next” arrow buttons  
6. Observe the issue

**Note:**
Same issue observed in MAC Os.

**User Impact:**
Users who rely on screen reader will be affected to navigate the control if screen reader is not reading the Name for button controls screen reader user will be affected.

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Application: Bot Framework V4 Emulator
Screen reader: NVDA 

OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

**MAS Reference Link:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={248054a6-5e68-4771-9e1e-242fb5025730}
[MAS4.1.2_Debug.zip](https://github.com/microsoft/BotFramework-Emulator/files/3569325/MAS4.1.2_Debug.zip)

Kumar2608:
Verified in latest bot emulator Version 4.5.2-79718 with NVDA(2019.2) and issue has partially fixed .Hence reactivating the bug
Now NVDA is reading as "Previous button previous"
Same for Next and show diff controls 

**Expected:** NVDA should read as "Previous button bot state tooltip:previous"

tonyanziano:
@awalia13 

The following narration:

> Previous button bot state tooltip:previous

can't be achieved with standard HTML markup alone using a `<button>` tag.

If a `<button>` tag contains a `title` attribute (tooltip), then the value of the tooltip will be narrated after the button is narrated but without the words "tooltip" in front of it.

As an example, the "Previous" button we currently have is a button with a title (tooltip) of "previous".

```
[...]

PR #1841, opened 2019-09-05
[ISSUE#1819][MAS4.1.2] [Screen Reader-Debug] Screen reader is not reading the Name for “Previous and Next” arrow buttons

denscollo:
Solves #1819 

### Description
This pull request enables the proper reading of the “Previous and Next” arrow buttons 

### Changes made
- Add aria-label to left and right arrows buttons
- Update set Previous/Next instead of Left/Right


packages/extensions/json/bf-extension.json
```diff
@@ -56,11 +57,12 @@
             "id": "rightArrow",
             "states": {
               "default": {
-                "icon": "rightArrow"
+                "icon": "rightArrow",
+                "aria-label": "Next"
               },
               "selected": {
-                "aria-selected": true,
-                "icon": "rightArrowSelected"
```
tonyanziano: Is there a reason the `icon` key was removed for the `selected` states?

denscollo: We discussed this a bit with the team and it was because it wasn't actually being used. But we put it back, since if the path on the CSS changes, it won't work. Thanks for noticing!

Issue #1744, opened 2019-08-19
[MAS4.2.1] [Screen Reader-Create a new bot configuration] Screen reader is reading the unnecessary information for “Bot name and End point URL “text fields  


Kumar2608:
**Actual Result:**
Screen reader is reading the unnecessary information for “Bot name and End point URL “text fields  
Screen reader is reading as “Bot name asterisk edit required invalid space scan off 

**Expected Result:**
Screen reader should read the proper info for “Bot name and End point URL “text fields  
Example: Screen reader should read as “Bot name asterisk edit field required “

Note: In scan mode narrator focus twice encountered on "Bot name and End point url" and reading as blank.
Please refer the attachment:MAS4.2.1_Blank.zip

**Note:** Issue repro’ing with NVDA also. Please refer the attachment <MAS4.2.1_Create New bot configuration_NVDA.wmv>

**Repro Steps:**
Turn on the narrator 
1.Open the Bot Framework V4 Emulator.
2.Welcome tab screen gets opened
3.Navigate to the Create a new bot configuration and select it
4.Dialog gest opened
5.Navigate to the “Bot name and End point URL” text fields 
6.Observe the issue 

**User Impact:**
Users who rely on screen readers will be affected if the screen reader is reading the inappropriate info for text fields users will be misguide the info for fields 

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Browser: Edge
Application: Bot Framework V4 Emulator
Screen reader: Narrator 

**MAS Reference Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={a12f3779-bdf1-40c2-bbcc-0e314ea94858}](url)
[MAS4.2.1_Create new bot configuration.zip](https://github.com/microsoft/BotFramework-Emulator/files/3515679/MAS4.2.1_Create.new.bot.configuration.zip)
![MAS4 2 1_Create new bot configuration](https://user-images.githubusercontent.com/53518994/63263019-1e953c80-c2a5-11e9-9ca3-67171ea1805b.png)
[MAS4.2.1_Blank.zip](https://github.com/microsoft/BotFramework-Emulator/files/3515730/MAS4.2.1_Blank.zip)

[...]

PR #1844, opened 2019-09-05
#1744 Fix Restart conversation menu

corinagum:
#1744 

This PR tweaks the "Restart conversation" menu items in the chat view.

**Please note: this component has been updated from 'listbox' role to 'menu' role.** 

Semantically, the behavior we want to emulate more closely matches the menu bar. For an accessible example, please see [w3.org menubar2](https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-2/menubar-2.html)

This fix returns focus to the caret (next to Restart conversation button) after an action has been selected from the submenu. 

Below is the transcribed behavior when navigating the toolbar that contains Restart Conversation.

- Focus jumps to "Restart conversation" button
   - NVDA says: "Restart conversation 1 of 3"
- <kbd>Tab</kbd> : NVDA says: "Restart conversation submenu 2 of 3"
- <kbd>Tab</kbd> : NVDA says: "Save transcript"

When focused on "Restart conversation menu":
<kbd>Enter</kbd> : NVDA says: "Restart with new user ID 1 of 2"
<kbd>&darr;</kbd> : NVDA says "Restart with same user ID 2 of 2"
<kbd>Enter</kbd> : NVDA says "Restart conversation submenu 2 of 3"

Below is an animated gif of the behavior using keyboard navigation.

ezgif-4-da0ca77dcfcd


packages/sdk/ui-react/src/widget/splitButton/splitButton.tsx
```diff
@@ -145,6 +149,7 @@ export class SplitButton extends React.Component<SplitButtonProps, SplitButtonSt
       onClick(newValue);
     }
     this.hidePanel();
+    this.caretRef.focus();
```
tonyanziano: is this necessary if `hidePanel()` makes the same call to `this.caretRef.focus()`?

corinagum: Ahhhh forgot to delete that line. Will fix.

compulim:
We go back-and-forth, then back-and-forth again. Almost rewrote it. But we finally figured out it should be `role="menu"` than `role="listbox"`. 🤣

So many work and think process, so little showing in thi0s PR.

Issue #1799, opened 2019-08-30
[MAS4.2.1] [Screen reader- Add language understanding] NVDA is reading the inappropriate  information for close button


Kumar2608:
**Actual Result:**
NVDA is reading as heading level 2 close button 

**Expected Result:**
NVDA should read as close button only.

**Note:** This issue exits for throughout the application for all close button controls 
Note: Same issue repro for below path
Bot Explorer->Services->+symbol->Add QnA maker
Bot Explorer->Services->+symbol->Add Dispatch

**Repro Steps:**
Turn on the NVDA 
1.Open the Bot Framework V4 Emulator.
2.Navigate to the bot explorer and select it 
3.Navigate to the services symbol and select it 
4.Navigate to the Add Language understanding(LUIS) and select it 
5.Dialog gets opened 
6.Navigate to the Close button 
7.Observe the issue  

**User Impact:**
Users who rely on screen reader will be affected the if the screen reader is reading the inappropriate information for control screen reader users will mis guide the control info.

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Application: Bot Framework V4 Emulator
screen reader: NVDA 

**MAS Reference Link:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={a12f3779-bdf1-40c2-bbcc-0e314ea94858}
[MAS4.2.1_Add language understanding.zip](https://github.com/microsoft/BotFramework-Emulator/files/3559894/MAS4.2.1_Add.language.understanding.zip)

PR #1845, opened 2019-09-05
[#1799] Fixed a screen reader issue with the dialog close button.

tonyanziano:
#1799

===

The narrator was reading out "Header level 2" when reading out the information for the dialog close button, but now it only reads out the information for the close button.


tonyanziano:
@corinagum you did, I just fixed conflicts :)

Issue #1811, opened 2019-08-31
[MAS1.4.4] [Supporting the Platform -Connect to a dispatch model manually] Close button is getting cropped in resize text mode 


Kumar2608:
**Actual Result:**
Close button is getting cropped in resize text mode (125%)

**Expected Result:**
Close button should not cropped in resize text mode 125%)

Same issue repro for below path 
Bot Explorer->Services->+symbol->Add language understanding-> add a LUIS app manually

**Repro Steps:**
1.Open the Bot Framework V4 Emulator.
2. Navigate to the bot explorer and select it 
3. Navigate to the services(+) and select it 
4.Menu gest opened 
5. Navigate to the Add dispatch and select it and dialog gest opened
6.Navigate to the Connect to a dispatch model manually and select it 
7.Dialog gest opened 
8.Apply the resize text mode (125%) 
9.Observe the issue 

**User Impact:**
Users who rely on AT will be affected if the close button is getting cropped then the low vision users will be affected. 

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Application: Bot Framework V4 Emulator
Resize mode: Settings->Display settings->125%

**Mas reference Link:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={b69e5ccc-dde2-4406-b550-4deb38648f6d}
![MAS1 4 4_Dispatch](https://user-images.githubusercontent.com/53518994/64062779-deb94800-cc08-11e9-989b-61fb43f7c459.png)

Kumar2608:
Verified in latest Bot emulator version -4.5.2-79718 .Issue got fixed hence we are closing this bug.
Now Close button is not cropped and scrollable .

PR #1859, opened 2019-09-10
#1811 Fix overflowing dialogs so they are scrollable

corinagum:
#1811

This fix adds a maximum height to dialogs so that zooming in or overflow will make the dialog scrollable, instead of creating unreachable parts of the modal.

Before scrolling down: 
image
After scrolling down:

image

@DesignPolice - just checking in that not having a visible scrollbar on the outside of the dialog is still the desirable appearance.


tonyanziano:
Just waiting on a response from @DesignPolice regarding the invisible scrollbar, and then we can merge this in.

Issue #1818, opened 2019-09-03
[MAS4.1.2] [Screen Reader-Debug] Screen reader is not reading the state for Expanded or collapsed control 


Kumar2608:
**Actual Result:**
Screen reader is not reading the state for Expanded or collapsed controls under JSON tab 
1.Root 2. Channel Data 3. Conversation 4.Entities 5.From 6.recipient 

**Expected Result:**
Screen reader should narrate the state for Expanded or collapsed controls
Example: Screen reader should narrate as Root expanded or Root collapsed 

**Repro Steps:**
Turn on The NVDA +
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened
3. Bot screen upon connecting via azure/teams
4.Debug screen gets appeared on emulator 
5. Navigate to the “1. Root 2. Channel Data 3. Conversation 4. Entities 5.From 6.recipient 
” controls and action perform on it 
6. Observe the issue

**User Impact:**
Users who rely on screen reader will be affected to navigate the control if screen reader is not reading the state for button controls screen reader user will be affected.

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Application: Bot Framework V4 Emulator
Screen reader: NVDA 

**MAS Reference Link:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={248054a6-5e68-4771-9e1e-242fb5025730}
[MAS4.1.2_Debug.zip](https://github.com/microsoft/BotFramework-Emulator/files/3569292/MAS4.1.2_Debug.zip)

Kumar2608:
Verified in latest Bot emulator version -4.5.2-79718 with NVDA(2019.2) issue got fixed hence we are closing this bug.
Now NVDA is reading the state for 1.Root 2. Channel Data 3. Conversation 4.Entities 5.From 6.recipient

PR #1863, opened 2019-09-11
[ISSUE#1818][MAS4.1.2] [Screen Reader-Debug] Screen reader is not reading the state for Expanded or collapsed control

denscollo:
Solves #1818

### Description

With this PR we enabled the screen reader to identify when the elements of the JSON viewer extension are `expanded` and `collapsed`.

### Changes made

- Screen reader reads collapse/expand

Since the screen reader wasn't reading the `aria-expanded` property from the `ul` tags, we added that property to the parent `li`


denscollo:
Thanks @tonyanziano for the feedback. We'll test it again and fix it!

denscollo:
@tonyanziano we finished applying the changes. Now it's ready for review! Thanks!

Issue #1846, opened 2019-09-06
[MAS4.2.1] [Screen Reader-Services] In browse mode NVDA is reading the unnecessary information for “Key and value” elements


Kumar2608:
**Actual Result:**
In browse mode NVDA is reading the unnecessary information for “Key and value” elements and reading the key and value info once at a time .

**Expected Result:**
In browse mode NVDA should not read the unnecessary information for “Key and value” elements and  NVDA focus should moves to the individual elements for key and value and should read the key and value info

**Repro Steps**:
Turn on the NVDA
1.Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened
3.Navigate to the BOT explorer and select it 
4.Mavigate to the Services (+) and select it 
5.Dialog gets opened 
6. Navigate to the Add other resource and select it 
7.Navigate to the Key and value elements using browse mode (Down arrow key)
8.Observe the issue

**User Impact:**
Users who rely on screen readers will be affected if the screen reader is reading the inappropriate info for elements then the screen reader users will be misguide the info for elements 

**Test Environment:**
OS Build: Version 1903(OS Build 18362.295)
Application: Bot Framework V4 Emulator
Screen reader: NVDA 

**MAS Reference Link:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={a12f3779-bdf1-40c2-bbcc-0e314ea94858}
[MAS4.2.1_Add other service.zip](https://github.com/microsoft/BotFramework-Emulator/files/3583666/MAS4.2.1_Add.other.service.zip)

Kumar2608:
Verified in latest bot emulator version 4.5.2-81808 with NVDA(2019.2) issue still repro'ing hence we are reactivating this bug

tonyanziano:
@Kumar2608 , @awalia13 ,

Would it be possible to include an updated video of the issue repro'ing please?

tonyanziano:
Currently, the control reads:

"Key 0 edit, Add a key (optional), blank"
"Value 0 edit, Add a value (optional), blank"

"Key 1 edit, Add a key (optional), blank"
"Value 1 edit, Add a value (optional), blank"

... etc.

The behavior / logic of the control has changed since the initial open of the issue, so a new repro video would be much appreciated.

[...]

PR #1871, opened 2019-09-17
[#1846] Fixed functionality and a11y issues with generic service editor dialog.

tonyanziano:
#1846 

===

The `<KvPair/>` component was actually broken and did not work. Additionally, the accessibility was not satisfactory.

Before, the control would try to auto-add new rows as you started filling in the previous rows, but did not work properly. The new, simplified approach provides buttons to add or remove rows.

The component will now also focus the newest row after adding a row, as well as announce to the screen reader when the user deletes a row.

**Before:**
oldKv

**After:**
newKv

### DIFF
--- a/packages/app/client/src/ui/shell/explorer/servicesExplorer/connectedServiceEditor/connectedServiceEditor.scss
+++ b/packages/app/client/src/ui/shell/explorer/servicesExplorer/connectedServiceEditor/connectedServiceEditor.scss
@@ -73,3 +73,20 @@ ul.kv-pair-container {
 .no-border {
   border: none;
 }
+
+.link-button {
+  display: block;
+  background: 0;
+  border: 0;
+  color: var(--dialog-link-color);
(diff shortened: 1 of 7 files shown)


packages/app/client/src/ui/shell/explorer/servicesExplorer/connectedServiceEditor/connectedServiceEditor.scss
```diff
@@ -73,3 +73,20 @@ ul.kv-pair-container {
 .no-border {
   border: none;
 }
+
+.link-button {
```
corinagum: just FYI I'll replace this with `<LinkButton>` when it's time to merge my PR :)

tonyanziano: Sounds good!

Issue #1849, opened 2019-09-09
[MAS4.2.12] [Screen reader -New bot configuration] After selecting “Hide and copy” controls NVDA is not narrating the any information


Kumar2608:
**Actual Result:**
After selecting “Hide and copy” controls NVDA is not narrating the any information present in the new bot configuration 
When the user selects the Copy controls NVDA focus is moves to whole pane instead of secret text filed 

**Expected Result:**
After selecting “Hide and copy” controls NVDA should narrate the “Hide and copy controls information 
When the user selects the Copy controls NVDA focus should moves to secret text field only and should narrate the info 

**Repro Steps:**
1.	Open the Bot Framework V4 Emulator.
2.	Welcome tab screen gets opened
3.	Navigate to the Create a new bot configuration and select it 
4.	Dialog gest opened and Navigate to the Hide, copy controls, and select it 
5.	Observe the issue 

**User Impact:**
Users who rely on screen readers will be affected  if the Screen reader is not narrating the controls information when the user selected the control then the screen reader users will not get the controls info.

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Application: Bot Framework V4 Emulator
Screen reader: NVDA 

**MAS Reference Link:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={b19c05bd-d82f-4d6d-a8e5-7081bf5c2dba}
[MAS4.2.12_New bot configuration.zip](https://github.com/microsoft/BotFramework-Emulator/files/3590040/MAS4.2.12_New.bot.configuration.zip)

Kumar2608:
Verified in latest bot emulator version 4.5.2-79874 with NVDA(2019.2) issue got fixed. Hence we are closing this bug.
Now NVDA is narrating the alert message and focus is moves to the expected control.

Issue #1820, opened 2019-09-03
 [MAS4.1.3] [Screen reader -Debug] When the user performs the action on “previous, Next, show diff” controls NVDA is not narrating the notification message information


Kumar2608:
**Actual Result:**
When the user performs the action on “previous, Next, show diff” controls NVDA is not narrating the notification message information 

**Expected Result:**
Screen Reader should narrate the notification message information when the user performs the action on “previous, Next, Show diff” controls   
Example: when the user performs the action on “previous” control NVDA should narrate as Previous bot state button 

**Repro Steps:**
Turn on The NVDA 
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened
3. Bot screen upon connecting via azure/teams
4.Debug screen gets appeared on emulator 
5. Navigate to the “Previous and Next” arrow buttons and Show\hide diff controls
6. Observe the issue

**User Impact:**
Users who rely on screen readers will affect to navigate if Screen reader is not narrating the change notification message when we are perform the action on the controls  

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Application: Bot Framework V4 Emulator
Screen reader: NVDA 

**MAS Reference Link:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={248054a6-5e68-4771-9e1e-242fb5025730}
[MAS4.1.3_Debug.zip](https://github.com/microsoft/BotFramework-Emulator/files/3569437/MAS4.1.3_Debug.zip)

tonyanziano:
This might be related to #1773 if we decide to create some sort of invisible screen-reader-only notification control.

tonyanziano:
I'll take this one as I am also doing related #1773, and #1849

Issue #1773, opened 2019-08-22
[MAS4.2.12] [Screen reader -Setting] After selecting “Save button” Narrator is not narrating the notification message information 


Kumar2608:
**Actual Result:**
After selecting “Save button” Narrator is not narrating the notification message information and focus is moving to the Locale text field 

**Expected Result:**
Screen Reader should narrate the notification message after selecting the “Save button ” and focus should moves to the path to ngrok text field.
Example: After selecting “Save button” narrator should narrate as File has been saved or file has been uploaded 

**Note:** Issue is repro’ing with NVDA. Please refer the attchement <MAS4.2.12_Setting_NVDA.wmv>
Same issue observed in Mac Os.

**Repro Steps:**
Turn on the narrator 
1.Open the Bot Framework V4 Emulator.
2.Navigate to Welcome tab screen and select it, screen gets opened  
3.Navigate to the open bot and select it  
4.Open bot dialog gets opened and connect to the open bot 
5.Live chat screen gets appears 
6.Navigate to the Edit ngrok settings and select it 
7.Emulator screen gets opened 
8.Navigate to the “Save button” and select it 
9. Observe the issue

**User Impact:**
Users who rely on screen readers will affect to navigate if Screen reader is not narrating the change notification message when we are selecting the save button 

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Browser: Edge
Application: Bot Framework V4 Emulator
Screen reader: Narrator

OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

**MAS Reference Link:**
[Sharepoint Link](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={b19c05bd-d82f-4d6d-a8e5-7081bf5c2dba)
[MAS4.2.12_Setting.zip](https://github.com/microsoft/BotFramework-Emulator/files/3530144/MAS4.2.12_Setting.zip)

[MAS4.2.12_Setting_NVDA.zip](https://github.com/microsoft/BotFramework-Emulator/files/3545929/MAS4.2.12_Setting_NVDA.zip)

Kumar2608:
Verified in latest bot emulator version-4.5.2-82590 with NVDA(2019.2) Issue got fixed Hence we are closing this bug.

Issue #1730, opened 2019-08-16
[MAS3.3.1] [Screen Reader-Open bot dialog] Screen reader is not reading the Error information when the use enters the invalid values in the Bot URL text field.


Kumar2608:
**Actual Result:**
Screen reader is not reading the Error information when the use enters the invalid values in the Bot URL text field which is present in the open bot dialog.

**Expected Result**:
Screen reader should narrate Error information when the use enters the invalid values in the Bot URL text field 
Example: When the user enters the invalid values in bot url text field screen reader should read as “Please choose a valid bot file or end point URL”.

**Note:** This issue exits throughout the application wherever error message appears  
Same issue repro with NVDA also. Please refer the attachment as <MAS3.3.1_Open a bot_Error Message_NVDA.wmv>
Same issue repro for below path
Bot Explorer->Endpoint->+symbol:
Screen reader is not narrating the error message and error suggestion information for End point URL text field.
Please refer the attachment:MAS3.3.1_End.point.zip

**Repro Steps:**
Turn on the narrator 
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened
3.Navigate to the open bot control and select it 
4.Open bot dialog gest opened 
5.Navigate to the Bot URL text filed.
6. Observe the issue.

**User Impact:**
Users who rely on assistive technologies ,The identification and description of an error is not reading the screen reader so learning cognitive, language, and learning disabilities who have difficulty understanding the info.

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Browser: Edge
Application: Bot Framework V4 Emulator
Screen reader: Narrator 

**MAS Reference Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={733120d0-56c5-479c-889f-fee36ea08ee0}](url)
[MAS3.3.1_Open bot dialog.zip](https://github.com/microsoft/BotFramework-Emulator/files/3509036/MAS3.3.1_Open.bot.dialog.zip)
![MAS3 3 1_Open bot dialog](https://user-images.githubusercontent.com/53518994/63159800-efc75e00-c039-11e9-84d2-a1ba5a99348d.png)

[MAS3.3.1_Open a bot_Error Message_NVDA.zip](https://github.com/microsoft/BotFramework-Emulator/files/3546010/MAS3.3.1_Open.a.bot_Error.Message_NVDA.zip)
[...]

PR #1873, opened 2019-09-17
[#1730, #1773, #1820, #1849]Added screen reader alerts for various actions throughout the app.

tonyanziano:
#1730, #1773, #1820, #1849

===

- Added `ariaAlertService.ts` in `packages/app/client`
  - creates an invisible `<span>` with `role="alert"` that will be read by any screen reader technologies and then removed after 5 seconds

image

- Added screen reader alerts for the following actions:
  - saving app settings
  - bot URL validation messages in Open Bot dialog
  - hiding, showing, and copying bot secret in Create Bot dialog
  - navigating to previous and next bot states in the inspector
  - toggling diff mode in the inspector
  - copying JSON in the inspector
- Fixed accessibility around bot secret input in Create Bot dialog
  - previously, secret input was always disabled, so it was invisible to screen readers
  - secret input is now read-only
  - fixed label on secret input (was not being narrated before)

### DIFF
--- a/packages/app/client/src/ui/a11y/ariaAlertService.ts
+++ b/packages/app/client/src/ui/a11y/ariaAlertService.ts
@@ -30,27 +30,28 @@
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
-// from the extension to Emulator
-export enum EmulatorChannel {
-  CreateAriaAlert = 'create-aria-alert',
-  EnableAccessory = 'enable-accessory',
-  Log = 'logger.log',
-  LogError = 'logger.error',
-  LogLuisDeepLink = 'logger.luis-editor-deep-link',
-  SetAccessoryState = 'set-accessory-state',
-  SetHightlightedObjects = 'set-highlighted-objects',
-  SetInspectorObjects = 'set-inspector-objects',
-  SetInspectorTitle = 'set-inspector-title',
-  TrackEvent = 'track-event',
+
+let singleton: AriaAlertService;
+class AriaAlertService {
+  constructor() {
+    singleton = this;
+  }
+
+  /** Creates an alert and inserts it into the DOM */
+  public alert(msg: string): void {
+    if (!msg) {
+      return;
+    }
+    const prevAlert = document.querySelector('span#alert-from-service');
+    prevAlert && prevAlert.remove();
+    const alert = document.createElement('span');
+    alert.innerText = msg;
+    alert.setAttribute('id', 'alert-from-service');
+    alert.setAttribute('role', 'alert');
+    alert.setAttribute('style', 'position: absolute; top: -9999px; overflow: hidden;');
+    document.body.appendChild(alert);
+  }
 }
 
-// From the Emulator to the extension
-export enum ExtensionChannel {
-  AccessoryClick = 'accessory-click',
-  BotUpdated = 'bot-updated',
-  ChatLogUpdated = 'chat-log-updated',
-  HighlightedObjectsUpdated = 'highlighted-objects-updated',
-  Inspect = 'inspect',
-  Theme = 'theme',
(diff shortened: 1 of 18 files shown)


packages/app/client/src/ui/a11y/ariaAlertService.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  /** Creates an alert and inserts it into the DOM */
+  public alert(msg: string): void {
+    if (!msg) {
+      return;
+    }
+    const alert = document.createElement('span');
+    alert.innerText = msg;
+    alert.setAttribute('role', 'alert');
+    alert.setAttribute('style', 'position: absolute; top: -9999px;');
```
corinagum: should we also make it a 0x0px `<div>` so in the unlikely chance it gets on screen, it's still invisible?

packages/app/client/src/ui/a11y/ariaAlertService.ts
```diff
@@ -0,0 +1,56 @@
...
+    document.body.appendChild(alert);
+    // remove the alert after it has been read
+    setTimeout(alert.remove.bind(alert), 5000);
```
corinagum: I think the general recommendation is that once something is put on the DOM, even if it's invisible, it should stay there (not be removed)

corinagum: If the screenreader is still talking and takes longer than 5 seconds, this item might not be spoken

packages/app/client/src/ui/a11y/ariaAlertService.ts
tonyanziano: I tried this and you can't hide an element with text content: [Example](https://jsfiddle.net/sxygke3n/1/)

I could give it an opacity of 0, but some screen readers do not read content with an opacity of 0.

packages/app/client/src/ui/a11y/ariaAlertService.ts
tonyanziano: I considered the possibility that the duration isn't long enough, but I tested out all the narrations and none of them were close to being cut off.

If that happens down the line we can easily bump up the duration.

tonyanziano: Regarding alert persistence in the DOM, I'm concerned that leaving the alerts in the DOM could lead to a potentially huge amount of lingering DOM elements that aren't being used.

In the case of the Bot State inspector, an alert would be created every time one of the previous or next bot state buttons is pressed.

I think a good compromise would be to follow the model in which the last alert persists in the DOM until another alert is spawned, and then it is removed.

It's used here:

[ARIA Example](https://www.w3.org/TR/WCAG20-TECHS/ARIA19.html)
[MDN Example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/alerts#Checking_for_validity_and_notifying_the_user_2)

tonyanziano:
Addressed feedback

packages/app/client/src/ui/a11y/ariaAlertService.ts
corinagum: Oh yeah, I meant should we change it from <span> to <div> so that it is adjustable to 0px by 0px. Fiddler shows that as working with `overflow: hidden`
https://jsfiddle.net/sxygke3n/1/

packages/app/client/src/ui/a11y/ariaAlertService.ts
corinagum: Instead of a timeout, the ARIA example looks like it's removing the previous alert when the next alert is being added to the DOM. So in the future if we need to address the timeout issue, I would prefer if we implement it that way.

That said, thanks for addressing my comments, everything else looks good to me.

packages/app/client/src/ui/a11y/ariaAlertService.ts
corinagum: (also good to know about the opacity 0 thing, I didn't know that)

tonyanziano:
@corinagum oh duh, `overflow`. I'll add that to the CSS. Good call

PR #1878, opened 2019-09-22
Update Emulator to Use New Cognitive Service API

bwamie:
Update Emulator to Use New Cognitive Service API
Add token refresh 5 minutes before it expires.


packages/app/client/src/state/sagas/chatSagas.ts
```diff
@@ -168,9 +168,11 @@ export class ChatSagas {
     } catch (e) {
       // No-op - this appId/pass combo is not provisioned to use the speech api
     }
-    if (token) {
-      const factory = yield call(createCognitiveServicesBingSpeechPonyfillFactory, {
-        authorizationToken: token,
+
+    if (speechRegionToken) {
+      const factory = yield call(createCognitiveServicesSpeechServicesPonyfillFactory, {
+        authorizationToken: speechRegionToken.access_Token,
```
compulim: When you create a ponyfill factory, the `authorizationToken` should be a *Promise function* that will be called every time the token is required, i.e. for every STT and TTS.

And you will be responsible for caching the token. This is because only the code which generate the token, know how long a token can be used before expiration. We can't look into the token.

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -33,21 +33,25 @@
...
+import { BotEndpointOptions, SpeechTokenInfo, SpeechRegionToken } from '@bfemulator/sdk-shared';
 import * as HttpStatus from 'http-status-codes';
 
 import { authentication, speech as speechEndpoint, usGovernmentAuthentication } from '../authEndpoints';
 import statusCodeFamily from '../utils/statusCodeFamily';
 
 // We will refresh if the token is going to expire within 5 minutes
 const TIME_TO_REFRESH = 5 * 60 * 1000;
+// We will refresh if the token is going to expire within 5 minutes
+const TIME_TO_REFRESH_SPEECH_TOKEN = 300000; // Refresh speech token 5 minutes before it expires (5 * 60 * 1000)
```
compulim: I think we cannot hardcode it here. We could hardcode it on the server (where the token is generated).

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -92,6 +110,29 @@ export default class BotEndpoint {
     }
   }
 
+  public async getSpeechToken(
```
compulim: This code looks duplicated, please refactor.

packages/sdk/shared/src/types/index.ts
```diff
@@ -53,3 +53,4 @@ export * from './logger';
 export * from './speechToken';
 export * from './transcriptRecord';
 export * from './user';
+export * from './speechRegionToken';
```
compulim: Sorting.

packages/sdk/shared/src/types/speechRegionToken.ts
```diff
@@ -0,0 +1,38 @@
...
+// THE SOFTWARE IS PROVIDED ""AS IS"", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+export interface SpeechRegionToken {
```
compulim: Sorting.

packages/sdk/shared/src/types/speechToken.ts
```diff
@@ -33,6 +33,8 @@
 
 export interface SpeechTokenInfo {
   access_Token: string;
+  region: string;
```
compulim: Sorting.

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -92,6 +110,29 @@ export default class BotEndpoint {
...
+    refresh: boolean = false,
+    duration: number = TIME_TO_REFRESH_SPEECH_TOKEN
+  ): Promise<SpeechRegionToken> {
+    if (this.speechRegionToken && !refresh && !this.isTokenExpired()) {
+      if (this.isTokenExpiringSoon()) {
+        this.renewTokenBeforeExpiry(); // Call and forget. No await since valid token still available
+      }
+
+      return {
+        access_Token: this.speechRegionToken.access_Token,
```
compulim: Although the service is returning weird casing + underscore, we should not "inherit" this weird casing in our code.

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -92,6 +110,29 @@ export default class BotEndpoint {
     }
   }
 
+  public async getSpeechToken(
+    refresh: boolean = false,
+    duration: number = TIME_TO_REFRESH_SPEECH_TOKEN
+  ): Promise<SpeechRegionToken> {
+    if (this.speechRegionToken && !refresh && !this.isTokenExpired()) {
+      if (this.isTokenExpiringSoon()) {
+        this.renewTokenBeforeExpiry(); // Call and forget. No await since valid token still available
```
compulim: We should await. The ponyfill factory is designed to be able to await.

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -92,6 +110,29 @@ export default class BotEndpoint {
...
+
+      return {
+        access_Token: this.speechRegionToken.access_Token,
+        region: this.speechRegionToken.region,
+        expiry: this.speechRegionToken.expiry,
+      };
+    }
+
+    if (!this.msaAppId || !this.msaPassword) {
+      throw new Error('bot must have Microsoft App ID and password');
```
compulim: Sounds weird to guard and throw error here.

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -63,25 +67,39 @@ export default class BotEndpoint {
...
+  private isTokenExpiringSoon(): boolean {
+    const currentDateTime = new Date();
+    const timeFromNowMilliseconds = currentDateTime.getTime() + TIME_TO_REFRESH_SPEECH_TOKEN;
+    return timeFromNowMilliseconds >= this.speechRegionToken.expiry;
+  }
 
-    if (!this.msaAppId || !this.msaPassword) {
-      throw new Error('bot must have Microsoft App ID and password');
-    }
+  private isTokenExpired(): boolean {
```
compulim: Don't duplicate code.

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -63,25 +67,39 @@ export default class BotEndpoint {
...
+    const currentDateTime = new Date();
+    const dateTimeUtcNowMilliseconds = currentDateTime.getTime();
+    return dateTimeUtcNowMilliseconds >= this.speechRegionToken.expiry;
+  }
 
-    const query = new URLSearchParams({ goodForInMinutes: duration } as any);
-    const res = await this.fetchWithAuth(new URL(`?${query.toString()}`, speechEndpoint.tokenEndpoint).toString());
+  private renewTokenBeforeExpiry() {
```
compulim: Don't use `then`, use async/await.

packages/app/shared/src/types/speechTypes.ts
```diff
@@ -33,6 +33,8 @@
 
 export interface SpeechTokenInfo {
   access_Token: string;
+  region: string;
```
compulim: I am not familiar with the source code. But why we have the same type in /app/ and /sdk/? I believe there should be only one.

packages/app/shared/src/types/speechRegionToken.ts
```diff
@@ -0,0 +1,38 @@
+//
```
compulim: Why there is a dupe file in `/sdk/`?

packages/app/shared/src/types/index.ts
```diff
@@ -41,3 +41,4 @@ export * from './serverSettingsTypes';
 export * from './speechTypes';
 export * from './luisTypes';
 export * from './clientAwareSettings';
+export * from './speechRegionToken';
```
compulim: Sorting.

packages/emulator/core/src/facility/botEndpoint.ts
bwamie: It is already hardcoded at the service. The server logic compares this when passed in the getToken request to ensure the refresh token works as expected geven that it uses this duration to create token expiry time. However, I would need to add it as another property in the Token response object, if thats okay with you.

compulim: Please let the server pass the hardcoded value, instead of hardcoding it twice in different code base.

packages/app/shared/src/types/speechTypes.ts
bwamie: I already found it that way. I am just updating expisting interphases and following the same pattern. No sure why they exist in two different places.

packages/app/shared/src/types/speechRegionToken.ts
bwamie: Same as above

packages/emulator/core/src/facility/botEndpoint.ts
bwamie: Its not duplicated. I rather introduced getTokenAsync to avoid duplication.  Let me know where you think I have duplicated. I will surely refactor.

packages/emulator/core/src/facility/botEndpoint.ts
bwamie: With which one? Its  not duplacted. If you mean with `isTokenExpiringSoon` then it is not. They  are different.

packages/app/shared/src/types/speechRegionToken.ts
tonyanziano: We should just choose one -- I vote for sdk -- and delete the other one. There should be minimal code churn because the type `SpeechTokenInfo` is only used in one place

packages/emulator/core/src/facility/botEndpoint.ts
compulim: I mean the `isTokenExpired` and `isTokenExpiringSoon` is too similar.

It may make sense to return the expiry, or building a function named `willTokenExpireWithin(durationInMs: number = 0)`.

packages/emulator/core/src/facility/botEndpoint.ts
compulim: I mean it looks a bit late to throw here.

tonyanziano: Yeah it would make more sense to move this up to the top of the function

packages/app/client/src/state/sagas/chatSagas.ts
bwamie: That would be nice but remember we get the region from the api service as well. And no option of region promise or RegionToken promise taken by ponyfill

compulim: Yap, once you set the region, you cannot change it. But you can change the authorization token as needed.

So, the `authorizationToken` should be a Promise function.

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -63,25 +64,32 @@ export default class BotEndpoint {
...
+  private async getTokenAsync(): Promise<SpeechRegionToken> {
+    const res = await this.fetchWithAuth(new URL(speechEndpoint.tokenEndpoint).toString());
 
     if (statusCodeFamily(res.status, 200)) {
       const body = (await res.json()) as SpeechTokenInfo;
 
       if (body.access_Token) {
-        this.speechToken = body.access_Token;
+        this.speechRegionToken = {
+          access_Token: body.access_Token,
```
compulim: Rename to `authenticationToken`.

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -92,6 +100,28 @@ export default class BotEndpoint {
     }
   }
 
+  public async getSpeechToken(refresh: boolean = false): Promise<SpeechRegionToken> {
+    if (!this.msaAppId || !this.msaPassword) {
+      throw new Error('bot must have Microsoft App ID and password');
+    }
+
+    if (this.speechRegionToken && !refresh && !this.willTokenExpireWithin(0)) {
+      if (this.willTokenExpireWithin(this.speechRegionToken.tokenLife * 0.5)) {
```
compulim:
```suggestion
      if (this.willTokenExpireWithin(this.speechRegionToken.tokenLife / 2)) {
```

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -63,25 +64,32 @@ export default class BotEndpoint {
     this.appPassword = msaPassword;
   }
 
-  public async getSpeechToken(refresh: boolean = false, duration: number = 10) {
-    if (this.speechToken && !refresh) {
-      return this.speechToken;
-    }
+  private willTokenExpireWithin(millisecondsToExpire: number): boolean {
```
compulim:
```suggestion
  private willTokenExpireWithin(millisecondsToExpire = 0: number): boolean {
```

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -63,25 +64,32 @@ export default class BotEndpoint {
...
+    const currentDateTime = new Date();
+    const timeFromNowMilliseconds = currentDateTime.getTime() + millisecondsToExpire;
```
compulim:
```suggestion
    const timeFromNowMilliseconds = Date.now() + millisecondsToExpire;
```

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -63,25 +64,32 @@ export default class BotEndpoint {
...
+    return timeFromNowMilliseconds >= this.speechRegionToken.expiry;
+  }
 
-    if (!this.msaAppId || !this.msaPassword) {
-      throw new Error('bot must have Microsoft App ID and password');
-    }
+  private async renewTokenBeforeExpiryAsyn() {
```
compulim:
```suggestion
  private async renewTokenBeforeExpiryAsync() {
```

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -63,25 +64,32 @@ export default class BotEndpoint {
...
     if (statusCodeFamily(res.status, 200)) {
       const body = (await res.json()) as SpeechTokenInfo;
 
       if (body.access_Token) {
-        this.speechToken = body.access_Token;
+        this.speechRegionToken = {
+          access_Token: body.access_Token,
+          region: body.region,
+          expiry: body.expiry,
+          tokenLife: body.tokenLife,
```
compulim: `tokenLife` is not used.

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -63,25 +64,32 @@ export default class BotEndpoint {
...
 
     if (statusCodeFamily(res.status, 200)) {
       const body = (await res.json()) as SpeechTokenInfo;
 
       if (body.access_Token) {
-        this.speechToken = body.access_Token;
+        this.speechRegionToken = {
+          access_Token: body.access_Token,
+          region: body.region,
+          expiry: body.expiry,
```
compulim: The name is too arbitrary.

Does it mean, "expire at 12:34" or "expire after 10 minutes"?

packages/emulator/core/src/authEndpoints.ts
```diff
@@ -57,6 +57,6 @@ export const v32Authentication = {
 };
 
 export const speech = {
-  // Access token for Bing Speech Api
-  tokenEndpoint: 'https://login.botframework.com/v3/speechtoken',
+  // Access token for Cognitive Services Api
+  tokenEndpoint: 'https://login.botframework.com/v3/speechtoken/speechservice',
```
compulim: The official name for the service is "Cognitive Services Speech Services". So I believe this endpoint should be changed.

```suggestion
  tokenEndpoint: 'https://login.botframework.com/v3/speechtoken/speechservices',
```

packages/emulator/core/src/authEndpoints.ts
```diff
@@ -57,6 +57,6 @@ export const v32Authentication = {
 };
 
 export const speech = {
-  // Access token for Bing Speech Api
-  tokenEndpoint: 'https://login.botframework.com/v3/speechtoken',
+  // Access token for Cognitive Services Api
```
compulim:
```suggestion
  // Access token for Cognitive Services API
```

packages/app/client/src/state/sagas/chatSagas.ts
```diff
@@ -168,9 +168,11 @@ export class ChatSagas {
     } catch (e) {
       // No-op - this appId/pass combo is not provisioned to use the speech api
```
compulim: This exception, may not be caused by `commandService.remoteCall` failed. Saga can emit exception like, when cancelling 
a saga. So the exception does not necessarily means it cannot get the token, but something else.

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -33,7 +33,7 @@
 
 import { URL, URLSearchParams } from 'url';
 
-import { BotEndpointOptions, SpeechTokenInfo } from '@bfemulator/sdk-shared';
+import { BotEndpointOptions, SpeechTokenInfo, SpeechRegionToken } from '@bfemulator/sdk-shared';
```
compulim: Sorting.

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -48,6 +48,7 @@ export default class BotEndpoint {
   public speechToken?: string;
```
compulim: Is this one still being used?

packages/emulator/core/src/facility/botEndpoint.ts
```diff
@@ -92,6 +100,28 @@ export default class BotEndpoint {
     }
   }
 
+  public async getSpeechToken(refresh: boolean = false): Promise<SpeechRegionToken> {
+    if (!this.msaAppId || !this.msaPassword) {
+      throw new Error('bot must have Microsoft App ID and password');
+    }
+
+    if (this.speechRegionToken && !refresh && !this.willTokenExpireWithin(0)) {
```
compulim: This "if-loop" is too complicated than it is needed.

Pseudo code:

```
if (!this.cachedToken || Date.now() > this.cachedToken.expireAt) {
  this.cachedToken = await getToken();
}

return this.cachedToken;
```

packages/app/client/src/state/sagas/chatSagas.ts
```diff
@@ -169,9 +170,9 @@ export class ChatSagas {
...
     }
 
-    if (speechRegionToken) {
+    if (speechRegionToken && speechRegionToken.accessToken) {
       const factory = yield call(createCognitiveServicesSpeechServicesPonyfillFactory, {
-        authorizationToken: speechRegionToken.access_Token,
+        authorizationToken: Promise.resolve(speechRegionToken.accessToken),
```
tonyanziano: This is a step in the right direction, but I think what we want is for the Promise to be a function that goes and fetches the token every time (either from the service endpoint or from some cache on the Emulator side):

```
const factory  = yield call(createCognitiveServicesSpeechServicesPonyfillFactory, {
  authorizationToken: async () => {
    // add caching mechanism here since the function is called every time

    const { token } = await fetch(<token-endpoint>).json();

    return token;
  },
  region: speechRegionToken.region,
});
```

The region can stay the same for the Emulator session (this shouldn't matter at the moment because our speech service is deployed to one region, and will need to be changed once the fix goes into Web Chat to change the promise signature).

packages/emulator/core/src/facility/botEndpoint.ts
bwamie: Renamed to `speechAuthenticationToken `

packages/app/client/src/state/sagas/chatSagas.ts
```diff
@@ -168,9 +174,11 @@ export class ChatSagas {
     } catch (e) {
       // No-op - this appId/pass combo is not provisioned to use the speech api
     }
-    if (token) {
-      const factory = yield call(createCognitiveServicesBingSpeechPonyfillFactory, {
-        authorizationToken: token,
+
+    if (speechAuthenticationToken && speechAuthenticationToken.access_Token && speechAuthenticationToken.region) {
+      const factory = yield call(createCognitiveServicesSpeechServicesPonyfillFactory, {
+        authorizationToken: () => Promise.resolve(speechAuthenticationToken.access_Token),
```
compulim: This should a Promise function. When called, will fetch the token from the server. And the token need to be cached for its half-life.

packages/emulator/core/src/facility/botEndpoint.ts
compulim: Are you going to remove it if it is not used? If it is used, where did it get used?

bwamie: line 107

packages/app/client/src/state/sagas/chatSagas.ts
```diff
@@ -157,23 +157,24 @@ export class ChatSagas {
...
-        authorizationToken: token,
+
+      const factory = yield call(createCognitiveServicesSpeechServicesPonyfillFactory, {
+        authorizationToken: speechAuthenticationToken,
+        region: 'westus2', // Currently, the prod speech service is only deployed to westus2
       });
+
       yield put(webSpeechFactoryUpdated(documentId, factory)); // Provide the new factory to the store
+    } catch (e) {
+      // No-op - this appId/pass combo is not provisioned to use the speech api
```
tonyanziano: I think @compulim 's comment about how this `catch` could be hit due to a saga error (`call` or `put`) still stands. However, we currently have no way of distinguishing between the two types of errors.

compulim: It's a pain but okay to ignore the error for now. 😉

packages/emulator/core/src/authEndpoints.ts
```diff
@@ -57,6 +57,6 @@ export const v32Authentication = {
 };
 
 export const speech = {
-  // Access token for Bing Speech Api
-  tokenEndpoint: 'https://login.botframework.com/v3/speechtoken',
+  // Access token for Cognitive Services API
+  tokenEndpoint: 'http://login.botframework.com/v3/speechtoken/speechservices',
```
compulim: Secured endpoint?

```suggestion
  tokenEndpoint: 'https://login.botframework.com/v3/speechtoken/speechservices',
```

tonyanziano: I hope so 😮

cwhitten:
Nice work @edwin4microsoft & @tonyanziano. Let's resolve conflicts and get it in.

PR #1894, opened 2019-10-02
[ISSUE#1819][MAS4.1.2] [Screen Reader-Debug] Screen reader is not reading the Name for “Previous and Next” arrow buttons

denscollo:
Solves #1819

### Description
This pull request enables the proper reading of the “Previous and Next” arrow buttons

### Changes made
- Update aria-label to be more descriptive
- Add aria label to show/hide diff button


tonyanziano:
I just pulled this down and the narrator reads (for the Previous button):

> previous button bot state tooltip previous button previous

In the original issue, the a11y team says that it should only narrate as:

> previous button bot state tooltip previous

I think you could even add a comma before "tooltip previous" so that there is a break in the narrator's speech.

However, I don't think that there's a way we can achieve what the a11y team wants for several reasons:

1. The narrator will always read `"button"` after the `aria-label` is read because it is a `<button>` element
1. The narrator will always read the `title` attribute on the button after `"<aria-label> button"` is read

So this means that if we want to use the `title` attribute, which we do, we will always have the following order of words read: `<aria-label> button <title attribute>`. The closest we can get to this would be:

> previous bot state button previous

I'll contact the a11y team and see what we can do about this without having to create some sort of custom component.

tonyanziano:
@denscollo I spoke to the a11y team about this and here is their response:

> Yes It is fine if the screen reader is reading "Previous bot state button previous" for the control.

This means that the button would need:

1. `aria-label="Previous bot state"`
2. `title="Previous"`

and it should read `"Previous bot state button previous"`

You can apply this same pattern to the other buttons:

ex. `aria-label="Show diff" title="Show diff"` ---> `"Show diff button show diff"`

denscollo:
@tonyanziano thanks Tony! I'll apply the feedback and have it ready in a bit.

Issue #1806, opened 2019-08-31
[MAS4.2.10][Screen reader-File->New bot configuration] In browse mode nvda is reading the at a time for Hide copy controls information


Kumar2608:
**Actual Result:**
In browse mode NVDA is reading the at a time for Hide copy controls information
In browse mode NVDA is reading as “Hide link and Copy link “when the focus is on Hide link and focus is not moving to the copy link while press the down arrow key 

Expected Result:
In browse mode NVDA should read as hide link when the focus is on Hide link after press the down arrow key focus should moves to the copy link and nvda should read as Copy link 

**Repro Steps:**
Turn on the NVDA 
1.Open the Bot Framework V4 Emulator.
2. Navigate to the File menu and select it
3.File menu gets opened  
4.Navigate to the New bot configuration and select it 
5.New bot configuration dialog gets opened 
6.Naviage to the Hide and copy controls using Browse mode (Down arrow key)
7.Observe the issue 

**User Impact:**
Users who rely on screen reader will be affected if the screen reader reading the at a time for both controls info then the screen reader users will get confuse.

**Test Environment:**
OS Build: Version 1903(OS Build 18362.239)
Application: Bot Framework V4 Emulator
Screen reader: NVDA

**MAS Reference Link:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={b24201d1-2276-4471-8183-e1e5353031ad}
[MAS4.2.10_File.zip](https://github.com/microsoft/BotFramework-Emulator/files/3562177/MAS4.2.10_File.zip)

Kumar2608:
Verified in latest bot emulator version 4.5.2-80829 with NVDA(2019.2) issue still repro'ing. Hence we are reactivating this bug.
Please refer the attachment:1806.zip
[1806.zip](https://github.com/microsoft/BotFramework-Emulator/files/3676274/1806.zip)

corinagum:
Investigation notes:

- Only repros when navigating via Ctrl + DownArrow in browse mode (no repro via tab)
- Clicking anywhere in the dialog that is not interact-able when NOT in browse mode reads the text **"If you don't have a bot configuration, button create a new bot configuration".** - Note that this is the text/button that originally opened the dialog.
- When Encrypt keys checkbox is *not checked*, the following is read in browse mode:
  - Encrypt keys stored in your bot configuration. 
  - Your keys are not encrypted
[...]

PR #1921, opened 2019-10-09
#1806 Fix AT readout of botCreationDialog secret buttons

corinagum:
#1806 

image

Various minor fixes
- Adds an aria-label "Bot encryption key" to the secret's textfield
- Changes placeholder text (which is read by the AT) to read "Your keys are not encrypted", `encryptKeys` boolean is true
- adds `aria-label` to the `<TextField>` component.
- Styling fixes for the Hide Secret and Copy Secret buttons, which also fixes the way the AT reads both buttons out in browse mode

### DIFF
--- a/packages/app/client/src/ui/dialogs/botCreationDialog/botCreationDialog.tsx
+++ b/packages/app/client/src/ui/dialogs/botCreationDialog/botCreationDialog.tsx
@@ -187,12 +187,12 @@ export class BotCreationDialog extends React.Component<BotCreationDialogProps, B
           </Row>
           <Row align={RowAlignment.Bottom} justify={RowJustification.Left}>
             <TextField
-              ariaLabel="Bot encryption key"
+              aria-label="Bot encryption key"
               inputContainerClassName={dialogStyles.key}
               inputRef={this.setSecretInputRef}
               label="Secret "
               value={secret}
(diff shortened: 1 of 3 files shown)


packages/app/client/src/ui/dialogs/botCreationDialog/botCreationDialog.tsx
```diff
@@ -187,11 +187,12 @@ export class BotCreationDialog extends React.Component<BotCreationDialogProps, B
           </Row>
           <Row align={RowAlignment.Bottom} justify={RowJustification.Left}>
             <TextField
+              ariaLabel="Bot encryption key"
               inputContainerClassName={dialogStyles.key}
               inputRef={this.setSecretInputRef}
               label="Secret "
               value={secret}
-              placeholder="Your keys are not encrypted"
+              placeholder={encryptKey ? 'Key encryption' : 'Your keys are not encrypted'}
```
tonyanziano: This can just be:

`encryptKey ? '' : 'Your keys are not encrypted'`

When the encryption checkbox is checked, we automatically generate a secret and populate the text field, so there is never a scenario in which encryption is enabled and the textbox is empty. So the placeholder value will never show or be read in that case.

PR #1926, opened 2019-10-10
Add Dynamic Speech Services Region Support

bwamie:
Add flexibility to us region returned by Bot Service speech token service. 
This will help scale to other regions without having to make changes to the emulator.
Incase one Speech Services region is down (not able to get speech service tokens), Bot Service should be able to divert traffic to other regions. 

Incase these changes are acceptable, I will go ahead and modify the tests accordingly.

### DIFF
--- a/packages/app/client/src/state/sagas/chatSagas.ts
+++ b/packages/app/client/src/state/sagas/chatSagas.ts
@@ -159,18 +159,21 @@ export class ChatSagas {
     const { GetSpeechToken: command } = SharedConstants.Commands.Emulator;
 
     try {
-      const speechAuthenticationToken: Promise<string> = ChatSagas.commandService.remoteCall(
+      const speechAuthenticationToken = yield call(
+        [ChatSagas.commandService, ChatSagas.commandService.remoteCall],
         command,
         endpoint.id,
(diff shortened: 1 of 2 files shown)


packages/app/client/src/state/sagas/chatSagas.ts
```diff
@@ -159,18 +159,21 @@ export class ChatSagas {
     const { GetSpeechToken: command } = SharedConstants.Commands.Emulator;
 
     try {
-      const speechAuthenticationToken: Promise<string> = ChatSagas.commandService.remoteCall(
+      const speechAuthenticationToken = yield call(
+        [ChatSagas.commandService, ChatSagas.commandService.remoteCall],
         command,
         endpoint.id,
         !!existingFactory
       );
```
tonyanziano: This will no longer be a promise that refreshes the token.

I understand that this is meant to allow us to hit multiple regions once the speech service is scaled out, but the plan was to wait for the Web Chat fix that allows us to provide a promise that returns both the region _and_ the token together.

I've opened an issue on Web Chat to track that: https://github.com/microsoft/BotFramework-WebChat/issues/2459

PR #1930, opened 2019-10-16
Fixed unnecessary Web Chat re-renders caused by interaction with activities

### LINKED ISSUE #1650: [DEV IMPROVEMENT] Consider electron-devtools-installer
Consider [Electron Dev Tools Installer](https://www.npmjs.com/package/electron-devtools-installer) to install react-dev-tools and redux-dev-tools for a better debug experience.

### LINKED ISSUE #1701: Adaptive Card content is cleared on submit
## Version
4.5.2

## Describe the bug
Since version 4.5.0 the adaptive cards are behaving incorrectly. Version 4.5.1 and 4.5.2 have fixes for the issues #1680 and #1683 but the experience is still not at the same level as 4.4.2.  

On 4.5.2 the form content is cleared on submit. Sometimes the data are correctly sent back to the bot, sometimes not. 

## Screenshots
See the difference between 4.4.2 and 4.52

### Version 4.4.2
emulator-4 4 2

### Version 4.5.2
emulator-4 5 2

## Expected behavior
Form value must not be cleared on Adaptive Card

[bug]

tonyanziano:
Fixes #1701 

===

Web Chat was being completely re-rendered, which was causing Adaptive Card state to be wiped. The following interactions with Web Chat would cause this:

1. Sending messages in the chat
2. Clicking on activities and selecting / highlighting them

The fix prevents these re-renders by fixing the way app state was being plumbed through the components.

The entire `document` object is no longer plumbed down through the `emulator.tsx` component and its children. Instead, `documentId` is plumbed through these components, and then each component uses this id to fetch whatever relevant information from the store that it needs.

---

Also added `electron-devtools-installer` which resolves #1650 , and adds React Dev Tools as well as Redux Dev Tools.

===

Still need to add / fix tests.


tonyanziano:
Don't merge this yet, I think there might be a bug with the activity middleware.

Issue #1897, opened 2019-10-03
[MAS4.2.11][Screen Reader- Create new bot configuration] After "Save and connect", the focus goes on the hidden element and announced as "secret copied to the clipboard".


Kumar2608:
**Actual Result:**
After "Save and Connect" button i.e. the last element of the dialog, the focus goes on the hidden element and announced "secret copied to the clipboard".

**Expected Result:**
The voiceover focus shouldn't lands on the hidden element or it shouldn't announced the status message again and again.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened.
3. Navigate to the "Create new bot configuration" button and activate it.
4. Open a "Create new bot configuration" dialog box gets opened.
5. Navigate to the "encrypt keys stored in your bot configuration." checkbox and check it.
6. Navigate to "copy" button and activate it.
7. Now, navigate through the pane and observe the issue after "Save and Connect" button.

**User Impact:**
If unnecessary information is announced after the control, then voiceover dependent users will get confused.

**Test Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

**Impacted MAS Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={eedc2941-f754-4739-b483-3f8180c5aaa5}](url)

**Attachment:**
[Focus goes on hidden element.zip](https://github.com/microsoft/BotFramework-Emulator/files/3685398/Focus.goes.on.hidden.element.zip)

ShikhaMishra11:
Verified ,issue is fixed now so closing the bug.

PR #1934, opened 2019-10-17
[ISSUE1897] [MAS4.2.11][Screen Reader- Create new bot configuration] After "Save and connect", the focus goes on the hidden element and announced as "secret copied to the clipboard

denscollo:
Solves #1897

### Description
This pull request fixes how the focus behaves  after clicking the "Save and Connect" button, no longer sending the focus on the hidden element and announcing "secret copied to the clipboard"

### Changes made
The problem was caused by the button, when clicked it was creating a new tag with the role alert and adding it to the body, which can then be browsed with the keyboard.
To solve this we added a timeout to this element, getting removed from the body after the notification is done.

image

### Tests
AlertElementIssue1897

### DIFF
--- a/packages/app/client/src/ui/a11y/ariaAlertService.ts
+++ b/packages/app/client/src/ui/a11y/ariaAlertService.ts
@@ -50,6 +50,10 @@ class AriaAlertService {
     alert.setAttribute('role', 'alert');
     alert.setAttribute('style', 'position: absolute; top: -9999px; overflow: hidden;');
     document.body.appendChild(alert);
+
+    setTimeout(() => {
+      document.body.removeChild(alert);
+    }, 1000);
   }
 }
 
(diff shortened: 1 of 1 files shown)


tonyanziano:
Is there any way we can hide the alert from being focused by giving it a `tab-index` of -1?

If you look at the original PR in which this alert service was added, @corinagum and I [had a discussion about whether or not to remove the alert](https://github.com/microsoft/BotFramework-Emulator/pull/1873#discussion_r325883220)  from the DOM after a period of time.

It looks like the suggested model is to leave the alert until another is created to replace it.

Also, was the duration of 1 second tried out with long alert messages? If the alert is long enough, will the screen reader be cut off when the alert is removed from the DOM?

denscollo:
@tonyanziano we checked this using the `tabindex = "-1"` and it doesn't work as expected.
We read the comments on the issue and we found a way to achieve this without removing the element, which is by setting the `aria-hidden` attribute in the timeout function. 
As for the duration of the time-out, we set it up to 5000 secs as you suggested.

The image below shows the tab index test, using the VO shortcut `ctrl + left/right arrows` (it stills reads the span element).
imagen

We just pushed the update with the aria-hidden attribute.

Issue #1896, opened 2019-10-03
[MAS1.3.1][Screen Reader-Connect the bot] Incorrect position of the controls are announced by voiceover for the buttons available at the top most menu bar.


Kumar2608:
**Actual Result:**
When focus lands on the top menu bar voiceover announced "Restart Conversation", "Save Transcript" etc. But the position of the buttons are not correctly announced.

**Expected Result:**
The voiceover should announce the correct position of the buttons.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened.
3. Navigate to the "Open Bot" button and activate it.
4. Open a Bot Dialog box gets opened.
5. Navigate to the "bot URL" edit box and enter URL- http://localhost:3978/api/messages
6. Navigate to "Connect" button and activate it.
7. "Live Chat" separate window appears.
8. Navigate to "Menu Bar" having controls i.e. "Restart Conversation", "Save Transcript" etc.
9. Observe the issue.

**User Impact:**
Users who depends on the Screen reader will find the difficulty to locate the position of the buttons, if the information is provided by voiceover is not correct.

**Test Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

**Impacted MAS Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={54f28d1f-a2d1-4dcd-84e1-5c9b87e8aba4}](url)

**Attachment:**
[Incorrect position of buttons.zip](https://github.com/microsoft/BotFramework-Emulator/files/3685236/Incorrect.position.of.buttons.zip)

ShikhaMishra11:
The issue still repro's on Mac. The voiceover still announces incorrect position of the button.
Please find the latest attachment for the reference.

Environment:
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

Attachement: 
[Bug1896.zip](https://github.com/microsoft/BotFramework-Emulator/files/3767717/Bug1896.zip)

denscollo:
Hi @ShikhaMishra11, we just tested this on Mac and it seems to be fixed. We used the lastest Emulator release.

Environment:
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-85071)

ShikhaMishra11:
[...]

PR #1935, opened 2019-10-21
[ISSUE#1896][MAS1.3.1][Screen Reader-Connect the bot] Incorrect position of the controls are announced by voiceover for the buttons available at the top most menu bar.

denscollo:
Solves #1896

### Description
Fixes how voiceover announces the position of the editor menu bar's buttons controls

### Changes made
We added an `aria-hidden` attribute to the separator bar between the `restart conversation` button and the sub-menu button. This way voiceover skips this element which was causing the problem.

Additionally, we noticed that in windows, the last element of the menu bar was not being announced as 3/3. To solve this we removed a `div` around the `save transcript` button and moved the role label `menuitem` inside it.

### Testing
In the next image, you can see the voiceover reading the order of the elements as intended.

asda


denscollo:
Thanks @tonyanziano ! I justed pushed the updated changelog

PR #1950, opened 2019-10-29
Handle list of hosts/domains in NO_PROXY

### LINKED ISSUE #1947: NO_PROXY is being treated as if it were a single hostname
## Version
4.5.2

## Describe the bug

Currently, when checking whether a given URL should go through the proxy, the code treats the `NO_PROXY` environment variable as if it's value represents a single host, and tests if that value is within the URL:

https://github.com/microsoft/BotFramework-Emulator/blob/90fb9d91f7ff16a4d53f10089ad795564d758a9c/packages/app/main/src/fetchProxy.ts#L58-L60

However, the `NO_PROXY` variable is [a comma-separated list of host names or domains](https://www.gnu.org/software/emacs/manual/html_node/url/Proxies.html). So, when `NO_PROXY` contains more than a single value, the check fails and the tested URL is always sent through the proxy.

This could be fixed with something like:

```typescript
// No Proxy

if (!process.env.HTTPS_PROXY) {
  return nodeFetch(...args);
}

const url: URL = new URL(typeof urlOrRequest === 'string' ? urlOrRequest : urlOrRequest.url);

const noProxyList = (process.env.NO_PROXY || '')
  .split(',')
  .map(x => x.trim().toLowerCase().replace(/^\*/, ''))
  .filter(x => x);

if (noProxyList.some(x => x.startsWith('.') ? url.hostname.endsWith(x) : url.hostname === x)) {
  return nodeFetch(...args);
}
```

I can create a PR for you, if you want.

## To Reproduce
1. Set up a proxy, defining the `HTTP_PROXY` and `HTTPS_PROXY` as appropriate.
2. Set `NO_PROXY` to contain more than 1 hostname (e.g. `localhost,127.0.0.1`)
3. Try to connect to a bot URL that is included in the `NO_PROXY` list

## Expected behavior
The emulator should not use the proxy and connect directly.

[bug]

poveden:
Fixes #1947


tonyanziano:
Thanks for the PR @poveden. We will review this once we publish version v4.6.0 of the Emulator, which should happen some time today.

tonyanziano:
Looks good to me. I'm going to update the branch so CI kicks in, and if it goes green I'll merge it in.

Thanks for the contribution 🙂

Issue #1903, opened 2019-10-04
[MAS4.2.5] [Labels and Relationships - Settings] Voiceover announce the same label for “Path to ngrok", "Locale", "Localhost Override"and "User ID" present in the "settings" pane.


Kumar2608:
**Actual Result:**
In settings pane, the voiceover announces the incorrect labels for edit fields i.e. "Path to ngrok", "Locale", "Localhost Override" and "User ID" whereas the labels are correctly defined in the code.
Voiceover announces "Enter User ID" edit text for all the edit fields.

**Expected Result:**
The voiceover should announce the label which is associated with the edit fields so that user can understand the purpose & provides valid input.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened.
3. Navigate to the "Open Bot" button and activate it.
4. Open a Bot Dialog box gets opened.
5. Navigate to the "bot URL" edit box and enter URL- http://localhost:3978/api/messages
6. Navigate to "Connect" button and activate it.
7. "Live Chat" separate window appears.
8. Navigate to "edit ngrok settings" controls available at right lower corner of the pane and activate it.
9. Observe the issue.

**User Impact:**
It will be difficult for visually impaired user to enter the required input, if the text box label's are not properly announced.

**Test Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

**Impacted MAS Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={07857fe1-d8ef-409e-a594-f440a8359566}](url)

**Attachment:**
[Voiceover announce the same label for edit boxes.zip](https://github.com/microsoft/BotFramework-Emulator/files/3689467/Voiceover.announce.the.same.label.for.edit.boxes.zip)

ShikhaMishra11:
Checked and verified. Issue resolved in new version of Bot Emulator. Hence, closing the issue.

Test Environment:
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.6.0-89098)

PR #1970, opened 2019-11-05
[ISSUE#1903][MAS4.2.5] [Labels and Relationships - Settings] Voiceover announce the same label for “Path to ngrok", "Locale", "Localhost Override"and "User ID" present in the "settings" pane

denscollo:
Solves #1903

### Description
Fixes how _Voiceover_ announces the TextFields' labels in the Settings editor pane. This error was extended to TextFields in other components as well.

### Changes made
We added a condition to the setting of the `aria-labelledby` attribute in the TextField component. This way, _Voiceover_ will read the error message only when related to the rendered element, otherwise, it will read the element's label.

Additionally, we noticed that this issue was happening on windows as well. The same solution applies to both platforms.

### Testing
In the following images, you can see how _Voiceover_ reads the labels instead of the error message:

image
image
image

### DIFF
--- a/packages/sdk/ui-react/src/widget/textField/textField.tsx
+++ b/packages/sdk/ui-react/src/widget/textField/textField.tsx
@@ -71,7 +71,7 @@ export class TextField extends Component<TextFieldProps, {}> {
       <div className={`${styles.inputContainer} ${inputContainerClassName}`}>
         {this.labelNode}
         <input
-          aria-labelledby={errorMessage ? 'errormessagesub' : 'label'}
+          aria-labelledby={errorMessage ? 'errormessagesub' : undefined}
           className={inputClassName}
           id={this.inputId}
           ref={this.setInputRef}
--- a/packages/sdk/ui-react/src/widget/textField/textField.tsx
+++ b/packages/sdk/ui-react/src/widget/textField/textField.tsx
@@ -71,7 +71,7 @@ export class TextField extends Component<TextFieldProps, {}> {
       <div className={`${styles.inputContainer} ${inputContainerClassName}`}>
         {this.labelNode}
         <input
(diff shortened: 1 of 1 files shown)


packages/sdk/ui-react/src/widget/textField/textField.tsx
```diff
@@ -71,7 +71,7 @@ export class TextField extends Component<TextFieldProps, {}> {
       <div className={`${styles.inputContainer} ${inputContainerClassName}`}>
         {this.labelNode}
         <input
-          aria-labelledby="errormessagesub"
+          aria-labelledby={errorMessage ? 'errormessagesub' : 'label'}
```
tonyanziano: I think that this should be:

`aria-labelledby={errorMessage ? 'errormessagesub' : undefined}`

[`aria-labelledby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) requires

>A space-separated list of element IDs

that are supposed to describe the element that the attribute has been attached to. Your change is saying that the `<input>` tag should be describe by some element with `id="label"`, however, if you look at the HTML markup for the `.labelNode` of the `<TextField>` component, it does not have an ID associated with it.

So what I think is actually happening here is that the `aria-labelledby` attribute tries to find the associated element with `id="label"`, fails, and then falls back to default accessibility behavior, which is to look at the nearest `<label>` tag with the `for="<input id>"` attribute assigned. This is the correct behavior.

I've tried changing it to `aria-labelledby={errorMessage ? 'errormessagesub' : undefined}` on Windows, and the behavior remains the same as your change.

I would try it out on Mac, and if it works, then we are good to merge it in. 👍

Issue #1914, opened 2019-10-07
[MAS2.1.1][Keyboard-Live chat dialog] "Root" list box present in "JSON" pane are not expanded or collapsed by keyboard.


Kumar2608:
**Actual Result:**
When we click "Root" list box it will not expanded or collapsed whereas same can be accessible by mouse.

**Expected Result:**
It should expanded or collapsed as it can be done by mouse.

**Note:**
1. Same issue observed on the Linux system. Please refer "JSON pane not expanded collapsed by key_Linux" attachment.
2. Same issue observed for "Conversation", "From" and "Recipient" list box available under "Root" list box.

**Observation:**
1. We already try to activate it with various shortcut keys but not be able to activate the same.
2. In Linux system the pane is expanded by left/right arrow.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened.
3. Navigate to the "Create new bot configuration" button and activate it.
4. "Create new bot configuration" dialog box gets opened.
5. Fill the required fields like bot name and endpoint URL example: http://localhost:3978/api/messages
6. Navigate to "Save and Connect" button and activate it.
7. Now, navigate through the pane and observe the issue after "Save and Connect" button.
8. "Live Chat" screen will appear.
9. Navigate to "[ConversationUpdate event detected]" button and activate it.
10. Observe the change in "JSON" tab and navigate to the tab.
11. Try to collapse "Root" list box by keyboard.
12. Observe the issue.

**User Impact:**
The keyboard dependent user will get confused as they will not be able to knew the presence of content or controls in the "Root" or other list box.

**Test Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

OS: Ubuntu 18.04.3 LTS
Application: BotFramework-Emulator-4.5.2-linux

**Impacted MAS Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={8492c4eb-c179-40ae-8777-cd044ed725a2}](url)

**Attachment:**
[not expanded or collapsed by keyboard.zip](https://github.com/microsoft/BotFramework-Emulator/files/3696197/not.expanded.or.collapsed.by.keyboard.zip)
[...]

PR #1988, opened 2019-11-13
[ISSUE#1914][MAS2.1.1][Keyboard-Live chat dialog] "Root" list box present in "JSON" pane are not expanded or collapsed by keyboard.

denscollo:
Solves #1914

### Description
Fix the capability to expand or collapse the root item on JSON viewer pane using the keyboard. 

### Changes made
Set the JSON tree view container _tabIndex_ value to "-1" to allow the navigation from the copy JSON Button to the root item.

### Testing
We test the solution in both Mac and Linux OS.
JsonWebViewer5

image

### DIFF
--- a/packages/sdk/ui-react/src/widget/collapsibleJsonViewer/collapsibleJsonViewer.tsx
+++ b/packages/sdk/ui-react/src/widget/collapsibleJsonViewer/collapsibleJsonViewer.tsx
@@ -135,7 +135,7 @@ export class CollapsibleJsonViewer extends Component<CollapsibleJsonViewerProps,
       ref.firstElementChild.setAttribute('role', 'tree');
       // <ul><li>
       this.nodesAdded(ref.firstElementChild.childNodes);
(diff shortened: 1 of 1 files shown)


tonyanziano:
After looking at the original issue, I'm not sure that the problem here is that the `root` level is not accessible via keyboard (if you watch the uploaded video for Mac, you can see the user tab into the root level of the JSON tree).

I think that the issue here is that the node is not expandable via the appropriate keyboard key, which I'm assuming is Space Bar? Right now, we can expand / collapse nodes in the JSON tree using either right or left arrow, however, Space and Enter do nothing, which I think is what the a11y team expects to expand / collapse the nodes.

I'll confirm with them that what I described is the actual issue, and if so, what button should expand or collapse nodes.

tonyanziano:
A11y team confirmed that left and right arrow keys are fine for expanding / collapsing the nodes, and that the control is indeed accessible.

Closing the PR since there is no issue.

Issue #1904, opened 2019-10-04
[MAS2.4.3] [Focus Order - Endpoint] Focus order is not logical under "Azure bot service configuration" dropdown.


Kumar2608:
**Actual Result:**
Focus goes in the edit fields under "Azure Bot service Configuration", whereas the dropdown is not expanded. After "Azure Bot Service Configuration" the focus lands on "Azure BotId" and other edit fields.

**Expected Result:**
After "Azure Bot Service Configuration", the focus should move to "Cancel" button when the dropdown are not expanded.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened.
3. Navigate to the "Endpoint" and click on "+".
4. "Add an Endpoint for your Bot" dialog box appears.
5. Navigate with keyboard.
6. Observe the issue.

**Note:**
Same issue also observed on Linux, please refer the attachment "Focus order is not logical_Linux.zip"

**User Impact:**
User who rely on keyboard would be impacted as focus order is not in the ideal sequential order.

**Test Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

OS: Ubuntu 18.04.3 LTS
Application: BotFramework-Emulator-4.5.2-linux

**Impacted MAS Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={0de7fbe1-ad7e-48e5-bcbb-8d986691e2b9}](url)

**Attachment:**
[Focus order is not logical.zip](https://github.com/microsoft/BotFramework-Emulator/files/3689762/Focus.order.is.not.logical.zip)

[Focus order is not logical_Linux.zip](https://github.com/microsoft/BotFramework-Emulator/files/3793183/Focus.order.is.not.logical_Linux.zip)

awalia13:
@Kumar2608 -  Please validate this bug.

ShikhaMishra11:
**Comment**
Issue partially fixed, as the focus still goes within the drop-down in voice over mode.
Please find the attachment: Focus order is not logical_retest

**Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version: 4.6.0-89961)

**Attachment:**
[Focus order is not logical_retest.zip](https://github.com/microsoft/BotFramework-Emulator/files/3855309/Focus.order.is.not.logical_retest.zip)

[...]

PR #1994, opened 2019-11-19
[ISSUE#1904][MAS2.4.3] [Focus Order - Endpoint] Focus order is not logical under "Azure bot service configuration" dropdown.

denscollo:
Solves #1904

### Description
Fixes the VO navigation when the Azure Bot Service Configuration is collapsed.

### Changes made
We added an `aria-hidden` attribute to the `abs-config-content` component that will be true or false if the content is collapsed or expanded. 

### Testing
abs-config


denscollo:
Thanks @tonyanziano ! We just updated the changelog  😄

Issue #1889, opened 2019-10-01
[MAS4.2.10][Screen Reader-Open bot dialog] Keyboard focus and voiceover focus is not synchronized.


Kumar2608:
**Actual Result:**
When user navigates in the voiceover mode, the keyboard focus and voiceover focus is not synchronized at "Buttons" and "Edit fields".

**Expected Result:**
The focus should be synchronized with voice over focus so that all the changes are done by user itself.

**Note:**

1. Same issue found on "My Bots" section. {Refer attachment} (Scenario: "Welcome Tab screen") .
2. Same issue found on "How to build a Bot" section. {Refer attachment} (Scenario: How to Build a Bot".
3. Same issue found on "Connect your Bot to LUIS Application" dialog.{Refer attachment} (Scenario: Add a LUIS app manually)
4. Same issue found at "Browse" button as "Microsoft App Id" and "Microsoft App Password" edit fields in "Open a Bot" dialog. {Refer attachment} (Scenario: Open Bot Dialog)
5. Same issue found at "Microsoft App Id" and "Microsoft App Password" edit fields in "New bot configuration" dialog. {Refer attachment} (Scenario: Create new bot configuration) 
6. Same issue found on the edit fields present at "Settings" and "Sign in with Azure account dialog". {Refer attachment} (Scenario: Settings and Sign in with Azure account dialog)
7. Same issue found at link present at "Add language understanding" and "Add QnA maker" pop-up. {Refer attachment} (Scenario: Add language understanding and Add QnA maker)

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened.
3. Navigate to the "Open Bot Dialog" button and activate it.
4. Open Bot Dialog box gets opened.
5. Navigate to the dialog box in voiceover mode 
6. Observe the issue.

**User Impact:**
The voiceover focus and keyboard focus should be synchronized so that all the changes done in the pane is knew by the user. Accidentally changes in the pane can be prevented.

****Test Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

**Impacted MAS Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={b24201d1-2276-4471-8183-e1e5353031ad}]

**Attachment:**
[...]

PR #1995, opened 2019-11-19
[ISSUE#1889][MAS4.2.10][Screen Reader-Open bot dialog] Keyboard focus and voiceover focus is not synchronized.

denscollo:
Solves #1889

### Description
This pull request fixes the navigation with **Voiceover**, now it's cursor is synchronized with the keyboard in the following places:

- "My Bots" section.
- "How to Build a Bot" section.
- "Connect your Bot to LUIS Application" dialog.
- "Open a Bot" dialog.
- "New bot configuration" dialog.
- "Add language understanding" and "Add QnA maker" pop-up.
- "Settings" section.
- "Sign in with Azure account" dialog.

### Changes made
To fix the behavior of _TextFields_ we removed the styling of `position: relative`, now the **Voiceover**'s cursor navigates normal thought these elements, without jumping to other places.

To fix the _LinkButtons_ we added the role _presentation_ to the paragraph containing these elements and to other containers around these paragraphs.

The _openBotDialog_ dialog has two custom elements, an _AutoComplete_ to which we added the _presentation_ role and to the form that contains it, and the _browse_ button, for which we added the `aria-hidden="true"` attribute to its label and made some styling changes so the **Voiceover** cursor is in the right place.

### Testing
In the following recordings, you can see how these changes fixes the **Voiceover**'s cursor following the keyboard's focus.
howtobuildabot
connectto
openabot


tonyanziano:
I went back to the original issue and asked if the issue can be re-evaluated, because I'm unable to reproduce any of these focus synchronization issues using VoiceOver on Mac 10.13.

Here's how VoiceOver navigates on my machine: 
out

I'm going to see if someone else on the team with a Mac can reproduce the focus issues using VoiceOver.

denscollo:
@tonyanziano are you using vO navigation to test it? (Ctrl + alt + arrow), also we are using MacOS v10.15 (we found that depending on the OS version, the behaviour can change).

tonyanziano:
@denscollo Oh no I haven't. I didn't know that was a thing 😐

I'll retry using VoiceOver navigation.

packages/app/client/src/ui/dialogs/openBotDialog/openBotDialog.scss
```diff
@@ -56,8 +57,7 @@
 .browse-button {
   & > input {
     position: absolute;
-    height: 0;
-    width: 0;
+    width: 100%;
```
tonyanziano: This shouldn't be necessary. I tested it with just the addition of the `position: relative` to the `.browse-button` class and it seems to be working.

packages/sdk/ui-react/src/widget/textField/textField.scss
```diff
@@ -46,7 +46,6 @@
 }
 
 .input-container {
-  position: relative;
```
tonyanziano: This is unnecessary. Positioning should not affect screen reader narration or focus order. It will affect visually where the focus border ends up, but in this case it should not matter.

I tested with and without this rule and the behavior was the same.

packages/sdk/ui-react/src/widget/autoComplete/autoComplete.tsx
```diff
@@ -82,7 +82,7 @@ export class AutoComplete extends Component<AutoCompleteProps, AutoCompleteState
       <div
         className={`${styles.autoCompleteContainer} ${invalidClassName} ${className}`}
         id={this.textboxId}
-        role="combobox"
+        role="presentation"
```
tonyanziano: This is unnecessary and breaks the combobox pattern that the `AutoComplete` component follows:

https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

denscollo: @tonyanziano We ran some tests and if we change the role to combobox, it won't navigate to this control, only to the Title of this control and then to the browse button (as seen on the attached gif below). Also, we reviewed the example that you sent us and it happens the same.

combobox

Do you want us to change it anyways?

tonyanziano: Hey @denscollo ,

Before I suggested the change I tested it, and it seemed to work fine. Here's an example of me testing it out this morning:

combobox

I wonder why it is behaving so much differently on your machine.

denscollo: @tonyanziano after reading the aria documentation you left, we noticed that the current implementation of the auto-completed component does not match the pattern. The <label> element is out of the combobox container. By updating the component to follow this pattern the issue was fixed. We'll be pushing the changes in a bit.

tonyanziano: Hi @denscollo ,

If you look at the sample code for the combobox on the w3 example I posted above, the DOM structure looks like this:

```
<div>       <!-- not important to accessibility, just for layout -->
  <div role="combobox">         <!-- container, important to pattern -->
    <input>                     <!-- text field, important to pattern -->
  </div>
  <ul>                          <!-- results, important to pattern -->
</div>
```

If you look at the DOM structure that currently exists before your changes:

```
<div role="combobox">         <!-- container, important to pattern -->
  <input>                     <!-- text field, important to pattern -->
  <ul>                        <!-- results, important to pattern -->
</div>
```

The only thing that we have excluded in the pattern is the outermost container, because we have wrapped the results in the `combobox` div instead of placing them outside, eliminating the need for the outermost div.

However, if you look at all the required roles, and attributes of the component, our DOM structure still satisfies those requirements.

The problem with your original changes at the top of this comment thread is that you changed the `role` of the combobox div to `role="presentation"`, which means that we no longer satisfy the component's role / attribute requirements.

I don't think it's necessary for us to add another container to the DOM structure, and mark it with `role="presentation"` because you are basically adding another element to the DOM and telling screen readers to ignore it. The current DOM implementation achieves the same exact layout without the extra `div`, and looks the same in the eyes of the screen reader.

Below are examples of VoiceOver's navigation over each of the implementations.

**Current implementation without extra `<div>`:**

old-dom


**Proposed implementation with extra `<div>`:**

new-dom


You can see that both implementations behave and read the same, so I don't think the DOM structure should be modified to include an extra `<div>` element.

I am going to revert the CSS changes, and change the DOM structure back to what it was, and merge this PR in for the sake of time.

I'm happy to continue discussing the reasoning behind this change with you offline if you'd like.

Issue #1890, opened 2019-10-01
[Functional - Open Bot Dialog] The focus goes on the disabled controls and functioning by space/return.


Kumar2608:
**Actual Result:**
In voiceover mode, the focus lands on the disabled control such as "Connect" button. The voiceover informs that the controls are disabled but if the user accidently press "spacebar" or "return" key the pane gets closed.

**Expected Result:**
As the controls are disabled, they shouldn't be active or accessed by any key.

**Note:**
Same issue observed at every disabled controls present throughout the application.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened.
3. Navigate to the "Open Bot Dialog" button and activate it.
4. Open Bot Dialog box gets opened.
5. Navigate to the dialog box in voiceover mode and select "Connect" disabled button.
6. Try to activate the button with "Spacebar" or "Enter".
7. Observe the issue.

**User Impact:**
If the disabled controls are accessible by keyboard the user might get confused.

**Test Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

**Attachment:**
[Disabled control are functioning.zip](https://github.com/microsoft/BotFramework-Emulator/files/3676372/Disabled.control.are.functioning.zip)

tonyanziano:
Is it recommended to completely hide disabled controls from the screen reader? Or is it better to have them be accessible by the screen reader, but not functional when pressed.

The latter case seems to make more sense to me because, wouldn't you want the non-sighted user to know about all the inputs on the screen? If a disabled buttons becomes enabled after filling out a form, how would the user know it has been enabled if they never knew it was there to begin with?

ShikhaMishra11:
@tonyanziano As i verified in the latest build of BotFramework Emulator, now the focus didn't lands on the disabled control. I guess it is fixed with #1910 

**Test Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (version: 4.6.0-92058)

tonyanziano:
Closing due to above comment.

PR #1999, opened 2019-11-21
[ISSUE#1890][Functional - Open Bot Dialog] The focus goes on the disabled controls and functioning by space/return.

denscollo:
Solves #1890

### Description
With these changes, the PrimaryButton and LinkButton elements will be hidden to screen readers while disabled.

In the issue report, the tester says the disabled buttons close the dialogs when pressing space, this happens because the Voiceover navigation mode goes through the elements even if they are not focusable by the keyboard, so the keyboard only follows it when it lands in a focusable element, and as a disabled button is not focusable, the keyboard focus remains in the last focusable element, in the case shown by the tester is the cancel button, so when pressing space bar or enter, the cancel button is clicked. This won't keep happening as **Voiceover** can't navigate to the Connect button when it is disabled

### Changes made
We added an attribute of `aria-hidden='true'` to the _PrimaryButton_ and _LinkButton_ elements when they have its `disabled` property set to true, this way the **Voiceover** screen reader skips them when navigating.

### Testing
In the following recording, you can see an example of a _LinkButton_ and a _PrimaryButton_ being hidden from **Voiceover** while they are hidden.

hiddenbuttons

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,7 +98,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - [1934](https://github.com/microsoft/BotFramework-Emulator/pull/1934)
   - [1935](https://github.com/microsoft/BotFramework-Emulator/pull/1935)
   - [1936](https://github.com/microsoft/BotFramework-Emulator/pull/1936)
- 
+  - [1999](https://github.com/microsoft/BotFramework-Emulator/pull/1999)
  - [client] Fixed an issue with the transcripts path input inside of the resource settings dialog in PR [1836](https://github.com/microsoft/BotFramework-Emulator/pull/1836)
  - [client] Implemented HTML app menu for Windows in PR [1893](https://github.com/microsoft/BotFramework-Emulator/pull/1893) 
  - [client/main] Migrated from Bing Speech API to Cognitive Services Speech API in PR [1878](https://github.com/microsoft/BotFramework-Emulator/pull/1878)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -98,7 +98,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - [1934](https://github.com/microsoft/BotFramework-Emulator/pull/1934)
   - [1935](https://github.com/microsoft/BotFramework-Emulator/pull/1935)
   - [1936](https://github.com/microsoft/BotFramework-Emulator/pull/1936)
- 
+  - [1999](https://github.com/microsoft/BotFramework-Emulator/pull/1999)
```
tonyanziano: this should go underneath the "Unreleased" header since 4.6.0 has already been released.

denscollo:
Hi @tonyanziano ,

What we understood that is happening is that Voiceover (navigation with VO + arrows), navigates through every visible element, even if it is not focusable by the keyboard. Because of this, the keyboard focus follows the VO navigation only when it lands on a place the keyboard can focus, resulting on this case, that when voiceover is placed in the save button, the keyboard focus remains in the cancel button, so when the user press space, he is actually pressing the cancel button.

We noticed that this behavior occurs everywhere, for example when navigating into a Checkbox and then to the text next to it, when pressing space, the Checkbox will be the one checked.
Because of this, to have disabled buttons still being visible to the user, we suggest making it focusable by the keyboard, so it can follow the voiceover navigation and if the user press space it won’t do any action.

Also, this behavior only happens with Voiceover (with VO + arrows). Other screen readers only navigate through visible objects focusable by keyboard, and disabling an element makes this object not focusable by default.

Issue #1881, opened 2019-09-26
[MAS2.1.2][Keyboard Navigation- Welcome Tab screen] Focus gets trapped on first and last element of the page when user navigates through the page.


Kumar2608:
**Actual Result:**
After last interactive element of the pane i.e. "View analytics" the focus should land on the first interactive element but the focus gets trapped on "View analytics".
Same issue observed when we navigate with "Shift + Tab" in backward navigation, the keyboard focus gets trapped at the first interactive element i.e. "Bot Explorer".

**Expected Result:**
After last interactive element when user press "Tab", the focus should land on the first element and vice-versa.

**Note:**
Same issue observed throughout the application.

**Repro Steps:**

1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened.
3. Navigate to the Welcome tab and reach at "View analytics".
4. Try to navigate further through tab key and observe the issue.

**User Impact:**
When a component can receive keyboard focus, but the user are not able to move away from the component using the keyboard. 
Users will not be able to access other components on the page and would be trapped on it.

**Test Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

**MAS Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame2.aspx?sourcedoc=%7bebdc3fe1-09a9-4aa2-a29b-e97764774e6c%7d](url)

**Attachment**
[MAS2.1.2.Keyboard.Navigation.zip](https://github.com/microsoft/BotFramework-Emulator/files/3658123/MAS2.1.2.Keyboard.Navigation.zip)

awalia13:
@tonyanziano  - We found this issue in MAC only.

ShikhaMishra11:
Fixed, Verified on MAC build 10.15 Beta(19A573a)

PR #2001, opened 2019-11-21
[ISSUE#1881][MAS2.1.2][Keyboard Navigation- Welcome Tab screen] Focus gets trapped on first and last element of the page when user navigates through the page

denscollo:
Solves #1881

### Description
Fixes how the focus gets trapped between the last and the first element of the pane.
 
### Changes made
We added a new condition to the **_EventHandlers_** class which will manage the focus when the emulator is running in _macOS_. Also, we included two auxiliary methods to find the first and last elements of the active pane and be able to move the focus between them.

Additionally, we added two unit tests to cover the new _keydown_ event.

### Testing
In the following image, you can see how the focus goes from the last element of the pane to the first when navigating with Tab key and vice-versa when using Shift+Tab navigation: 

FocusTrap


CHANGELOG.md
```diff
@@ -98,6 +98,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - [1934](https://github.com/microsoft/BotFramework-Emulator/pull/1934)
   - [1935](https://github.com/microsoft/BotFramework-Emulator/pull/1935)
   - [1936](https://github.com/microsoft/BotFramework-Emulator/pull/1936)
+  - [2001](https://github.com/microsoft/BotFramework-Emulator/pull/2001)
```
tonyanziano: this needs to be under the "Unreleased" header and not the "4.6.0" header

denscollo: Thanks for noticing! I just updated the changelog entry.

packages/app/client/src/utils/eventHandlers.ts
```diff
@@ -121,7 +135,25 @@ class EventHandlers {
         } as Notification);
       }
     }
+
+    if (isMac()) {
+      const tabPressed: boolean = key === 'tab' || keyCode === 9;
```
corinagum: I believe keyCode is deprecated. Can we remove the reference? key === tab should be sufficient.

packages/app/client/src/utils/eventHandlers.ts
```diff
@@ -121,7 +135,25 @@ class EventHandlers {
...
+      const lastDecendants = EventHandlers.getLastDecendants(document.querySelector('main'));
+      const firstElement = document.querySelector('nav').firstElementChild as HTMLElement;
```
corinagum: move to line 152

packages/app/client/src/utils/eventHandlers.ts
```diff
@@ -121,7 +135,25 @@ class EventHandlers {
...
+      const lastElement = lastDecendants[lastDecendants.length - 1] as HTMLElement;
```
corinagum: move to line 149

packages/app/client/src/utils/eventHandlers.ts
```diff
@@ -121,7 +135,25 @@ class EventHandlers {
         } as Notification);
       }
     }
+
+    if (isMac()) {
+      const tabPressed: boolean = key === 'tab' || keyCode === 9;
+      const lastDecendants = EventHandlers.getLastDecendants(document.querySelector('main'));
```
corinagum: Move to line 149

denscollo:
@corinagum if we move lines 142 & 143 to the ones that you told us, the `isFirstElement` and `isLastElement` variables are not going to work. If you want us to move those line, we need to update the if clause to be something like: 
`      if (shiftPressed && document.activeElement === document.querySelector('nav').firstElementChild as HTMLElement) {
    }`

And something similar would happen to the other if clause, since it uses lastDecendants to obtain the lastElement.

Let us know if you want us to proceed with this change.

Issue #1908, opened 2019-10-04
[MAS4.2.10] [Focus Order - Add other service] With voiceover consolidated focus goes on "Key and Value".


Kumar2608:
**Actual Result:**
With voiceover the consolidated focus lands on "Key and Value" whereas they are associated with different edit boxes.

**Expected Result:**
The individual focus should lands on them as well as the focus order should be logical, after "Key" the focus should land on the edit field associated with key. After which the focus should lands on "Value" and its associated edit field.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened.
3. Navigate to the "Service" and click on "+".
4. Select "Add other service" from the "Menu".
5. Navigate through the dialog.
6. Observe the issue.

**User Impact:**
Users who rely on screen readers will be affected if the screen reader announced "Key and Value" in a same flow, it may misguide the user.

**Note:**
Suggestion: The key and value are defined as a list but it is beneficial if they are defined as a table with "Key and Value" as a table header.

**Test Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

**Impacted MAS Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={b24201d1-2276-4471-8183-e1e5353031ad}](url)

**Attachment:**
[consolidated focus goes on Key and Value.zip](https://github.com/microsoft/BotFramework-Emulator/files/3690524/consolidated.focus.goes.on.Key.and.Value.zip)

ShikhaMishra11:
Fixed, Verified on MAC build 10.15 Beta(19A573a

PR #2002, opened 2019-11-21
[ISSUE#1908][MAS4.2.10][Focus Order - Add other service] With voiceover consolidated focus goes on "Key and Value".

denscollo:
Solves #1908

### Description
This pull request fixes the navigation with **Voiceover** through the key-value pair text fields in the _Add other service ..._ dialog.

Now voiceover goes from the label `key` to its text field, then to the `value` label and then to its text field.

### Changes made
We moved the labels used for the titles inside the first _TextField_ element of the pairs, this way the label is associated with the input field it corresponds.

Also, we changed how the key-value pairs are displayed in the dialog, replacing a list for a table using rows as the tester suggested.

Finally, we made some changes to the styling to have the same visual representation as before. For this, we moved some style properties inside a new class to be used for the rows.

### Testing
In the following recording, you can see **Voiceover** navigation as it should through the fields.
kvpair


denscollo:
Thanks @tonyanziano for the feedback! We'll be working on this and let you know when we have it ready.

PR #2005, opened 2019-11-22
[ISSUE#1911][MAS1.3.1][Screen Reader-Add other service] After error, the voiceover announced incorrect label of the edit fields.

### LINKED ISSUE #1911: [MAS1.3.1][Screen Reader-Add other service] After error, the voiceover announced incorrect label of the edit fields.
**Actual Result:**
The voiceover announces the correct label before the any of the field has error. After error it announces the same error message to every edit field.
Voiceover announces "Service Name*, Invalid required data unselected" before the error "The field cannot be empty", after the error it reads same error message at every edit field.

**Expected Result:**
The voiceover should reads the error only on the edit field where the error exist and on the other edit field the labels should be announced which are already defined for them.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened.
3. Navigate to the "Service" and click on "+".
4. Select "Add other service" from the "Menu".
5. Enter any data in "Service Name" and delete them.
6. An error "The field cannot be empty" appears, ignore the error and press "Tab".
7. Observe the issue.

**User Impact:**
Users who depends on the Screen reader will get confused if same error is on the every controls.

**Note:**
Same issue observed if user "edit and delete" any required field.
Same issue observed in edit fields present at "Add a LUIS app manually" pane.

**Test Environment:**
OS: MacOs Version 10.15
Application: Bot Framework V4 Emulator (Version 4.5.2-80829)

**Impacted MAS Link:**
[https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={54f28d1f-a2d1-4dcd-84e1-5c9b87e8aba4}](url)

**Attachment:**
[Incorrect label after error.zip](https://github.com/microsoft/BotFramework-Emulator/files/3691026/Incorrect.label.after.error.zip)

denscollo:
Fixes #1911

### Description
This pull request fixes how the Voiceover reads the error label in _TextField_ components,

### Changes Made
Updated the _TextField_ component to use an `aria-label` instead of an `aria-describedby` attribute. This way we can create a label with the combination of the component label and the error message.

### Testing
issue1911

### DIFF
--- a/packages/sdk/ui-react/src/widget/textField/textField.tsx
+++ b/packages/sdk/ui-react/src/widget/textField/textField.tsx
@@ -103,7 +103,7 @@ export class TextField extends Component<TextFieldProps, {}> {
   protected get errorNode(): React.ReactNode {
     const { errorMessage } = this.props;
     return errorMessage ? (
-      <sub id="errormessagesub" className={styles.sub}>
+      <sub id="errormessagesub" className={styles.sub} aria-live={'polite'}>
         {errorMessage}
       </sub>
     ) : null;
--- a/packages/sdk/ui-react/src/widget/textField/textField.tsx
+++ b/packages/sdk/ui-react/src/widget/textField/textField.tsx
@@ -71,7 +71,7 @@ export class TextField extends Component<TextFieldProps, {}> {
       <div className={`${styles.inputContainer} ${inputContainerClassName}`}>
         {this.labelNode}
         <input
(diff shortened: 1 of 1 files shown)


denscollo:
@corinagum thanks for the feedback! We'll run some tests to make sure about this and let you know.

Issue #1979, opened 2019-11-08
[MAS2.1.1] [Keyboard Navigation-All headers] Header menu is not accessible using keyboard.


ShikhaMishra11:
**Actual Result:**
Header menu is not accessible using keyboard i.e. file, debug, edit etc..

**Expected Result:**
Header menu(i.e. file, debug, edit etc.) should be accessible using keyboard as it is accessible by mouse. The application should have any shortcut key to access the header.

**Repro Steps:**
1. Open the Bot Framework V4 Emulator.
2. Welcome tab screen gets opened.
3. Try to navigate to the header menu using keyboard and observe the issue.

**Observation:**
We try to access the header's with the shortcut key "Alt+F" mentioned in windows bug: https://github.com/microsoft/BotFramework-Emulator/issues/1802. This bug is fixed for the windows.

**User Impact:**
Keyboard dependent users won't be able to use the feature's available on the header menu if they are not accessible using keyboard in any way.

**Test Environment:**
OS: Ubuntu 18.04.3 LTS
Application: BotFramework-Emulator-4.5.2-linux

**MAS Reference Link:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={8492c4eb-c179-40ae-8777-cd044ed725a2}

**Attachment:**
[Header menu is not accessible using keyboard.zip](https://github.com/microsoft/BotFramework-Emulator/files/3824256/Header.menu.is.not.accessible.using.keyboard.zip)

awalia13:
@tonyanziano we have verified the issue and observed still we are not able to access the header items through tab key and also through short cut key 'ALT +F'.

tonyanziano:
This has been fixed and is ready to be verified again @awalia13

ShikhaMishra11:
@tonyanziano As checked in latest nightly build "BotFramework-Emulator-4.6.0-95378-linux", the header's are accessible by tab keys.

So, we are closing the issue.

PR #2022, opened 2019-12-10
[#1979] Fixed app menu tab navigation issue on Linux.

tonyanziano:
#1979

===

- Actually fixing it this time 😅
- When the tab focus logic was running it was only looking inside the nav bar for the first focusable element, when it actually needs to look in the app menu on Linux

Video of it working:

[linux-tab-nav.zip](https://github.com/microsoft/BotFramework-Emulator/files/3947592/linux-tab-nav.zip)

### DIFF
--- a/packages/app/client/src/utils/eventHandlers.ts
+++ b/packages/app/client/src/utils/eventHandlers.ts
@@ -137,13 +137,20 @@ class EventHandlers {
 
     if (isMac() || isLinux()) {
       const tabPressed: boolean = key === 'tab';
-      const lastDecendants = EventHandlers.getLastDecendants(document.querySelector('main'));
-      const firstElement = document.querySelector('nav').firstElementChild as HTMLElement;
-      const lastElement = lastDecendants[lastDecendants.length - 1] as HTMLElement;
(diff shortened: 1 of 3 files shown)


packages/app/client/src/utils/eventHandlers.ts
```diff
@@ -137,13 +137,20 @@ class EventHandlers {
...
     if (isMac() || isLinux()) {
       const tabPressed: boolean = key === 'tab';
-      const lastDecendants = EventHandlers.getLastDecendants(document.querySelector('main'));
-      const firstElement = document.querySelector('nav').firstElementChild as HTMLElement;
-      const lastElement = lastDecendants[lastDecendants.length - 1] as HTMLElement;
-      const isFirstElement: boolean = document.activeElement === firstElement;
-      const isLastElement: boolean = document.activeElement === lastElement;
 
       if (tabPressed) {
+        const lastDecendants = EventHandlers.getLastDecendants(document.querySelector('main'));
```
cwhitten: the only other way I've done this is with `document.querySelectorAll` and slicing the resulting array up.

Issue #1405, opened 2019-04-02
Please add local directLine websocket handling into the Bot Emulator Framework


vseanclark:
We have implemented a custom channel that is dependent on the directLine messages via the websocket.  When running with the Bot Emulator Framework locally, we do not receive these websocket messages as documented.  Because of this, our local code behaves differently than the code when executed in the cloud.  

Our enterprise policy does not allow using NGROK as the suggested workaround for this issue.

We request the addition of the local directLine websocket handling into the Bot Emulator Framework.

justinwilaby:
DirectLine is undergoing a major overhaul which may mandate the use of websockets. Once this is fully baked, we will implement the new protocol in the Emulator. 

With that said, I will open a discussion with the team about whether or not we want to implement the existing protocol via websockets. In the meantime, keep an eye on #1394 since this feature will allow you to use any channel for locally running bots including direct line via web sockets and sill connect to the Emulator.

tonyanziano:
@vseanclark we have just merged in a PR (#2034) that switches communication between the client (Web Chat) and the Emulator to Web Sockets instead of REST polling via GET.

It may be worth checking out to see if it resolves your issues.

It will be available in the newest nightly at around 10:30 PM PST tonight.

cwhitten:
Closing as fixed!

PR #2034, opened 2020-01-08
[#2042] Changed convo infrastructure to use Web Sockets.

tonyanziano:
related to #1405 , #2042 

===

## Highlights

- Activities from the bot are now sent to Web Chat over a Web Socket connection (no more endless polling / Network tab spam!!! 🎉🎉🎉)
- Pulled heavy-duty conversation initialization logic out of `emulator.tsx` component
  - this logic has been moved to `botSagas.ts` and `chatSagas.ts` so now they can be called more declaratively instead of relying on changes to props satisfying the right criteria
- Consolidated conversation creation logic so that opening a bot via .bot file, URL, or opening a transcript all flow through (more-or-less) the same code path
- Consolidated code path for opening a transcript via a `.chat` file versus a `.transcript` file
  - they essentially call the same code, except when handling a `.chat` file, we first use `chatdown` to convert the file to a transcript before extracting the activities
- Removed global user state
  - users are now scoped to the conversation level
  - no more having to call the `SetCurrentUser` command and manually maintaining the state
  - removed the `users` property from persisted `server.json` file (no more endless list of randomly generated user IDs from previous sessions)

---

**Passing E2E tests:**

image

---

**Diagram of High-level Changes**

image


packages/app/client/src/state/sagas/chatSagas.ts
```diff
@@ -121,64 +143,267 @@ export class ChatSagas {
...
   }
 
-  public static *newChat(action: ChatAction<Partial<ChatDocument & ClearLogPayload>>): Iterable<any> {
-    const { documentId, resolver } = action.payload;
+  public static *newTranscript(action: ChatAction<OpenTranscriptPayload>): Iterable<any> {
+    const { filename } = action.payload;
+    // start a conversation
+    const serverUrl = yield select(getServerUrl);
+    const user = { id: yield select(getCustomUserGUID) || uniqueIdv4(), name: 'User', role: 'user' };
+    const payload2 = {
```
tonyanziano: change to better variable name

packages/app/client/src/state/sagas/chatSagas.ts
```diff
@@ -121,64 +143,267 @@ export class ChatSagas {
...
+      newChat(documentId, mode, {
+        conversationId,
+        directLine,
+        userId: user.id,
+      })
+    );
+
+    // initialize speech
+    if (msaAppId && !msaPassword) {
+      // TODO: TEST SPEECH
```
tonyanziano: make sure to test speech and remove this comment

packages/app/main/src/commands/clientInitCommands.ts
```diff
@@ -103,8 +104,8 @@ export class ClientInitCommands {
       ProtocolHandler.parseProtocolUrlAndDispatch(protocolArg);
     }
 
-    // Parse command line args to see if we are opening a .bot or .transcript file
-    const fileToBeOpened = args.find(arg => /(\.transcript)|(\.bot)$/.test(arg));
+    // Parse command line args to see if we are opening a .bot, .chat, or .transcript file
+    const fileToBeOpened = args.find(arg => /(\.transcript)|(\.chat)|(\.bot)$/.test(arg));
```
srinaath: const fileToBeOpened: boolean = !!args.find(arg => /(\.transcript)|(\.chat)|(\.bot)$/.test(arg)); maybe? just to be more explicit

packages/app/main/src/server/routes/channel/conversations/handlers/createConversation.ts
```diff
@@ -31,89 +31,52 @@
...
-import { validateCreateConversationRequest } from './errorCondition/createConversationValidator';
+function validateRequest(payload): any {
+  if (!payload.bot) {
+    return new Error('Missing bot object in request.');
+  } else if (!payload.botEndpoint) {
+    return new Error('Missing botEndpoint object in request.');
+  } else if (payload.members.length !== 1 || payload.members[0].role !== 'user') {
+    return new Error('Missing user inside of members array in request.');
+  }
+  return undefined;
```
srinaath: Do we need the return undefined here?

packages/app/main/src/server/routes/emulator/handlers/updateConversation.ts
```diff
@@ -0,0 +1,71 @@
...
+      res.send(NOT_FOUND);
+      return next();
+    }
+
+    // update the conversation object and reset as much as we can to resemble a new conversation
+    state.conversations.deleteConversation(currentConversationId);
+    currentConversation.conversationId = conversationId;
+    currentConversation.user.id = userId;
+    const user = currentConversation.members.find(member => member.name === 'User');
+    user.id = userId;
```
srinaath: Should check if user is defined before using it

packages/app/main/src/server/routes/emulator/handlers/updateConversation.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (!currentConversationId) {
+      res.send(NOT_FOUND);
+      return next();
+    }
+
+    // update the conversation object and reset as much as we can to resemble a new conversation
+    state.conversations.deleteConversation(currentConversationId);
+    currentConversation.conversationId = conversationId;
+    currentConversation.user.id = userId;
+    const user = currentConversation.members.find(member => member.name === 'User');
```
srinaath: Can we indicate the type here for more readability?

packages/app/main/src/server/webSocketServer.ts
```diff
@@ -0,0 +1,106 @@
...
+    });
+    // dynamically generate the web socket server port
+    const port = await new Promise<number>((resolve, reject) => {
+      this._restServer.once('error', err => reject(err));
+      this._restServer.listen(null, () => {
+        resolve(this._restServer.address().port);
+      });
+    });
+    this.port = port;
+    console.log(`Web Socket host server listening on ${port}...`);
```
srinaath: Eslint warning

packages/app/client/src/state/sagas/botSagas.ts
```diff
@@ -83,140 +86,99 @@ export class BotSagas {
...
       BotSagas.commandService.remoteCall(SharedConstants.Commands.Telemetry.TrackEvent, 'bot_open', {
-        method: null, // this code path can be hit by multiple methods
         numOfServices: 0,
         source: 'url',
       });
     }
+    if (!isLocalHostUrl(action.payload.endpoint)) {
+      BotSagas.commandService
+        .remoteCall(SharedConstants.Commands.Telemetry.TrackEvent, 'livechat_openRemote')
+        .catch(_e => void 0);
```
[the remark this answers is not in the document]
tonyanziano: Definitely don't need to. I was just using that as a convention to show it's unused.

packages/app/client/src/ui/editor/emulator/emulator.tsx
```diff
@@ -423,21 +274,23 @@ export class Emulator extends React.Component<EmulatorProps, {}> {
...
-    const key = event.key.toLowerCase();
-    if (ctrlOrCmdPressed && shiftPressed && key === 'r') {
-      await this.onStartOverClick();
+    if (this.props.activeDocumentId === this.props.documentId) {
+      // Meta corresponds to 'Command' on Mac
+      const ctrlOrCmdPressed = event.getModifierState('Control') || event.getModifierState('Meta');
+      const shiftPressed = ctrlOrCmdPressed && event.getModifierState('Shift');
+      const key = event.key.toLowerCase();
+      if (ctrlOrCmdPressed && shiftPressed && key === 'r') {
+        await this.onStartOverClick();
```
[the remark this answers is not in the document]
tonyanziano: good catch! leftover from old logic

packages/app/main/src/commands/clientInitCommands.ts
tonyanziano: fileToBeOpened is actually a string that is used in the `openFileFromCommandLine()` method call below.

I'll add the boolean coercion to the `if` statement

packages/app/client/src/commands/emulatorCommands.ts
```diff
@@ -57,85 +54,19 @@ export class EmulatorCommands {
...
-        contentType: Constants.CONTENT_TYPE_TRANSCRIPT,
-        documentId: filePath,
-        fileName,
-        filePath,
-        isGlobal: false,
+      openBotViaUrlAction({
+        appId: endpoint.appId,
+        appPassword: endpoint.appPassword,
+        channelService: endpoint.channelService as ChannelService,
+        endpoint: endpoint.endpoint,
```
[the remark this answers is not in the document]
tonyanziano: I don't disagree with your point, but it would involved changing the definition of the `IEndpointService` type which we pull from a library `botframework-config`. We would then have to change every instance of the code that consumes this property.

tonyanziano:
I think the coverage number has gone down as a result of removing a lot of tested code, but I will try to write some more to lessen the drop.

tonyanziano:
Opened another PR with a bunch of new tests to compensate for drop in code coverage:

https://github.com/microsoft/BotFramework-Emulator/pull/2048

packages/app/main/src/utils/loadSettings.spec.ts
```diff
@@ -88,25 +95,29 @@ jest.mock('./ensureStoragePath', () => ({
...
-    expect(settings.users.currentUserId).not.toBe('default-user');
-    expect(Object.keys(settings.users.usersById)[0]).toBe(settings.users.currentUserId);
+  it('should return the default settings when the requested settings file does not exist', () => {
+    mockIsFile.mockReturnValueOnce(false);
+    const settings = loadSettings('appData/server.json', mockSettings);
+
+    expect(settings).toEqual(mockSettings);
+  });
+
+  it('should return the default settings whenthere is an error trying to load the settings', () => {
```
srinaath: space between when and there

packages/app/main/src/server/webSocketServer.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+jest.mock('restify', () => ({
+  createServer: (...args) => mockCreateServer(...args),
+}));
+
+describe('WebSocketServer', () => {
+  beforeEach(() => {
+    mockCreateServer.mockClear();
+  });
+
+  it('should return the corresponding socket for a conversation id', () => {
```
srinaath: Is there a test to make sure that we are generating only one socket connection per conversation. I see your code handling it correctly but just want to make sure one of these tests also handles that case

packages/app/client/src/state/sagas/chatSagas.spec.ts
```diff
@@ -30,34 +30,43 @@
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
```
srinaath: Awesome test coverage on this core functionality! Just for more readability, do you think grouping the tests in different describe functions where each one is dedicated for a saga function would be great.

srinaath: In addition tests where you can dispatch the action that triggers the saga function would also be good even it its as simple as dispatching the action and making sure a spy on the saga function is hit

tonyanziano: Good suggestions! I think that's a good idea, these tests are pretty long

tonyanziano: I can add tests that test iterating through the root saga function and making sure the correct `takeEvery()` calls are yielded, which sets up the integration with redux.

Spying on the sagas themselves is not very easy because you have to mock everything that would allow a successful run through the happy path of the saga. Unless I'm missing some simple pattern that would allow us to do that.

Adding all that code to test the integration between two well-known libraries (`redux` & `redux-saga` -- which should already be tested by `redux-saga`) doesn't really seem worth it to me.

packages/app/client/src/state/sagas/resourcesSagas.spec.ts
```diff
@@ -274,37 +276,31 @@ describe('The ResourceSagas', () => {
...
+          path,
         },
-      ]);
+      };
+      const gen = ResourcesSagas.doOpenResource(mockAction);
+      const putValue = gen.next().value;
+
+      expect(putValue).toEqual(put(openTranscript(path)));
+      expect(gen.next().done).toBe(true);
     });
```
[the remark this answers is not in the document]
tonyanziano: same as my comment for `chatSagas.spec.ts`

PR #2060, opened 2020-01-24
Moved shared store code to app/shared.

### LINKED ISSUE #2038: [Enhancement] Move state store code to /app/shared
Currently, the action and reducer code, is duplicated in both `/packages/app/client` and `/packages/app/main`.

The number of files here could be halved by moving all this shared code into `/packages/app/shared/state` where both the main and client code bases could import it from.

The only thing left on either side would be the store initialization code and that side's sagas.

tonyanziano:
Fixes #2038 

===

Basically moved all reducer and action code to `packages/app/shared/state`.

So 99% of this PR is changing imports from:

```
import { closeBot } from '../../state/actions/botActions'
import { bot } from '../../state/reducers/bot';
```

to

`import { bot, closeBot } from '@bfemulator/app-shared';`

350 files... whoops 😅


tonyanziano:
I didn't really add any new code, I just deleted a bunch of duplicate tested code and I believe that's why the coverage dropped:

**Before:**

2633 of 4488 branches covered (58.67%)

**After:**

2428 of 4267 branches covered (56.9%)

srinaath:
This PR is awesome @tonyanziano . So much redundancy removed

PR #2066, opened 2020-01-28
Fixed an issue with adding a QnA KB.

### LINKED ISSUE #1875: Connecting to a QnA maker knowledge base button throws an error
## Version
Tested on both 4.5.2-79456 and today's non-published bits

## Describe the bug
Opening the dialog to connect a QnAM kb manually throws an error

## To Reproduce
1. Open a bot file
2. Open context menu to add a service
3. Select "Add a QnaMaker"
4. Select "Add a QnA Maker knowledge base manually" button
5. See error in console

## Expected behavior
Should open the Connect a QnA dialog to show

## Screenshots
qnaerror

## Additional context
Please don't fix until I get my `<LinkButton>` PR in... xD

[bug]

tonyanziano:
Fixes #1875

===

- After signing into Azure, trying to add a QnA maker KB would cause the in-memory QnA maker service to be initialized with a blank hostname, which was causing `msbot` to throw and crash the app

### DIFF
--- a/packages/app/client/src/state/sagas/servicesExplorerSagas.ts
+++ b/packages/app/client/src/state/sagas/servicesExplorerSagas.ts
@@ -64,6 +64,7 @@ import {
 import { sortExplorerContents } from '../actions/explorerActions';
 import { SortCriteria } from '../reducers/explorer';
 import { RootState } from '../store';
+import { QnAMakerSampleHostname } from '../../constants';
 
 import { AzureAuthSaga } from './azureAuthSaga';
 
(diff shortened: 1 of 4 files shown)


packages/app/client/src/state/sagas/servicesExplorerSagas.ts
```diff
@@ -166,7 +166,8 @@ export class ServicesExplorerSagas {
     if (result === 1) {
       action.payload.connectedService = BotConfigurationBase.serviceFromJSON({
         type,
-        hostname: '' /* defect workaround */,
+        hostname:
+          'https://myqna.azurewebsites.net' /* a QnA maker service needs to be initialized with a valid hostname or msbot will throw */,
```
srinaath: Are we fine hardcoding this instead of pulling it from a constants?

tonyanziano: Good call, we actually use this in one more place in the code, so I'll turn it into a constant. 👍

PR #2067, opened 2020-01-29
Removed bubble user background color. Leave it to default option

srinaath:
Revert the bubble background to use default color from webchat


srinaath:
BEFORE:
before-filename-hidden


AFTER:
after-background-remove

PR #2079, opened 2020-02-13
Enabled communication with remote Direct Line Speech bots

### LINKED ISSUE #2040: Modify WebChat implementation to support connecting to a Direct Line Speech bot remotely
WebChat supporting DirectLine Speech requires an update to the way we configure the control in the Emulator. When the user connects to a bot we need to add a field to denote if it is a DL-Speech bot and if so, the WebChat needs to be instantiated with some additional user-driven parameters. See https://github.com/microsoft/BotFramework-WebChat/blob/master/DIRECT_LINE_SPEECH.md for more detail, and reach out to @compulim with questions.

[enhancement]

tonyanziano:
Fixes #2040

===

This PR adds the ability to connect to a remote Direct Line Speech bot by entering a speech key and speech region when connecting to a bot.

Example:

region: `westus`
key: `some-speech-key-guid`

image

### DIFF
--- a/packages/app/client/src/state/sagas/chatSagas.ts
+++ b/packages/app/client/src/state/sagas/chatSagas.ts
@@ -535,7 +535,6 @@ function createWebChatActivitySniffer(conversationId: string, serverUrl: string)
         console.error(`Failed to log DL Speech activity: ${errText}`); // eslint-disable-line no-console
       }
     }
-
     return next(action);
   };
 }
--- a/packages/app/client/src/state/sagas/chatSagas.ts
+++ b/packages/app/client/src/state/sagas/chatSagas.ts
@@ -41,6 +41,7 @@ import {
   open as openDocument,
   setInspectorObjects,
   updatePendingSpeechTokenRetrieval,
+  updateSpeechAdapters,
   webChatStoreUpdated,
   webSpeechFactoryUpdated,
   ChatAction,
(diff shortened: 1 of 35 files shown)


tonyanziano:
Linting is very unhappy due to a patch bump from `eslint-plugin-import@2.20.0` to `2.20.1` as a side effect of regenerating the `package-lock`. This causes all of our `import` statements to fail linting resulting in a 400+ file change to satisfy the updated linter.

I'm going to make a separate PR that locks the `eslint-plugin-import` package to `2.20.0` so that we do not have to deal with this massive linting delta.

packages/app/client/src/state/utils/throwErrorFromResponse.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (response.text) {
+    errText = yield response.text();
+  }
+  let responseStatus = '';
+  if (response.statusText) {
+    responseStatus = `${response.status} ${response.statusText}`;
+  } else {
+    responseStatus = response.status + '';
+  }
+  throw new Error(`${error || 'Error'} ${responseStatus}: ${errText}`);
```
srinaath: Is this the general format we are following. I feel like we can have error.desc: StatusCode + errText, error.message: errText, and error.number: statusCode. In that way consuming it might be easier in cases where we are just looking for status codes

packages/app/client/src/state/sagas/chatSagas.ts
```diff
@@ -451,6 +519,27 @@ export class ChatSagas {
...
+      );
+      if (!res.ok) {
+        let errText = '';
+        if (res.text) {
+          errText = await res.text();
+        }
+        console.error(`Failed to log DL Speech activity: ${errText}`); // eslint-disable-line no-console
+      }
+    }
+
```
srinaath: can remove extra space

packages/app/client/src/state/utils/throwErrorFromResponse.ts
tonyanziano: Sure, I like the idea of breaking it up into different properties. 👍

packages/app/shared/tsconfig.json
```diff
@@ -1,8 +1,10 @@
 {
```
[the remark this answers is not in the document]
tonyanziano: Like you pointed out, I can only move some of the fields to the root config due to the discrepancy between some of the packages' build folders.

It looks like the TypeScript compiler `tsc` is only called from 5 different package directories, so ideally we could broaden the scope of the GitHub ticket you mentioned to clean up some of the overall unnecessary TS build config as well as align the build folders.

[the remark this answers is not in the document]
tonyanziano: https://github.com/microsoft/BotFramework-Emulator/issues/2083

Issue #2086, opened 2020-02-26
Emulator update does not restart conversation which requires a close and re-open of the app.


xtianus79:
## Version
Version 4.7.0-107479 (4.7.0-107479.107479)

## Describe the bug
restart conversation doesn't seem to have any affect on the emulator.  Thus needing to close and reopen after every change. 

## To Reproduce
setup a bot and click restart conversation. 

to restart

[bug]

tonyanziano:
@xtianus79 could you please press (`Ctrl + Shift + I` for Windows, or `Cmd + Shift + I` for Mac) in order to open the developer tools window, and try clicking the Restart Conversation button?

Once you do, please send a screenshot of the console tab with any error messages.

I have tried reproducing this on version 4.7.0-107479 and the button works as expected for me.

xtianus79:
[here is image](https://imgur.com/a5fvtmP) @tonyanziano this is the screenshot.  It works sometimes but after it breaks it won't refresh anymore.

tonyanziano:
@xtianus79 thanks for the screenshot, that's very helpful.

I think I may know where this error is originating from. I'll take a look.

tonyanziano:
I have found the issue that is causing the error you are seeing. [The error is resulting from this line here](https://github.com/microsoft/BotFramework-Emulator/blob/master/packages/app/client/src/state/utils/throwErrorFromResponse.ts#L53), because we are calling `text()` outside of the context of `response` which is wrong.

However, once I fixed that, I noticed that the code is still not surfacing the correct error that you should be seeing which is the result of not being able to post an activity to your bot. I'm working on fixing that now.

tonyanziano:
@xtianus79 There is a PR out #2087, that will show a more detailed error in the console so we can debug why restarting a conversation is sometimes failing for you.

tonyanziano:
Hi @xtianus79 ,

Please update to [the newest nightly version, 4.7.0-109326](https://github.com/microsoft/botframework-emulator-nightlies/releases/tag/v4.7.0-109326), and try to reproduce this issue again.

When you run into the issue, open the dev tools console window, and you should now see a detailed error message about what went wrong. Feel free to post that error here and we can help you debug.

xtianus79:
@tonyanziano yes that is working!!! thank you sir

xtianus79:
[...]

PR #2087, opened 2020-02-27
Thrown errors in client sagas are now logged to console.

tonyanziano:
This will help debug issues like #2086 

===

Due to the way sagas work, the errors thrown in our nested sagas are seen as uncaught and redux saga hides most of the internal properties of the thrown error and just shows the `message`, which isn't very helpful.

`redux-saga` provides an `onError` hook that you can use for catching these errors at the root level, but the `onError` will only get called if the caught error is of type `Error`. Since we defined our own error type `ResponseError` the `onError` hook never gets called. We need to update to `redux-saga@1.0.0` for that check to be fixed, but this might incur some other breaking changes that I don't think this PR should deal with. [See this issue for reference.](https://github.com/redux-saga/redux-saga/issues/1308)

For now, we will surface the error in its entirety through `console.error()`.

**Before the change:**

saga-err-before

**After the change:**

saga-err-after

### DIFF
--- a/packages/app/client/src/state/utils/throwErrorFromResponse.ts
+++ b/packages/app/client/src/state/utils/throwErrorFromResponse.ts
@@ -50,7 +50,7 @@ export function* throwErrorFromResponse(errorMessage: string, response: Response
     status,
   };
   if (text) {
-    const errText = yield text();
+    const errText = yield text.call(response);
     error.innerMessage = errText;
   }
   if (statusText) {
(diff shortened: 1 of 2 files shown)


packages/app/client/src/state/utils/throwErrorFromResponse.ts
```diff
@@ -50,13 +50,18 @@ export function* throwErrorFromResponse(errorMessage: string, response: Response
     status,
   };
   if (text) {
-    const errText = yield text();
+    const errText = yield text.call(response);
```
tonyanziano: This was also throwing the following error:

```
TypeError: Failed to execute 'text' on 'Response': Illegal invocation
```

because `text()` was being called outside of the `response` context.

tonyanziano:
@srinaath once we bump `react-saga` to 1.0.0+ we can use the top-level `onError` hook to accomplish this

Issue #1683, opened 2019-07-14
Drop down list is collapse when I'm trying to select a value 


Shon-Elias:
## Version
What version of the Emulator are you using?
Version 4.5.1

## Describe the bug
I'm trying to select a value from the list but every time I open the list and hover the list it collapsed (**Input.ChoiceSet**)

## To Reproduce
1. using the emulator Version 4.5.1
2. try to select a value from Input.ChoiceSet
[bug]

tonyanziano:
This is probably due to the same logic that caused #1680 .

We should be able to add a check for `<select>` tags like we did for `<input>`'s in #1681

tonyanziano:
This issue has been fixed in the newest release, [4.5.2](https://github.com/microsoft/BotFramework-Emulator/releases/tag/v4.5.2) available now.

PR #2090, opened 2020-03-05
Fixed an issue that was causing adaptive card focus to be blurred whe…

tonyanziano:
…n clicking an activity.

===

Fixes a regression of #1683 

===

**The problem:**

When react-redux's `mapStateToProps` returns an object with the newest version of a component's props, it does a shallow compare (`===`) of each of the new props' values with the previous props' values and if any of them return false, then it re-renders the component. [Read more here.](https://react-redux.js.org/using-react-redux/connect-mapstate#return-values-determine-if-your-component-re-renders)

Since `currentUser` was being assigned to a new object ever time `mapStateToProps` was run, even though the previous `currentUser` and the new one had the same key/value pairs.


srinaath:
Awesome fix @tonyanziano

PR #2098, opened 2020-03-11
4.8 blockers

srinaath:
Copy to clipboard fixed
Making sure restart conversation does not show up in Transcript or Debug mode

Signed-off-by: Srinaath Ravichandran <srravich@microsoft.com>


cwhitten:
thanks!

Issue #2106, opened 2020-03-13
Version 4.8 issues error "[onTurnError]: Error: missing activity.from.id"


hinojosachapel:
## Version
4.8.

## Describe the bug
I was happily using version 4.7. Everything was fine. Then the "Update available" notification appeared and I proceeded with it. After that, I was unable to use the emulator. A consistent error was thrown from my bot: "[onTurnError]: Error: missing activity.from.id". After some time wasted figuring out what was happening, I uninstalled version 4.8 and reinstalled 4.7. And everything was right again. Now I continue to use 4.7 till this issue get fixed.

## To Reproduce
1. Just open a bot in the emulator version 4.8
2. See error "[onTurnError]: Error: missing activity.from.id".

## Expected behavior
To work as expected, just like in version 4.7.
[bug]

tonyanziano:
Hi @hinojosachapel ,

Thanks for reporting this issue. I'm sorry to hear that you're having trouble with the latest release.

Is there a certain scenario with your bot that causes this error to appear? Or are you just trying to send a message to your bot when you open it?

Is this happening at the stage of the conversation when it is first opened and the conversation update activity is sent, or is this later in the conversation when you try to send a message to the bot?

tonyanziano:
I have found the bug and am working on a fix.

tonyanziano:
@hinojosachapel I have made a PR that will fix the bug, and we aim to release a new patch (v4.8.1) of the Emulator later today or tomorrow that will include this fix.

tonyanziano:
@hinojosachapel please update to the latest nightly version -- **4.8.0-113681** -- and try talking to a bot to verify that the issue has been fixed.

Please re-open the issue if not.

hinojosachapel:
@tonyanziano Thanks a lot. I can confirm that the issue has been fixed. Good work! Regards.

PR #2108, opened 2020-03-16
Fixed a bug that was causing a blank User 'id' field.

tonyanziano:
#2106

===

Fixed some faulty logic that was not behaving as expected.

The `select()` effect from `redux-saga` returns an object, so it is always truthy no matter what and will resolve to the yielded value.

Ex:

```
const val = yield select(() => false) || 'some other value';
console.log(val);  // false
```

So even when the custom user ID was set to the empty string, the start conversation logic was always initializing the conversation with the blank ID rather than defaulting to a freshly generated GUID.

### DIFF
--- a/packages/app/client/src/state/sagas/chatSagas.spec.ts
+++ b/packages/app/client/src/state/sagas/chatSagas.spec.ts
@@ -824,10 +824,13 @@ describe('The ChatSagas,', () => {
       // put webSpeechFactoryUpdated
       expect(gen.next().value).toEqual(put(webSpeechFactoryUpdated(payload.documentId, undefined)));
 
+      // select custom user GUID
+      expect(gen.next().value).toEqual(select(getCustomUserGUID));
+
       // call updateConversation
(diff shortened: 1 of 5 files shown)


packages/app/client/src/state/sagas/chatSagas.spec.ts
```diff
@@ -824,10 +824,13 @@ describe('The ChatSagas,', () => {
...
       expect(gen.next().value).toEqual(put(webSpeechFactoryUpdated(payload.documentId, undefined)));
 
+      // select custom user GUID
+      expect(gen.next().value).toEqual(select(getCustomUserGUID));
+
       // call updateConversation
       const conversationId = chat.conversationId;
       const userId = chat.userId;
-      expect(gen.next().value).toEqual(
+      expect(gen.next('').value).toEqual(
```
srinaath: Also can we assert that userID is not blank in the case of yield select() returning an empty string? Basically, making sure the unique ID is being generated for userID?

PR #2118, opened 2020-03-20
Fix #2109

### LINKED ISSUE #2109: req.conversation.postActivityToUser is not a function
## Version
4.8.0

## Describe the bug
I want to save then resume a conversation by applying **SendConversationHistoryAsync** method with a new **Transcript** object (not saved by the middleware) filled with a custom list of activities and the right conversation ID. When the emulator receives that request, i have an error which was not here in 4.7.
"Processed 0 of 4 activities successfully.TypeError: req.conversation.postActivityToUser is not a function"

## To Reproduce
1. Save conversation ID with the list of activities in storage
2. Retrieve it from the storage
3. Post it back to the user/emulator by using SendConversationHistoryAsync
4. "Processed 0 of 4 activities successfully.TypeError: req.conversation.postActivityToUser is not a function" error on the emulator

## Expected behavior
I want the emulator to replay the exact list of activities i post to the emulator using the SendConversationHistoryAsync method. **I know emulator is for testing purpose**. If i am wrong in the implementation of that behavior, please explain me why !

## Screenshots
Bug emulator

## Additional context
1. It is sent using the **Conversations** property of a newly **ConnectorClient** class instantiated with the right values.
2. I tried to change the information inside activities like updating activities id, timestamps. But i did try other properties without success (properties which i thought could be responsible).
3. In the 4.7, i had the activities showing but not user sent activities, only bot activities were displayed.
4. I did not tried to debug emulator code on that scenario.
5. I am kind of new using all of these tools, excuse me if i went wrong !

[bug]

TemoeBOUISSOU:
Fixes #2109 

===

I did my best, i am not sure of the impacts on your side, but this implementation was the simpliest one i found !

### DIFF
--- a/packages/app/main/src/server/routes/channel/conversations/handlers/sendHistoryToConversation.ts
+++ b/packages/app/main/src/server/routes/channel/conversations/handlers/sendHistoryToConversation.ts
@@ -36,15 +36,20 @@ import * as HttpStatus from 'http-status-codes';
 import { Next, Request, Response } from 'restify';
 
 import { createResourceResponse } from '../../../../utils/createResponse/createResourceResponse';
+import { WebSocketServer } from '../../../../webSocketServer';
+import { Conversation } from '../../../../state/conversation';
 
 export function sendHistoryToConversation(req: Request, res: Response, next: Next): any {
   const { activities }: { activities: Activity[] } = req.body;
   let successCount = 0;
   let firstErrorMessage = '';
-
-  for (const activity of activities) {
+  const { conversation }: { conversation: Conversation } = req as any;
+  for (var activity of activities) {
     try {
-      (req as any).conversation.postActivityToUser(activity, true);
+      activity = conversation.prepActivityToBeSentToUser(conversation.user.id, activity);
+      const payload = { activities: [activity] };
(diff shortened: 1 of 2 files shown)


packages/app/main/src/server/routes/channel/conversations/handlers/sendHistoryToConversation.ts
```diff
@@ -36,15 +36,20 @@ import * as HttpStatus from 'http-status-codes';
...
   const { activities }: { activities: Activity[] } = req.body;
   let successCount = 0;
   let firstErrorMessage = '';
-
-  for (const activity of activities) {
+  const { conversation }: { conversation: Conversation } = req as any;
+  for (var activity of activities) {
     try {
-      (req as any).conversation.postActivityToUser(activity, true);
+      activity = conversation.prepActivityToBeSentToUser(conversation.user.id, activity);
```
srinaath: We can create a mutated object here.
const updatedActivity = conversation.prepActivityToBeSentToUser(conversation.user.id, activity);
instead of updating the activity which we are iterating over.

srinaath:
@TemoeBOUISSOU  LGTM barring a couple of minor changes.

PR #2155, opened 2020-06-05
Fixed a bug where OAuth cards were opening the file explorer if ngrok…

### LINKED ISSUE #2130: Clicking "Sign In" on an OAuth Card opens the file explorer
In the recent version 4.8.1 of Emulator, 
I'm running the Authentication sample Bot, in that while Signing in , the emulator is not redirecting to a browser window to generate the token, instead it is opening the explorer window. 

Is there any fix or workaround for this issue. 

because of this issue I'm unable to debug my project, also this issue is not there in the previous version,

tonyanziano:
… was not configured.

Fixes #2130 

===

**The bug:**

- Clicking the sign-in button on an OAuth card would open the OS file explorer if ngrok was not configured in the app settings at the time the card was generated (see issue for reference video)

**The cause:**

- Whenever the bot sends an Adaptive Card to the user, the Emulator first scans it to see if it is an OAuth card, and if it is, we go ahead and use the current ngrok URL to construct a sign-in link to append to the card before forwarding it to Web Chat and the user
- However, if ngrok is not configured and the Emulator cannot start an instance of ngrok, then receiving a token from the token service will be impossible, so we should instead generate a fake sign-in link that the Emulator will interpret and know to send a fake, emulated token
- The Emulator was detecting that there was no ngrok, but it was not generating the fake sign-in link to complete the delivery of the emulated token to the bot. It was just propagating the error the log panel and stating that it was falling back to the emulated token:

ngrok-err

**The fix:**

- Upon detecting that ngrok is not configured, the Emulator will now properly construct the fake sign-in link before propagating the error up to the log panel

**Extra:**

- Got rid of some related, unused code

===

[Video of fix in action](https://recordit.co/0m7wEshTem)


tonyanziano:
Green CI build for reference:

https://fuselabs.visualstudio.com/BotFramework-Emulator/_build/results?buildId=136683&view=results

PR #2157, opened 2020-06-08
fix: Update npm outdated packages and audit fixes

srinaath:
Description:
This PR updates all npm oudated packages and audit fixes. Majority of the changes are due to the default prettier rule for brackets around single arguments passed to function. This PR also fixed a few Component governance libs marked as vulnerable

Some Component governance High vulnerable libs could not be updated because of no support for selective resolutions without yarn.

Current High vulnerable Dependancies that could not be upgraded
`dot-prop 4.2.0, dot-prop 3.0.0` - Transitive dependency of latest lerna
`lodash` - This can be ignored as lodash creator does not intend on fixing this for now
`serialize-javascript` - Cannot be updated as latest webpack references a vulnerable version of it as a transitive dependency


package.json
```diff
@@ -1,35 +1,4 @@
...
-    "rebuild:keytar:electron": "rimraf node_modules/keytar/build && electron-rebuild",
-    "rebuild:keytar:node": "npm rebuild keytar",
-    "start": "cd packages\\app\\client && npm run start",
-    "start:test-bot": "lerna run --scope @bfemulator/tools start:test-bot --stream",
-    "test": "npm run rebuild:keytar:node && jest --no-cache --runInBand --testPathIgnorePatterns=/packages/app/main/e2e/",
-    "test:coveralls": "jest --runInBand --coverage --coverageReporters=text-lcov | coveralls",
-    "test:e2e": "npm run rebuild:keytar:electron && jest --no-cache --runInBand --forceExit --testPathPattern=/e2e/.*\\.spec\\.ts$",
-    "test:gen-lcov": "npm run test -- --coverage --coverageReporters=lcov --coverageReporters=text",
-    "uploadcoverage": "cat ./coverage/lcov.info | coveralls"
-  },
```
tonyanziano, line 17: Why were all these removed? This will break our build system

packages/app/main/package.json
```diff
@@ -94,16 +94,15 @@
...
     "botbuilder": "^4.5.1",
     "chalk": "^2.4.1",
     "concurrently": "^3.5.1",
     "cross-env": "^5.1.3",
     "del": "^3.0.0",
-    "electron": "4.1.1",
-    "electron-builder": "20.29.0",
-    "electron-builder-http": "19.27.5",
-    "electron-builder-lib": "20.23.1",
+    "electron": "8.0.2",
```
tonyanziano, line 104: Is bumping Electron a necessity to fix any of the security vulnerabilities? I'm apprehensive about this due to the following:

- The last time we bumped electron up multiple major versions (2.x -> 4.1.1), there were some breaking changes that took us forever to find and fix and some that were only active in production. Have you gone through the list of breaking changes in Electron 5, 6, 7, and 8 and verified that we do not have any affected coder in our code base?
- If we change this version, we will have to modify our build system to pull the appropriate internal Microsoft-approved build of Electron
- We have to make sure this version has a Microsoft-approved build or is whitelisted by Microsoft

packages/extensions/qnamaker/package.json
```diff
@@ -10,12 +10,26 @@
...
+    "@babel/plugin-transform-react-jsx": "^7.10.1",
+    "@babel/preset-env": "^7.10.2",
+    "@babel/preset-typescript": "^7.10.1",
+    "@types/jest": "^25.2.3",
+    "babel-core": "^6.26.3",
     "babel-jest": "24.8.0",
-    "jest": "24.8.0",
+    "husky": "^4.2.5",
+    "jest": "^26.0.1",
+    "keytar": "^6.0.1",
```
tonyanziano, line 25: why is `keytar` being added to all the packages as a dev dependency? it is only required in the `packages/app/main` package and the root `package.json`

package.json
srinaath: Oh man. Definitely a bad cuty copy operation. Nice catch. I wanted to clear the dev dependancies alone

PR #2177, opened 2020-08-18
Handle change to middleware

srinaath:
The `webchat` middleware 4.10.0 returns false for `conversationUpdate, invoke` events. This PR makes sure it returns true before we render the results


packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
```diff
@@ -203,6 +207,7 @@ export class Chat extends PureComponent<ChatProps, ChatState> {
         return null;
 
       default:
+        debugger;
```
tonyanziano, line 210: might want to remove this 😅

tonyanziano (dismissed):

PR #2212, opened 2020-12-17
Increased wait time to 40seconds to spin ngrok process

srinaath:
This PR

1.  Increases the time to wait for the ngrok process before reporting an unsuccesful ngrok spawn operation
2.  Binding the `this` context on process exit. Currently. on process exit emulator does not cleanup ngrok process that was spawned  
    @garypretty

### DIFF
--- a/packages/app/main/src/ngrok.ts
+++ b/packages/app/main/src/ngrok.ts
@@ -200,7 +200,9 @@ export class NgrokInstance {
       };
 
       this.ngrokProcess.stdout.on('data', onNgrokData);
-      process.on('exit', this.kill);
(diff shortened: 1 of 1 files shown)


packages/app/main/src/ngrok.ts
```diff
@@ -200,7 +200,9 @@ export class NgrokInstance {
       };
 
       this.ngrokProcess.stdout.on('data', onNgrokData);
-      process.on('exit', this.kill);
+      process.on('exit', () => {
+        this.kill();
```
tonyanziano, line 204: nice catch

tonyanziano (approved):

PR #2226, opened 2021-01-29
Bumped Electron from 4.1.1 to 11.0.1

tonyanziano:
- Bumped `electron` to 11.0.1 🚀🚀🚀
- Bumped `typescript` to 4.1.3
- Bumped `keytar` to 7.3.0
- Bumped `electron-rebuild` to 2.3.4
- Bumped `eslint` to 7.19.0

Fixed some breaking API changes as a result of bumping Electron

Here's a working Windows CI build: https://fuselabs.visualstudio.com/BotFramework-Emulator/_build/results?buildId=207729&view=artifacts&pathAsName=false&type=publishedArtifacts

The other pipelines will be fixed to accommodate these changes


corinagum (approved):
LGTM

tonyanziano:
Thanks for the review @corinagum . I'm going to fix the linting and the tests as well and re-push so I might ping you again for a second review :)

packages/app/client/src/commands/uiCommands.ts
```diff
@@ -121,7 +121,7 @@ export class UiCommands {
   // ---------------------------------------------------------------------------
   // Shows a bot creation dialog
   @Command(UI.ShowOpenBotDialog)
-  protected async showOpenBotDialog(isDebug: boolean = false): Promise<void> {
+  protected async showOpenBotDialog(isDebug = false): Promise<void> {
```
corinagum, line 124: `: boolean` is not necessary? (for learning)

tonyanziano: I believe eslint said something along the lines of: "type `boolean` is trivially inferred, no need to provide a type"

So basically since we assigned it a default value of `boolean`, it's able to infer that the argument should be a `boolean`. I think if we left the default assignment out, then we would need to type it

corinagum (approved):
LGTM - one non-blocking question

packages/app/client/src/commands/uiCommands.ts
corinagum: oh that's weird. *shrug*. thanks for info

PR #2237, opened 2021-03-03
Fixed opening files via electron's dialog API

tonyanziano:
After bumping Electron from 4 to 11 we forgot to update our utility function that wraps the `dialog.showOpenDialog` API to return the correct value. This was causing any type of action that was using a native "Open file" dialog to silently fail -- opening a bot, transcript, etc.

[Here is the old documentation that shows the return value of v4.x's `showOpenDialog`](https://github.com/electron/electron/blob/v4.1.1/docs/api/dialog.md#dialogshowopendialogbrowserwindow-options-callback) -- it was an array of strings.

[The updated API from v11.x](https://www.electronjs.org/docs/api/dialog#dialogshowopendialogbrowserwindow-options) returns an object with a `filePaths` array containing the information that the 4.x version returned.

[CI build running here](https://fuselabs.visualstudio.com/BotFramework-Emulator/_build/results?buildId=219624&view=results)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-## v4.11.0 - 2021 - 3 - 03
+## v4.12.0 - 2021 - 3 - 03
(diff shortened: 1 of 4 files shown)


corinagum (approved):
LGTM

PR #2238, opened 2021-03-03
Fixed a bug that was causing Electron's context menu to fail

tonyanziano:
The `showMenuAndWaitForInput()` function was returning the entire `Electron.MenuItem` which contains functions as part of the object (click handler, etc.).

With Electron 11, these cannot be sent over IPC back to the client side because functions cannot be serialized. This was causing the result to never make it back to the client side and for the context menu action to silently fail.

I changed the API to return an object containing solely the `id` property of the selected menu item because all of the consumers of this API only care about the `id` property anyways.

[CI running here](https://fuselabs.visualstudio.com/BotFramework-Emulator/_build/results?buildId=219628&view=results)


srinaath (dismissed):

corinagum (approved):
LGTM

PR #2242, opened 2021-03-08
Added circular JSON handling to forwardToMain

tonyanziano:
When connecting to a Direct Line Speech enabled bot, the Web Chat speech adapter object embedded in one of the redux payloads contained a circular JSON structure -- `privReco` property -- that was causing the `JSON.stringify()` call in `forwardToMain.ts` to fail.

This was ultimately causing the Redux action to fail, which resulted in an `undefined` speech adapter object being sent to Web Chat.  Web Chat could not complete the connection to the DL Speech channel.

---

This PR adds error handling for this specific case, and will bypass forwarding the speech adapters update action over to the main process of the app. It is not needed in the main process's state store anyways.

All other errors will be thrown like normal.

As an example, here's what the error message looked like:

```
Converting circular structure to JSON
    --> starting at object with constructor 'p'
    |     property 'privReco' -> object with constructor 'D'
    --- property 'privRecognizer' closes the circle"
```

### DIFF
--- a/packages/app/client/src/state/middleware/forwardToMain.spec.ts
+++ b/packages/app/client/src/state/middleware/forwardToMain.spec.ts
@@ -79,4 +79,33 @@ describe('forwardToMain middleware', () => {
     expect(mockNext).toHaveBeenCalledWith(mockAction);
     expect(mockIpcRenderer.sendSync).toHaveBeenCalledWith('sync-store', mockAction);
   });
+
+  it('should skip forwarding the action to the main process if the payload has a circular structure', () => {
+    const innerObject = { outerObject: undefined };
+    const outerObject = { innerObject: undefined };
+    innerObject.outerObject = outerObject;
+    outerObject.innerObject = innerObject;
+    const mockNext = jest.fn();
+    const mockAction: any = {
+      type: 'actionType',
+      payload: { outerObject },
+    };
+    forwardToMain(null)(mockNext)(mockAction);
+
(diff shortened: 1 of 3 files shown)


packages/app/client/src/state/middleware/forwardToMain.spec.ts
```diff
@@ -79,4 +79,33 @@ describe('forwardToMain middleware', () => {
...
+
+    expect(mockNext).toHaveBeenCalledWith(mockAction);
+    expect(mockIpcRenderer.sendSync).not.toHaveBeenCalled();
+  });
+
+  it('should surface errors that are not circular JSON errors', () => {
+    const mockNext = jest.fn();
+    const mockAction: any = {
+      type: 'actionType',
+      payload: { someProp: 123 },
```
corinagum, line 103: Should this be testing with the circular object, with the expectation that no circular JSON error should be thrown? Then, it will throw `some error >:(` on line 106.

corinagum (approved):
LGTM - just one question

packages/app/client/src/state/middleware/forwardToMain.spec.ts
tonyanziano: If I pass it the circular object it will hit the `return next(action);` statement and never throw

corinagum: k thank you

PR #2275, opened 2021-07-12
Bumped Web Chat to 4.14.0

tonyanziano:
Bumped `botframework-webchat` version to `@4.14.0`

### Minor
- fixed some breaking changes with the `formidable` library.
- there was an indirect bump (via `botframework-connector`) to a version of `botframework-schema@4.14.0` that was causing a type mismatch with the locally installed version in the Emulator (^4.3.4) so I locked the local version to 4.14.0


packages/app/client/package.json
```diff
@@ -110,7 +110,7 @@
     "base64url": "3.0.0",
     "botframework-config": "4.4.0",
     "botframework-schema": "^4.3.4",
-    "botframework-webchat": "4.13.0",
+    "botframework-webchat": "4.14.0",
     "botframework-webchat-core": "4.13.0",
```
srinaath, line 114: Does core need no updating?

tonyanziano: good catch!

corinagum:
@tonyanziano btw, could you remove me from codeowners? :)

srinaath (dismissed):

srinaath (approved):

PR #2287, opened 2021-09-01
Fixes activity id returned in direct line POST response and web chat send status

stevkan:
Currently, Emulator does not return the activity id in the POST response after a file has been uploaded. Additionally, the send status hangs showing 'Sending' before timing out resulting in 'Send failed. Retry.'

Returning the correct source for the activity id value corrects the POST response. Including the `sendToSubscribers()` method addresses the send status and, further, allows the file to be downloaded.


tonyanziano (approved):
Looks good. Thanks for the contribution @stevkan !

tonyanziano:
Fix the two linting issues and I'll re-approve and merge

tonyanziano (approved):

PR #2324, opened 2021-11-12
[Accessibility] [Screen Reader] Fix screen reader announcement when navigating through "My bots" list

luislera:
### Describe the issue

If Screen reader users navigate on the Welcome screen and screen reader focus moves on the first bot item element, list 1 is announced with the same control. It would be confusing for screen reader users to know which list 1 is available with control.

**Actual behavior:**

When the focus moves on the first item element, list 1 is announced with the same control.
The screen reader announces list 1 with most of the control on the application.

**Expected behavior:**

When the focus moves on the first item element, the screen reader should not announce list 1 with the same control.
Because it’s confusing for users to know which list 1 is available.

**Repro Steps:**

3. Navigate to the welcome screen.
4. Verify that list 1 is announced with control or not.

- Added the region presentation to avoid including "list" with the first item when announcing the element name

### Screenshots

imagen


tonyanziano (approved):
Thanks!

PR #2326, opened 2021-11-12
[Accessibility] [Keyboard Navigation] Fix screen reader announcement when navigating through the Hide and Copy buttons

luislera:
### Describe the issue

If users navigate on New Bot Configuration Dialog and focus move on the hide button and screen reader announce list with it, then it would be difficult for users to understand which list item is available with the hide button.

**Actual behavior:**

When the user navigates on New Bot Configuration Dialog and focus moves on the hide button, the screen reader announces list with it and it confusing for users which list is available with the hide button.

**Expected behavior:**

When the user navigates on New Bot Configuration Dialog and the focus moves on the hide button, list should not be announced with the hide button, so that the user would not get confused.

**Repro Steps:**

5. Navigate on the dialog and enter inputs in edit fields.
6. Verify the screen reader announces "list" with the Hide button or not.

- Added the region presentation to avoid including "list" when reading the button's text

### Screenshots

imagen


tonyanziano (approved):
Looks good 👍

PR #2327, opened 2021-11-12
[Accessibility] [Screen Reader] Fix the screen reader announcement for required edit fields

luislera:
### Describe the issue

If users navigate on New Bot Configuration Dialog and the focus moves on required edit field and screen reader invalid announce with required edit field, then it would be difficult for users to understand which invalid data entered and what is invalid on edit field.

**Actual behavior:**

When the user navigates on New Bot Configuration Dialog and the focus moves on required field and screen reader announces invalid required edit field and it confusing for users which invalid data entered on required edit field.

**Expected behavior:**

When the user navigates on New Bot Configuration Dialog and the focus moves on required field, screen reader should not announce invalid with required edit field, so that the user would not get confused.

**Repro Steps:**

5. Navigate on the dialog and enter inputs in edit fields.
6. Verify that screen reader announces invalid with edit field or not.

- Added aria-required to avoid screen reader announces as invalid all the required edit fields

### Screenshots

imagen

### DIFF
--- a/packages/sdk/ui-react/src/widget/textField/textField.tsx
+++ b/packages/sdk/ui-react/src/widget/textField/textField.tsx
@@ -61,6 +61,7 @@ export class TextField extends Component<TextFieldProps, Record<string, unknown>
       errorMessage,
       children,
       inputRef,
+      required,
       ...inputProps
     } = this.props;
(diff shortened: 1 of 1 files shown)


tonyanziano (approved):
Nice!

PR #2330, opened 2021-11-15
External/southworks/fix/settings/fix split screen

luislera:
### Describe the issue

If users navigate on setting screen and split the screen and navigate on the screen and screen content and browser control is not visible properly, then it would be difficult for users to know information and difficult for users to complete their task.

**Actual behavior:**

After selection of setting control Emulator settings screen open and then select Split control, the page gets split but after split screen, content and browser control does not visible properly, users confuse on the screen to know information and control.

**Expected behavior:**

After selection of setting control Emulator settings screen open and then select Split control page get split, So after split screen, content and browser control should be visible properly, the user easily works on screen as normal view.

**Repro Steps:**

6. Navigate to the Settings icon on the left pane and select it.
7. Settings page opens, navigate on the page.
8. Verify that the user is able to read content on screen after split screen.

- Replaced fieldset/legend tags with div/span to avoid confusing the user when the screen reader announces the group name and also, to be responsive when using the split screen feature

### Screenshots

imagen

### DIFF
--- a/packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.scss
+++ b/packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.scss
@@ -24,19 +24,13 @@
     }
   }
 
-  fieldset {
-    position: relative;
-    margin: 0;
-    padding: 0;
-    border: none;
-  }
-
-  legend {
+  .legend {
     font-family: var(--default-font-family);
     font-size: 19px;
     font-weight: 200;
     margin: 0 4px 16px 4px;
     padding: 0;
+    display: block;
   }
 }
 
--- a/packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.scss.d.ts
+++ b/packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.scss.d.ts
@@ -1,5 +1,6 @@
 // This is a generated file. Changes are likely to result in being overwritten
 export const appSettingsEditor: string;
+export const legend: string;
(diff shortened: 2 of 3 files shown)


packages/app/client/src/ui/editor/appSettingsEditor/appSettingsEditor.scss
```diff
@@ -24,19 +24,13 @@
...
 
-  fieldset {
-    position: relative;
-    margin: 0;
-    padding: 0;
-    border: none;
-  }
-
-  legend {
+  .legend {
```
tonyanziano, line 27: Since you added a new class to the styles here, can you make sure that a definition for `styles.legend` gets generated for `appSettingsEditor.scss.d.ts`?

There should be a line in that file that says something like `export const legend: string;`.

This should be generated automatically if you have the client development server running (`npm start` in `/packages/app/client` and you reference the class in a `.tsx` file.

If it is not auto generating the file, feel free to add the line manually.

luislera: Hey @tonyanziano. Yes, this line was automatically added. That is why we did not push this file.

imagen

Should we push it?

tonyanziano: Yes please push that file and feel free to push these files in the future.

luislera: @tonyanziano Done!

tonyanziano (approved):
Thanks!

PR #2333, opened 2021-11-19
[Accessibility] [Screen Reader] Show a message after copying the JSON from the inspector panel

luislera:
### Describe the issue

If users navigate on JSON section under Chat with Bot screen, and select copy JSON and it get copied but visually message not appear on screen, then it would difficult for users to know JSON get copied or not, perform action get completed or not.

**Actual behavior:**

When users connect using the Bot screen and navigate to the JSON section and select Copy JSON control and it gets selected, but the copied message does not visually appear on screen, and it would be difficult for users to know the action is completed or not.

**Expected behavior:**

When users connect using the Bot screen and navigate to the JSON section and select Copy JSON control and it gets selected, a copied message should visually appear on the screen, so that users easily know JSON get copied.

**Repro Steps:**

5. Bot Gets connected and a new Tab opens, navigate to the Type a Message field and enter any chat message.
7. Navigate on JSON Section, and select the Copy JSON button.
8. Verify whether the copied message appears on the screen or not.

- Added a message box to be shown after the Copy JSON button is pressed

### Screenshots

imagen

### DIFF
--- a/packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.spec.tsx
+++ b/packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.spec.tsx
@@ -518,11 +518,11 @@ describe('The Inspector component', () => {
 
       it('"create-aria-alert"', () => {
         event.channel = 'create-aria-alert';
-        event.args = ['I am an alert!'];
+        event.args = ['Activity JSON copied to clipboard.'];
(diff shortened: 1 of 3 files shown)


tonyanziano (approved):
The only thing that concerns me is that this UX might be annoying for users (having to dismiss a message box after copying), but we can address it later on if that's the case.

PR #2339, opened 2021-11-19
[Accessibility] [Keyboard Navigation] Remove hidden buttons from the Inspection panel to avoid being accessible using the mouse

luislera:
### Describe the issue

If users Navigate on the JSON section and the hide diff button is available on this section and it's not accessible and visible using the keyboard navigation, then it would be difficult for users to know which control is available on there and the purpose of that control.

**Actual behavior:**

Navigating on the JSON section, the Hide diff button is available but that control is not accessible and visible using the keyboard navigation.

**Expected behavior:**

Navigating on the JSON section, the Hide diff button is available so that control should be accessible and permanently visible using the keyboard navigation.

**Repro Steps:**

5. Bot Gets connected and a new Tab opens, navigate to the Type a Message field and enter any chat message.
7. Navigate on JSON Section.
8. Verify whether the hide diff button is accessible or visible using the keyboard navigation.

- Removed the disabled buttons (left arrow, right arrow, diff) from the DOM so the user cannot interact with them

### Screenshots

Disabled buttons were removed from the DOM
imagen

### DIFF
--- a/packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.scss
+++ b/packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.scss
@@ -42,6 +42,10 @@
       background-repeat: no-repeat;
     }
   }
+
+  button[aria-disabled="true"] {
+    display: none;
+  }
 }
 
 .inspector-container {
--- a/packages/app/client/src/ui/editor/panel/panel.scss
+++ b/packages/app/client/src/ui/editor/panel/panel.scss
@@ -53,9 +53,9 @@
       }
     }
 
-    // left arrow diff button
-    & > button[name="leftArrow"] {
-      margin-left: auto;
+    // json button
+    & > button[name="json"] {
+      margin-right: auto;
     }
 
     // toggle diff button


packages/extensions/json/bf-extension.json
```diff
@@ -90,10 +90,6 @@
                 "aria-selected": true,
                 "title": "Hide diff",
                 "aria-label": "Hide diff"
-              },
-              "disabled": {
-                "aria-disabled": true,
-                "title":"Disabled diff"
```
tonyanziano, line 96: So I think that we actually need to keep these attributes because they are supposed to be present on the "Show / Hide diff" button when it is actually disabled.

After looking into the issue, I think that the real problem is that this button is interactable and clickable while it is actually supposed to be in a "disabled" state. If you look at this GIF, you can see that I am actually able to click on these buttons that _should be disabled_:

invisible-disabled-buttons

So what we need to do is instead fix this initial state where the buttons are invisible, but can be clicked. These inspector accessory buttons that are disabled _should not be interactable_.

I think the easiest solution would just be to remove these elements from the DOM when they are disabled so that they are not interactable. These definitely aren't supposed to be clickable unless you are inspecting a "Bot State" trace.

tonyanziano (changes requested):

tonyanziano (approved):
Thanks!

PR #2340, opened 2021-11-29
[Accessibility] [Keyboard Navigation] Fix luminosity ratio in Bot Explorer list items

luislera:
### Describe the issue

If users Navigate on the Bot explorer section and the luminosity ratio is less than for added endpoint http://local... text and it would be difficult for users to understand the text.

**Actual behavior:**

Luminosity ratio is less than for added endpoint http://local... text, under Bot Explorer.

**Expected behavior:**

Luminosity ratio should be equal or more than for added endpoint http://local... text, under Bot Explorer.

**Repro Steps:**

6. Navigate to BOT Explorer on the left pane and select it.
7. Bot Explorer Pane opens, navigate on the pane.
8. Verify that luminosity is equal to or less than for text or not.

- Updated list item foreground color to meet the luminosity ratio criteria.

### Screenshots

Color contrast analyzer
imagen

### DIFF
--- a/packages/app/client/src/ui/styles/themes/light.css
+++ b/packages/app/client/src/ui/styles/themes/light.css
@@ -6,7 +6,6 @@ html {
   --box-shadow-color: var(--neutral-6);
 
   /* Highlight colors */
-  --list-item-color: var(--neutral-9);
   --list-item-color: #605E5C;
   --list-item-selected-color: var(--neutral-2);
   --list-item-hover-bg: var(--neutral-4);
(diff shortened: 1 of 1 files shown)


packages/app/client/src/ui/styles/themes/light.css
```diff
@@ -7,6 +7,7 @@ html {
 
   /* Highlight colors */
   --list-item-color: var(--neutral-9);
```
tonyanziano, line 9: this line should be removed so that we don't have duplicate CSS definitions for the same variable

tonyanziano (changes requested):
One small change and then it should be good.

tonyanziano (approved):

PR #2341, opened 2021-11-29
[Accessibility] [Supporting Platforms] Add a tooltip for keyboard navigation to the left panel

luislera:
### Describe the issue

If users Navigate on Resources and Bot Explorer control and tooltip is not appearing for that control, then it would be difficult for users to understand control, if users don't know about the control icon, they would be unable to identify the control icons.

**Actual behavior:**

When users navigate on Resource and Bot Explorer control Present side pane and tooltip is not appearing for that control and it confusing for users which control is that.

**Expected behavior:**

When users navigate on Resource and Bot Explorer control Present side pane So tooltip should appear for that control.

**Repro Steps:**

6. Navigate to the Resources icon on the left pane
7. Verify that tooltip is appearing for control or not.

- Added a span element to be shown as a tooltip for each option on the left panel.

### Screenshots

imagen

Keyboard tooltip example
imagen

### DIFF
--- a/packages/app/client/src/ui/shell/navBar/navBar.scss
+++ b/packages/app/client/src/ui/shell/navBar/navBar.scss
@@ -147,9 +147,9 @@
     > span {
       position: absolute;
       z-index: 1;
-      background-color: #fff;
-      color: #000;
-      border: 1px solid #000;
+      background-color: var(--nav-link-tooltip-bg);
+      color: var(--nav-link-tooltip-color);
+      border: var(--nav-link-tooltip-border);
       white-space: nowrap;
       font-size: 12px;
       text-align: center;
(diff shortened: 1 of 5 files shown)


packages/app/client/src/ui/shell/navBar/navBar.scss
```diff
@@ -140,6 +143,23 @@
...
+      white-space: nowrap;
+      font-size: 12px;
+      text-align: center;
+      padding: 4px;
+      top: 22px;
+      left: 22px;
+      transition-property: visibility;
+      transition-duration: 2s;
+      visibility: hidden;
+    }
```
tonyanziano, line 162: Looks good, but I think we should also style this when in High Contrast mode to follow the same sort of color scheme that VS Code uses:

image

tonyanziano (changes requested):

tonyanziano (approved):
Great, thanks!

PR #2353, opened 2021-12-14
 [Accessibility] [Visual Requirements] Fix the role announced by the screen reader when focusing the left panels

luislera:
### Describe the issue

If screen reader announcing role as a toolbar for chat file and transcripts control, rather than drop-down or button, so it would be difficult for users to understand the purpose of the control and would be get confused about the control role.

**Actual behavior:**

Screen reader announcing role as a toolbar for "Chat files and Transcripts" control under Resource section. And it's confusing for users to understand the control role.

**Expected behavior:**

Screen reader should not announce role as a toolbar for "Chat files and Transcripts" control, it should be announced role as a drop-down or button for "Chat files and Transcripts" control, rather than toolbar under Resource section.

**Repro Steps:**

6. Navigate to the Resources icon on the left pane and select it.
7. Resources Pane opens, navigate on the pane.
8. Verify the screen reader announce the correct role or not.

- Added a new prop to conditionally set the role on left panels.

### Screenshots

botExplorerBar component
imagen

resourcesBar component
imagen

servicePane component
imagen

### DIFF
--- a/packages/app/client/src/ui/shell/explorer/botExplorerBar/botExplorerBar.tsx
+++ b/packages/app/client/src/ui/shell/explorer/botExplorerBar/botExplorerBar.tsx
@@ -56,8 +56,9 @@ export default class BotExplorerBar extends React.Component<BotExplorerBarProps,
           title="Endpoint"
           ariaLabel="Endpoints"
           elementRefHandler={this.setEndpointsPanelRef}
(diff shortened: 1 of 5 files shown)


tonyanziano (approved):
Awesome, thanks!

PR #2356, opened 2021-12-15
[Accessibility] [Screen Reader] Fix the screen reader announcement for the "Microsoft App password" field

luislera:
### Describe the issue

If screen reader users navigate on the Open a bot dialog box and focus on the Microsoft app password field and screen reader announcing password word twice, then it would be confusing for users to understand how many password field is available there.

**Actual behavior:**

Screen reader announcing "Password" word twice with "Microsoft app password" field, under Open a bot dialog box.

**Expected behavior:**

Screen reader should not announce the "Password" word twice with the "Microsoft app password" field, under the Open a bot dialog box.
It should announce only one time.

**Repro Steps:**

4. Open A bot Dialog opens, navigate on the dialog and enter inputs in edit fields.
5. Verify that screen reader twice password announce or not.

- Added a prop to handle the aria-label property to fix the problem issue on Linux only

### Screenshots

openBotDialog component
imagen

botCreationDialog component
imagen

### DIFF
--- a/packages/app/client/src/ui/dialogs/botCreationDialog/botCreationDialog.tsx
+++ b/packages/app/client/src/ui/dialogs/botCreationDialog/botCreationDialog.tsx
@@ -31,7 +31,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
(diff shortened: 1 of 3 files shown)


tonyanziano (approved):
Looks good!

PR #2357, opened 2021-12-15
[Accessibility] [Screen Reader] Fix the JSON button role in the Inspection panel

luislera:
### Describe the issue

If screen reader users navigate on JSON section and screen reader button announcing with JSON, then it would be confusing for users because it is not functional, it's not working as like button.

**Actual behavior:**

When the focus moves on the JSON section, the Screen reader announces its role as a Button with JSON, but it is not functional, it is not working as a button.

**Expected behavior:**

When the focus moves on the JSON section, So Screen reader role as a Button should not announce with JSON, because it is not functional, and it is confusing for the users, after performing an action on control nothing happens.

**Repro Steps:**

5. Bot Gets connected and a new Tab open, navigate to the Type a Message field and enter any chat message.
7. Navigate on JSON Section.
8. Verify whether the screen reader button announces with JSON or not.

- Updated the button role conditionally based on the button name

### Screenshots

imagen

### DIFF
--- a/packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.tsx
+++ b/packages/app/client/src/ui/editor/emulator/parts/inspector/inspector.tsx
@@ -269,6 +269,7 @@ export class Inspector extends React.Component<InspectorProps, InspectorState> {
         key={config.id}
         disabled={!enabled}
         name={config.id}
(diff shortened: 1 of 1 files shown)


tonyanziano (approved):
Thanks!

PR #2361, opened 2021-12-17
 [Accessibility] [Screen Reader] Fix the focus on the Get Started with Channel page

luislera:
### Describe the issue

If users navigate on getting started with channel screen and after open screen, focus indicator does not land on first interactive control, so it would be difficult for users to understand focus order on the screen.

**Actual behavior:**

After selection of Get Started with Channel (Bot Inspector) control screen gets open but the focus indicator does not land on the first interactive control.

**Expected behavior:**

After selection of Get Started with Channel (Bot Inspector) control screen get open, So the focus indicator should be not land on the first interactive control. So that users easily navigate on the screen.

**Repro Steps:**

6. Navigate to Help Menu on the ribbon and select it.
7. View menu opens, navigate to Get Started with Channel (Bot Inspector) and select it.
8. Getting started with Bot Inspector page opens, navigate on the page.
9. Verify that focus get to land on the first interactive control or not.

- Added a node package to render the HTML of the markdown file and focus the first interactive element.

### Screenshots

imagen

### DIFF
--- a/packages/app/client/src/ui/editor/markdownPage/markdownPage.tsx
+++ b/packages/app/client/src/ui/editor/markdownPage/markdownPage.tsx
@@ -33,6 +33,7 @@
 import * as React from 'react';
 import { Component } from 'react';
 import MarkdownIt from 'markdown-it';
+import ReactHtmlParser from 'react-html-parser';
 
 import { GenericDocument } from '../../layout';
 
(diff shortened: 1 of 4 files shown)


packages/app/client/src/ui/editor/markdownPage/markdownPage.tsx
```diff
@@ -80,15 +81,24 @@ export class MarkdownPage extends Component<MarkdownPageProps> {
   }
 
   public render() {
+    const hrefConst = '#bot-state-inspection';
```
tonyanziano, line 84: I think a name like `firstFocusableHref` or `firstFocusableElement` would be more clear

packages/app/client/src/ui/editor/markdownPage/markdownPage.tsx
```diff
@@ -80,15 +81,24 @@ export class MarkdownPage extends Component<MarkdownPageProps> {
...
+    const transform = node => {
+      if (node.name == 'a' && node.attribs && node.attribs.href == hrefConst) {
+        return (
+          <a ref={this.setBotInspectorRef} href={hrefConst}>
+            jump to Bot State Inspection
+          </a>
+        );
+      } else {
+        return undefined;
+      }
```
tonyanziano, line 94: Since we only use the `<MarkdownPage>` for the Bot Inspector documentation this logic should work, but if we ever add another rendered Markdown page we will have to change this so that the first focusable element of the page is passed as a prop or something.

tonyanziano (changes requested):
One minor comment about a variable name. Looks good otherwise!

packages/app/client/src/ui/editor/markdownPage/markdownPage.tsx
luislera: We made some changes and use a prop to set the first focusable element. We used a different MD file just to ensure it is working as expected.

tonyanziano: Awesome! Nice work

tonyanziano (approved):
Great, thanks!

PR #2373, opened 2022-03-21
chore: update critical dependencies

OEvgeny:
The following third-party dependencies need to be updated:

**Critical**|**version**
:-----:|:-----:
ejs|3.1.6
property-expr|2.0.3
url-parse|1.5.8

Bumped `electron-builder` and `lint-staged` packages to update these.


.eslintrc.react.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
   },
   rules: {
     'react/no-deprecated': 'warn',
+    'react/prop-types': 'off',
```
OEvgeny, line 17: The rule causes false-positives as we don't use `PropTypes` we use TypeScript instead

cwhitten (approved):

PR #2378, opened 2022-04-11
a11y: fix platform-specific issues

OEvgeny:
1. [VSTS-63988](https://fuselabs.visualstudio.com/Composer/_workitems/edit/63988) The [default `"complementary"`](https://github.dev/microsoft/BotFramework-WebChat/blob/6a5359f4b74827f9b8919eaf802aa0374217ed15/packages/component/src/BasicWebChat.tsx#L76) role cannot be included into another landmark. Use `"region"` role for the chat instead.

image

2. [VSTS-64729](https://fuselabs.visualstudio.com/Composer/_workitems/edit/64729) On linux, when using orca, it doesn't announce an actual role for menu items. Fixed by adding the "menu item" text `aria-description` property only on linux

image

3. [VSTS-64703](https://fuselabs.visualstudio.com/Composer/_workitems/edit/64703) Sidebar section header buttons had a `toolbar` role which is not descriptive. Fixed by changing the role to `button` and adding toolbar around section action buttons.
 
image
image

4. [VSTS-64523](https://fuselabs.visualstudio.com/Composer/_workitems/edit/64523) The selected tab announcement doesn't happen to me when using orca. Changed the approach by using `aria-description` property on linux.

image

Tested with orca version 41.0.

Screen reader settings:
image
image

### DIFF
--- a/packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
+++ b/packages/app/client/src/ui/editor/emulator/parts/chat/chat.tsx
@@ -136,7 +136,7 @@ export class Chat extends PureComponent<ChatProps, ChatState> {
             username={currentUser.name || 'User'}
             webSpeechPonyfillFactory={webSpeechPonyfillFactory}
           >
-            <BasicWebChat />
+            <BasicWebChat role="region" />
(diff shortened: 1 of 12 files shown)


tonyanziano (approved):
Looks great! Thanks 👍

PR #2379, opened 2022-04-14
fix: keyboard shortcuts don't work when telemetry and updates dialogs appear together

OEvgeny:
Normally, there is no way for a user to open a dialog when another dialog has been opened. However, this occurs when the "Updates available" dialog is opened as a result of the 'update-available' event.

Added a method to wait for the currently displayed dialog to be hidden, and used it for the "Updates available" dialog.

 Fixes [VSTS-63986](https://fuselabs.visualstudio.com/Composer/_workitems/edit/63986)


tonyanziano (approved):
Awesome! We definitely needed this.

PR #2359, opened 2021-12-15
[Accessibility] [Screen Reader] Add a visual message when using the full screen option

luislera:
### Describe the issue

If users Navigate on toggle full screen menu item and select it of that control also exit full screen, the screen reader is not announcing the same information about for both time and it would be confusing for screen reader users perform action is correct or not.

**Actual behavior:**

Navigating on View menu item and select it to toggle full screen control and exit full screen also with shortcut key which provided with control, visually message is not appearing on screen also screen reader is not provided the same information after performing the action on both times.

**Expected behavior:**

Navigating on View menu item and select it to toggle full screen and exit full screen also with shortcut key which provided with control, visually message should appear also screen reader should be provided proper information about full screen and exit full screen. There should appear a visual message also the screen reader should provide that information.

**Repro Steps:**

1. Open Bot Framework Emulator installed on the system.
2. Bot Framework Emulator Application with Welcome screen opens.
3. Navigate on the welcome screen and select Create a New BOT configuration link.
4. New BOT Configuration dialog opens.
5. Navigate on the dialog and enter inputs in edit fields then select the save and connect button.
6. Navigate to View Menu on the ribbon and select it.
7. View menu opens, navigate to Toggle Full screen control and select it. The application gets full screen.
8. Verify that visual message appears, and screen reader announces full screen information or not.

### Changes included in the PR

- Added a visual message that will be displayed when entering/exiting the full screen mode

### Screenshots

Test cases executed successfully

eventHandlers component
imagen

appMenuTemplate component
imagen

PR #2380, opened 2022-04-20
a11y: announce theme switching and fullscreen commands

OEvgeny:
* Add announcement for theme switching for macOS
* Make the theme switching message to start with visible content
* Move the full screen command to the UI commands and add announcement
* Refactor code to use full screen command instead of calling BrowserWindow API directly
* Update tests

[VSTS-64005](https://fuselabs.visualstudio.com/Composer/_workitems/edit/64005)

image

### DIFF
--- a/packages/app/client/src/commands/uiCommands.spec.ts
+++ b/packages/app/client/src/commands/uiCommands.spec.ts
@@ -41,6 +41,7 @@ import {
   SharedConstants,
 } from '@bfemulator/app-shared';
 import { CommandRegistry, CommandServiceImpl, CommandServiceInstance } from '@bfemulator/sdk-shared';
+import { remote } from 'electron';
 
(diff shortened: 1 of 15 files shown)


OEvgeny:
For the record, this removes dialog windows popping-up after fullscreen mode switching, introduced in #2359
as a matter of the following reasons:

- After checking accessibility requirements, I haven't found where the requirement came from. I haven't found a Microsoft product that meets this requirement either.
- The UX is not great: users need to explicitly close the dialog after entering or leaving fullscreen mode.
- It doesn't work well on all OSes: on Mac, there is an issue toggling the mode from the menu when opening a dialog is involved.

In case the original bug gets reopened, we should bring this to the accessibility hours meeting. And even if we had to introduce the visual part, we'd need a design for this particular case.

tonyanziano (approved):
Looks good!

PR #2412, opened 2023-01-23
Bump electron to the latest minor version and cleanup deps

OEvgeny:
- Fix electron build version
- Update style types
- Update json token validation package to avoid security complaints


cwhitten (approved):
Approved when CI is green

OEvgeny:
@cwhitten it is, just can't access to gh for some reason

PR #2417, opened 2023-02-15
fix: auth flow issues

OEvgeny:
This fixes the following auth flow issues:
- Race condition caused by delayed subscription to navigation events causes auth proccess to silently stuck
- Potentially vulnerable URL parsing logic
- Possible identity spoofing/theft caused by the race condition


cwhitten (approved):

packages/app/main/src/services/azureAuthWorkflowService.ts
```diff
@@ -153,9 +152,21 @@ export class AzureAuthWorkflowService {
...
-    return new Promise<BrowserWindow>(resolve => {
-      browserWindow.once('ready-to-show', () => resolve(browserWindow));
-    });
+
+    const setupWindow = async () => {
+      // setup navigation events before the window is ready
+      const authResponse = this.waitForAuthResult(browserWindow, replyUrl);
+
+      // if redirect happened before ready-to-show event, window won't recieve the event
+      await Promise.race([authResponse, new Promise(resolve => browserWindow.once('ready-to-show', resolve))]);
```
a-b-r-o-w-n, line 161: This is really nice. Good usage of `Promise.race`

a-b-r-o-w-n (approved):

